PR #3, opened 2021-10-11
Create API FlowCollector in flows.netobserv.io

jotak:
Created via `operator-sdk create api ...`
Includes go model and generated manifest / CRD


jotak:
We need to decide here a couple of things, like the domain, group and name of the custom resource.
It was designed here: https://docs.google.com/document/d/1TTgywlUn52faDgRPbCbxu_1Wpkmraa92zyWpLC0LN9A/edit# however I changed a couple of things since the design doc, because it felt weird or incorrect:

- domain: currently `netobserv.io` (it doesn't have to refer to an actual domain afaik but anyway we should consider owning it). When it will go downstream, it's going to be `openshift.io`.
- group: currently `flows`. It's supposed to categorize stuff within the domain. At the moment we only have this one, but that would keep the door open for later additions in the "netobserv" operator space in other areas like network policies, bpf metrics, etc.
- the resource itself is `FlowCollector`. Which means the user will run commands like `oc describe/edit flowcollector` to get a description or edit with CLI.

This is not set in stone, if we need to revisit later (e.g. to better align with whatever will be designed for the openshift/downstream operator) it still will be possible

cc @stleerh @mariomac   @OlivierCazade  @pweil-

PR #7, opened 2021-10-25
NETOBSERV-81 define resource limits for goflow-kube

jpinsonneau:


controllers/goflowkube/goflowkube_test.go
```diff
@@ -0,0 +1,125 @@
+/*
```
jotak, line 1: FYI - though I couldn't find any written rule about that, in general we don't write license headers per file, considering the LICENSE file at the project root is sufficient.

jotak: btw thanks for writing these tests!

jpinsonneau: You are welcome !

I saw that in [/api/v1alpha1/flowcollector_types.go#L1](https://github.com/jpinsonneau/network-observability-operator/blob/14deb157fc3ddd0db59bb68f3be543490b67b2cd/api/v1alpha1/flowcollector_types.go#L1) but not in the other ones.

We can remove them for sure :+1:

PR #8, opened 2021-10-25
NETOBSERV-68: generate ConfigMap for CNO

mariomac:
https://issues.redhat.com/browse/NETOBSERV-68

Maintains a `ovs-flows-config` `ConfigMap` whose contents are synchronized with the existing `FlowCollector` configuration.


controllers/ovs/flowsconfig.go
```diff
@@ -0,0 +1,144 @@
...
+
+	flowsv1alpha1 "github.com/netobserv/network-observability-operator/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ovsConfigName = "ovs-flows-config"
+)
```
jpinsonneau, line 23: Any reason to not use constants.go file ?

controllers/flowcollector_controller.go
```diff
@@ -18,11 +22,21 @@ import (
...
 // See also https://github.com/operator-framework/operator-lib/issues/74
 const operatorNamespace = "network-observability"
+const finalizerName = "flows.netobserv.io/finalizer"
 
 // FlowCollectorReconciler reconciles a FlowCollector object
 type FlowCollectorReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme      *runtime.Scheme
+	flowsConfig ovs.FlowsConfigController
```
jotak, line 31: Maybe this can be renamed, something like `ovsConfigController` or `ovsFlowsConfigController` ? The `ovs` prefix to avoid confusion with the flow collector CR / controller and `Controller` since it's actually not a config object, also to avoid confusion with the other `ovs.flowsConfig` struct.

controllers/flowcollector_controller.go
```diff
@@ -67,15 +84,41 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
+	// ovs-flows-config map for CNO
+	if err := r.flowsConfig.Reconcile(ctx, desired); err != nil {
+		log.Error(err, "Failed to reconcile ovs-flows-config ConfigMap")
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *FlowCollectorReconciler) checkFinalizerStatus(ctx context.Context, desired *flowsv1alpha1.FlowCollector) error {
```
jotak, line 99: side note: I haven't much looked into finalizers yet but maybe we should also use it to clean up other resources that we're creating?

controllers/ovs/flowsconfig.go
jotak: I think as long a constant isn't needed across several packages it's better to keep them in their package

controllers/ovs/flowsconfig_types.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/netobserv/network-observability-operator/api/v1alpha1"
+)
+
+type flowsConfig struct {
+	v1alpha1.FlowCollectorIPFIX `json:",inline"`
+	SharedTarget                string `json:"sharedTarget,omitempty"`
+	NodePort                    int32  `json:"nodePort,omitempty"`
+}
+
+// TODO: for better maintainability, implement map-to-struct/struct-to-map using reflection
```
jotak, line 16: I'm surprised there's nothing like that in kube client/api to deal with config maps.
Alternatively we could use this: https://pkg.go.dev/github.com/mitchellh/mapstructure?utm_source=godoc

controllers/ovs/flowsconfig_types.go
```diff
@@ -0,0 +1,66 @@
...
+		if mf, err := strconv.ParseInt(val, 10, 32); err != nil {
+			return nil, fmt.Errorf("wrong nodePort value: %w", err)
+		} else {
+			config.NodePort = int32(mf)
+		}
+	}
+	return &config, nil
+}
+
+func (fc *flowsConfig) asStringMap() map[string]string {
```
jotak, line 48: Some unit tests on conversion functions would be neat ; e.g. to make sure converting back and forth keeps it consistent

config/rbac/flowtracing_config_subscriber_role.yaml
```diff
@@ -0,0 +1,32 @@
+kind: Role
```
jotak, line 1: I wonder if this is better placed here or in the CNO repo.. is it something discussed already with the cno team?

controllers/ovs/flowsconfig.go
```diff
@@ -0,0 +1,144 @@
...
+		rlog.Info("Provided IPFIX configuration differs current configuration. Updating")
+		return c.client.Update(ctx, c.flowsConfigMap(desired))
+	}
+
+	rlog.Info("No changes needed")
+	return nil
+}
+
+func (c *FlowsConfigController) Finalize(ctx context.Context) error {
+	err := c.client.DeleteAllOf(ctx, &corev1.ConfigMap{
```
jotak, line 72: what's the difference with a simple call of `client.Delete` ? we're deleting a single object from its name, no?

controllers/ovs/flowsconfig.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return err
+	}
+
+	if current == nil {
+		rlog.Info("Provided IPFIX configuration. Creating " + ovsConfigName + " ConfigMap")
+		return c.client.Create(ctx, c.flowsConfigMap(desired))
+	}
+
+	if desired != current {
```
jotak, line 62: if I'm correct, `desired` and `current` are pointers, they'll never be equal?

controllers/ovs/flowsconfig.go
```diff
@@ -0,0 +1,144 @@
...
+	case constants.DeploymentKind:
+		svc := corev1.Service{}
+		if err := c.client.Get(ctx, types.NamespacedName{
+			Namespace: c.operatorNamespace,
+			Name:      constants.GoflowKubeName,
+		}, &svc); err != nil {
+			return nil, err
+		}
+		// TODO: if spec/goflowkube is empty or port is empty, fetch first UDP port in the service spec
+		conf.SharedTarget = fmt.Sprintf("%s:%d", svc.Name, coll.Spec.GoflowKube.Port)
```
jotak, line 126: the service ClusterIP is necessary here, as OVS won't resolve domain names

jotak: However I wonder if the cluster IP is immediately available after service creation or if we need to introduce some delay / retries ...

controllers/ovs/flowsconfig_types.go
mariomac: Yeah, the pain here is that it's not just about serializing/deserializing as map, but also that the configmap data is in the form `map[string]string` and it requires to convert from string to `int` in some fields.

controllers/flowcollector_controller.go
mariomac: This function is for removing resources that are not managed by Kubebuilder, such as the flows-config configmap. I guess that CRDs that are created/manager by Kubebuilder should be removed automatically. But definitely we should check.

controllers/ovs/flowsconfig_types.go
jotak: I haven't tried it but I think `mapstructure` lib can deal with non-string fields too, nested fields, etc.

config/rbac/flowtracing_config_subscriber_role.yaml
mariomac: No, I haven't. My reasoning was "NOO is the owner/manager of this resource, so NOO is the responsible of managing the access to it".

I'm also thinking in the namespace. If we change it to e.g. `openshift-network-observability` in the upstream version, we shouldn't need to do the change in the CNO.

controllers/ovs/flowsconfig.go
mariomac: d'oh!

mariomac: Proof that this PR needs more testing

controllers/ovs/flowsconfig.go
mariomac: If it is not available, we can return a status error asking for a retry

controllers/ovs/flowsconfig_types.go
mariomac: I tried it and worked like a charm when passing from `map[string]string` --> `flowsConfig`. It required extra tweaking when passing from `flowsConfig` --> `map[string]string` but the result is anyway much better than hardcoding the struct fields.

controllers/ovs/flowsconfig_types.go
mariomac: They are currently tested through the public APIs, as part of the expected behavior of the controller. With the usage of `mapstructure` library, the test coverage is full now.

controllers/ovs/flowsconfig.go
mariomac: Thinking more about it, I think we should implement the DNS resolution in the CNO (task [NETOBSERV-66](https://issues.redhat.com/browse/NETOBSERV-66) ), as this limitation is an internal detail from OVS.

controllers/ovs/flowsconfig.go
mariomac: Delete just works. Changed

config/rbac/flowtracing_config_subscriber_role.yaml
mariomac: Thinking again, you are right, it must rely in the CNO namespace to make it independent from any change in the upstream/downstream namespace, so we should implement it as part of https://issues.redhat.com/browse/NETOBSERV-66

jotak: do you think we need to add a field in our CR for the CNO namespace?

mariomac: Yes, as an optional argument that defaults to `openshift-network-operator`

controllers/ovs/flowsconfig.go
jotak: ok, sounds good to me

jotak:
I have tested and see the CM generated, it looks pretty good, I have just one remark, supposedly `int` fields are represented as strings, I'm not sure if that's expected:

```
$ k get cm ovs-flows-config -n openshift-network-operator -o yaml
apiVersion: v1
data:
  cacheActiveTimeout: 60s
  cacheMaxFlows: "400"
  sampling: "100"
  sharedTarget: goflow-kube:2055
kind: ConfigMap
```

(see `cacheMaxFlows` and `sampling`)

Something else, I'm not sure if it's related to this PR or was already an issue, but while undeploying the operator, it seems like the operation is stuck maybe during CR finalizer (I'm not sure about that, but running `kubectl patch flowcollector flowcollector-sample -p '{"metadata":{"finalizers": []}}' --type=merge` unblocks it). Do you have an idea about what's going on? Else we should create a task for further investigation.

jotak:
Actually I'm just realizing that the `sharedTarget` isn't correct, even considering that the CNO will resolve to IP, it should include the namespace since the CNO is in a different namespace : `  sharedTarget: goflow-kube.network-observability:2055`

mariomac:
@jotak unfortunately the `ConfigMap` type from the Kubernetes API just allows string fields: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L6074

On the other side, I'll check what's happening during the undeployment and fix the sharedTarget issue. Thanks for the report!

controllers/ovs/flowsconfig.go
mariomac: Update: CNO DNS configuration prevents proper/simple DNS resolution and the RFE spec states that we should provide the IP already resolved, so I have implemented the DNS resolution already in this PR ([commit](https://github.com/netobserv/network-observability-operator/pull/8/commits/4d9e9b187c5e37e6b661ccffc8658f71f5e02335))

jotak:
/lgtm
@mariomac feel free to merge

PR #9, opened 2021-10-26
NETOBSERV-63 HPA

jpinsonneau:
add hpa spec for flow collector with MinReplicas, MaxReplicas and TargetCPUUtilizationPercentage
target goflow-kube Deployment


controllers/goflowkube/goflowkube_reconciler.go
```diff
@@ -63,6 +69,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, desired *flowsv1alpha1.FlowC
 			newSVC := buildService(desired, r.OperatorNamespace)
 			r.createOrUpdate(ctx, oldSVC, newSVC, serviceKind)
 		}
+
+		//Delete or Create / Update Autoscaler according to HPA option
+		if desired.HPA == nil {
+			r.delete(ctx, oldASC, autoscalerKind)
```
jotak, line 75: what if `oldASC` is nil? We probably shouldn't try to delete it then

controllers/goflowkube/goflowkube_reconciler.go
```diff
@@ -63,6 +69,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, desired *flowsv1alpha1.FlowC
...
+		} else if oldASC == nil || autoScalerNeedsUpdate(oldASC.(*ascv1.HorizontalPodAutoscaler), desired) {
+			newASC := buildAutoScaler(desired, r.OperatorNamespace)
+			r.createOrUpdate(ctx, oldASC, newASC, autoscalerKind)
+		}
 	case daemonSetKind:
```
jotak, line 80: in the daemonset case we should certainly delete the HPA to not leave any unused object out there

controllers/goflowkube/goflowkube_reconciler.go
jpinsonneau: also changed in daemonSetKind case for oldSVC

jpinsonneau:
lgtm

PR #13, opened 2021-11-04
Netobserv 51 console plugin

OlivierCazade:
This pull request add reconcile function for the network observability console plugin.

This pull request work with the pull request in the console plugin repository:
https://github.com/netobserv/network-observability-console-plugin/pull/1

Please note that console plugin image does not exist yet on quay.io, you will need to build your own image to test this pull request.


config/rbac/role.yaml
```diff
@@ -27,6 +27,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
OlivierCazade, line 30: Side question related to the whole file.

This file is fully generated and any modification will be overwritten by kubebuilder.

Should we consider this file as a build artifact ? Is it part of good practice to still track it in the repository ?

jotak: I don't have a strong opinion.. one advantage to keep it in git is to have a quick & easy way to see which permissions are required, without having to fetch repo and deploy. But that's all, if people think it's better to remove (e.g. to have less PR conflicts) i'm fine either way

jpinsonneau:
Is it a good practice to duplicate _reconciler.go files since it's the exact same code except HPA ?

We could use a common file taking options I guess

OlivierCazade:
> Is it a good practice to duplicate _reconciler.go files since it's the exact same code except HPA ?
> 
> We could use a common file taking options I guess

The limit is always tricky to find, during development I also thought about making some factorisation here, but golang philosophy is to prefer small copy to small dependencies.

Note that there are a few more differences than the HPA :
- console plugin does not support daemonset
- console plugin also create a ConsolePlugin resource to register the plugin to the console

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -0,0 +1,144 @@
...
+				Namespace: ns,
+				Labels:    buildLabels(),
+			},
+		},
+		&rbacv1.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   pluginName,
+				Labels: buildLabels(),
+			},
+			Rules: []rbacv1.PolicyRule{},
```
jotak, line 125: since the rules are empty maybe we don't need a ClusterRole here, or is it necessary for other reasons?

api/v1alpha1/flowcollector_types.go
```diff
@@ -35,6 +35,9 @@ type FlowCollectorSpec struct {
 
 	// Loki contains settings related to the loki client
 	Loki FlowCollectorLoki `json:"loki,omitempty"`
+
+	// ConsolePlugin contains settings related to the console dynamic plugin
+	ConsolePlugin FlowCollectorConsolePlugin `json:"consoleplugin,omitempty"`
```
jotak, line 40: nit: could you rename the json field in camel case `consolePlugin` (and I just realize `goflowkube` is also not camel case maybe that's why you did that? i guess we'll have to rename it too, to stay consistent with the other fields)

controllers/flowcollector_controller.go
```diff
@@ -83,5 +101,6 @@ func (r *FlowCollectorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.DaemonSet{}).
 		Owns(&corev1.Service{}).
+		Owns(&osv1alpha1.ConsolePlugin{}).
```
jotak, line 104: What happens when it doesn't run in the context of openshift? (e.g. on kind) Creating the ConsolePlugin resource will fail obviously, but without breaking the rest?

jotak (commented):
code lgtm (just a few remarks), still i'd like to test

README.md
```diff
@@ -20,6 +20,22 @@ IMG="docker.io/myuser/netobserv:latest" make image-build image-push deploy
 
 By default, the operator is deployed in namespace "network-observability".
 
+### Enabing the console plugin
```
jotak, line 23: Typo: "Enabling"

controllers/flowcollector_controller.go
OlivierCazade: Thanks, I did not think about regular kubernetes cases.

I was thinking about adding a flag to enable/disable the plugin deployment, what do you think ?
Do you think we should try to automatically detect that the pod is running under OpenShift ?

jotak: actually, I think the console itself can be deployed in non-openshift kube. So the best solution would be to detect if the ConsolePlugin crd is installed
(we can leave it aside at the moment if you prefer - just create a follow-up task then - but we must make sure it won't make the whole operator crashing ; it must still deploy / reconcile goflow and the like)

jotak:
Testing OK I can see the Network traffic menu item in the console \o/

controllers/consoleplugin/consoleplugin_objects.go
OlivierCazade: Fixed, thanks

controllers/flowcollector_controller.go
OlivierCazade: I added detection to the controller, the controller looks for console group.

Note that this check is only done at startup, should be watch for created groups in case the console is deployed after the operator ?

jotak: I'd say it's ok as it is, the console being created after the NOO would really be a corner case IMO.

jotak:
/lgtm
Thanks!
@memodi or @jechen0648 do you want to test before or after merge?
It works with this other PR on the new console plugin repo: https://github.com/netobserv/network-observability-console-plugin/pull/1

The two PRs can be tested together by building the operator from PR as usual, and configuring the FlowCollector with the plugin image name:

```yaml
  consolePlugin:
    image: 'quay.io/ocazade0/network-observability-console-plugin:latest'
    imagePullPolicy: IfNotPresent
    port: 9001
```

But not sure if this image is up to date, @OlivierCazade can you confirm or provide an up-to-date image ?

After the plugin PR is merged, we will be able to setup post-merge CI / image generation as we did for the goflow image.

PR #15, opened 2021-11-09
NETOBSERV-82: Enforcing crd validation to only have one instance per cluster

OlivierCazade:
Using the validation scheme to only accept default as the CRD name, only one instance of the CRD can be created per cluster.

Jira link : https://issues.redhat.com/browse/NETOBSERV-82

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -16,6 +16,14 @@ patchesStrategicMerge:
 #- patches/cainjection_in_flowcollectors.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
+
+patches:
+- target:
+    kind: CustomResourceDefinition
+    name: flowcollectors.flows.netobserv.io
+  path: patches/singleton_in_flowcollectors.yaml
+
+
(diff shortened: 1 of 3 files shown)


mariomac (approved):
LGTM. Just a minor comment: is `default` the common name for singleton instances in RedHat? E.g. I see that the default network is named `cluster`.

OlivierCazade:
> LGTM. Just a minor comment: is `default` the common name for singleton instances in RedHat? E.g. I see that the default network is named `cluster`.

Thanks, I could not find official policy but cluster looks like the most used name, I have switched to it.

mariomac (approved):
LGTM

jotak:
I get a validation error when applying the CR:

```
error: error validating "./config/samples/flows_v1alpha1_flowcollector.yaml": error validating data: ValidationError(FlowCollector): unknown field "spec" in io.netobserv.flows.v1alpha1.FlowCollector; if you choose to ignore these errors, turn validation off with --validate=false
```

is it just me?

jotak (changes requested):
looks like all the `openAPIV3Schema` content is replaced with what's in the patch 😬

PR #16, opened 2021-11-09
[NETOBSERV-34] [NETOBSERV-87] ConfigMap fields

jpinsonneau:
Binded FlowCollectorGoflowKube port + FlowCollectorLoki fields to ConfigMap
Set static Labels

Usage of structs & json.Marshal makes the code easy to read / update

### DIFF
--- a/controllers/goflowkube/goflowkube_objects.go
+++ b/controllers/goflowkube/goflowkube_objects.go
@@ -91,14 +91,16 @@ func buildMainCommand(desired *flowsv1alpha1.FlowCollectorGoflowKube) string {
 	return fmt.Sprintf(`/goflow-kube -loglevel "%s" -config %s/%s`, desired.LogLevel, configPath, configFile)
 }
 
-func buildConfigMap(desired *flowsv1alpha1.FlowCollectorGoflowKube, ns string) *corev1.ConfigMap {
+func buildConfigMap(desiredGoflowKube *flowsv1alpha1.FlowCollectorGoflowKube,
+	desiredLoki *flowsv1alpha1.FlowCollectorLoki, ns string) *corev1.ConfigMap {
 	data := fmt.Sprintf(`
 {
 listen: "netflow://:%d",
 loki: {
-	labels: ["SrcNamespace","SrcWorkload","DstNamespace","DstWorkload"]
+	url: %s,
+	labels: ["SrcNamespace","SrcWorkload","DstNamespace","DstWorkload"],
 }}
-`, desired.Port)
+`, desiredGoflowKube.Port, desiredLoki.URL)
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      configMapName,
--- a/controllers/goflowkube/goflowkube_objects.go
+++ b/controllers/goflowkube/goflowkube_objects.go
@@ -1,6 +1,7 @@
 package goflowkube
 
 import (
+	"encoding/json"
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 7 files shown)


jotak:
oh you're right i forgot about propagating the loki config to goflow-kube 🤦
All fields in `FlowCollectorLoki` should actually be propagated to this config : https://github.com/netobserv/goflow2-kube-enricher/blob/main/pkg/config/config.go#L23-L51

jpinsonneau:
> oh you're right i forgot about propagating the loki config to goflow-kube facepalm All fields in `FlowCollectorLoki` should actually be propagated to this config : https://github.com/netobserv/goflow2-kube-enricher/blob/main/pkg/config/config.go#L23-L51

Sure I'll do that. What about the missing ones in `FlowCollectorLoki` ?

I saw at least `Labels []string`  which was hardcoded but there is also `Timeout time.Duration`, `IgnoreList []string`, `TimestampLabel model.LabelName` & `TimestampScale time.Duration` missing in [FlowCollectorLoki](https://github.com/netobserv/network-observability-operator/blob/d13375e0466eff534b99cb5fc0f31c29c138d183/api/v1alpha1/flowcollector_types.go#L104). Do we keep them as internal options ?

jotak:
> 
> Sure I'll do that. What about the missing ones in `FlowCollectorLoki` ?
> 
> I saw at least `Labels []string` which was hardcoded but there is also `Timeout time.Duration`, `IgnoreList []string`, `TimestampLabel model.LabelName` & `TimestampScale time.Duration` missing in [FlowCollectorLoki](https://github.com/netobserv/network-observability-operator/blob/d13375e0466eff534b99cb5fc0f31c29c138d183/api/v1alpha1/flowcollector_types.go#L104). Do we keep them as internal options ?

Hmm `Timeout` was probably forgotten by mistake, but the other ones are left aside on purpose (basically, it's because goflow-kube is more generic, in theory it can be used as a collector for any IPFIX supplier, whereas in the operator it is narrowed down to the OVS use case which allows us to make a couple of assumptions, for instance regarding the timestamp labels and so on)

So yeah, `Timeout` must be added.

controllers/goflowkube/goflowkube_objects.go
```diff
@@ -93,22 +111,37 @@ func buildMainCommand(desired *flowsv1alpha1.FlowCollectorGoflowKube) string {
...
-{
-listen: "netflow://:%d",
-loki: {
-	url: %s,
-	labels: ["SrcNamespace","SrcWorkload","DstNamespace","DstWorkload"],
-}}
-`, desiredGoflowKube.Port, desiredLoki.URL)
+
+	configStr := `{}`
+	config := &ConfigMap{
```
jotak, line 116: maybe we could use the model from goflow-kube (importing `"github.com/netobserv/goflow2-kube-enricher/pkg/config"`) (if that doesn't pull too many dependencies) ?

jpinsonneau: That's fair but it will need conversions or at least change for `FlowCollectorLoki` types :
- time.Duration / metav1.Duration
- model.LabelSet / map[string]string)
- int64 & int32 / int

jotak: (never mind, let's keep duplicating the model to avoid any dependency hell!)

mariomac (approved):

jpinsonneau:
sorry @mariomac, I've added `PrintOutput` flag for [NETOBSERV-87](https://issues.redhat.com/browse/NETOBSERV-87) after your review

It's exactly the same way as other flags so I prefer to make a single PR

mariomac (approved):

PR #17, opened 2021-11-10
Improve documentation

jotak:
- Using existing images built by CI
- note on pull policy
- applying CR
- configuriing OVS

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,55 @@ This operator uses the [Go Operator Framework](https://sdk.operatorframework.io/
 
 The operator is NOT functional at this stage.
 
-## Build / push / deploy
+By default, the operator is deployed in namespace "network-observability".
+
+## Deploy an existing image
+
+Images are built and pushed through CI to [quay.io](https://quay.io/repository/netobserv/network-observability-operator?tab=tags). To refer to the latest version of the `main` branch, use `IMG=quay.io/netobserv/network-observability-operator:main` or simply `VERSION=main`. To refer to older versions, use the commit short-SHA as the image tag.
+
+E.g. to deploy the latest build of branch `main`:
 
 ```bash
-# With default image to quay.io/netobserv
-make image-build image-push deploy
-
-# With custom image
-IMG="docker.io/myuser/netobserv:latest" make image-build image-push deploy
+VERSION=main make deploy
 ```
 
-By default, the operator is deployed in namespace "network-observability".
+## Build / push / deploy
+
+The repository `quay.io/netobserv/network-observability-operator` is only writable by the CI, so you need to use another repository (such as your own one) if you want to use your own build.
+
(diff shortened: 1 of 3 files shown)


jpinsonneau (approved):
LGTM

Should we mention usage of kustomize here or in general documentation ?

jotak:
> Should we mention usage of kustomize here or in general documentation ?

Like what? Changing the target namespace, and so on? (I don't know, I never change these values, but if it's useful for you it probably can be useful also for someone else)

OlivierCazade (approved):

README.md
```diff
@@ -8,17 +8,55 @@ This operator uses the [Go Operator Framework](https://sdk.operatorframework.io/
...
 
-By default, the operator is deployed in namespace "network-observability".
+### Enabling OVS IPFIX export
+
+This part will eventually be done automatically by the operator, but for the time being it requires manual intervention.
+
+#### On KIND
+
+```bash
+GF_IP=`kubectl get svc goflow-kube -ojsonpath='{.spec.clusterIP}'` && echo $GF_IP
```
jpinsonneau, line 48: you need to add -n network-observability here

README.md
```diff
@@ -8,17 +8,55 @@ This operator uses the [Go Operator Framework](https://sdk.operatorframework.io/
...
+kubectl set env daemonset/ovnkube-node -c ovnkube-node -n ovn-kubernetes OVN_IPFIX_TARGETS="$GF_IP:2055"
+```
+
+#### On OpenShift
+
+In OpenShift, a difference with the upstream `ovn-kubernetes` is that the flows export config is managed by the `ClusterNetworkOperator`.
+
+```bash
+GF_IP=`oc get svc goflow-kube -ojsonpath='{.spec.clusterIP}'` && echo $GF_IP
```
jpinsonneau, line 57: same here I guess ?

jpinsonneau (dismissed):

README.md
```diff
@@ -8,17 +8,55 @@ This operator uses the [Go Operator Framework](https://sdk.operatorframework.io/
...
+```
+
+Note, the default image pull policy is `IfNotPresent`, so if you previously deployed the operator on a cluster and then create another build with the same image name/tag, it won't be pulled in the cluster registry. So you need either to provide a different image name/tag for every build, or modify [manager.yaml](./config/manager/manager.yaml) to set `imagePullPolicy: Always`, then re-deploy.
+
+Then, you can deploy a custom resource, e.g.:
 
-# With custom image
-IMG="docker.io/myuser/netobserv:latest" make image-build image-push deploy
+```bash
+kubectl apply -f ./config/samples/flows_v1alpha1_flowcollector.yaml
```
memodi, line 38: need to apply namespace here "-n network-observability"

jotak: actually, here there's no need to provide a namespace because this resource is cluster-scope, not namespace-scope

memodi (approved):

mariomac (approved):

PR #20, opened 2021-11-18
Configure such as the default case works out of the box

jotak:
- Use images tagged "main" by default with `make deploy`
- Fix wrong default kubebuilder images
- Update readme


jotak:
With this PR, running just `make deploy` should deploy everything merged in `main`

jpinsonneau (approved):
LGTM

PR #21, opened 2021-11-18
NETOBSERV-40 Create grafana dashboard

jotak:
This first dashboard shows:
- Volumetry per namespace
- Volumetry per workload
- Raw logs

Capture d’écran de 2021-11-18 16-41-27


memodi:
tested and installed loki-grafana via helm, was able to view dashboards by importing these files.

/lgtm
/label qe-approved

PR #22, opened 2021-11-22
[NETOBSERV-66] Bug fixes: deploying as a DaemonSet and removing configmap

mariomac:
Fixes two issues:
* When deploying Goflow as a Daemonset, it returns an "unexpected kind DaemonSet" error.
* During the undeployment, if the ovs-flows-config map did not exist, it keeped retrying without finalizing the undeployment.

Jira issue: https://issues.redhat.com/browse/NETOBSERV-66

### DIFF
--- a/controllers/flowcollector_controller_test.go
+++ b/controllers/flowcollector_controller_test.go
@@ -4,12 +4,12 @@ import (
 	"net"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jotak:
/lgtm
thanks @mariomac

PR #26, opened 2021-11-24
NETOBSERV-91 crd generate

jpinsonneau:
Call [crdoc](https://github.com/fybrik/crdoc) on `make generate`
output docs/crd.md

Is it ok to everyone to call `go install fybrik.io/crdoc@latest` from the makefile or do you have a better way ?


README.md
```diff
@@ -77,3 +77,4 @@ spec:
 ## Resources
 
 - [Advanced topics](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/) (generic / operator framework)
+- [API Reference](https://github.com/netobserv/network-observability-operator/docs/crd.md)
```
jotak, line 80: I don't think this link will work, it would be something like `.../blob/main/docs/crd.md` instead
But even better would be to make a relative link, so that it's consistent across branches
I think `[API Reference](./docs/crd.md)` would work

jotak: Also can you rename `crd.md` to `FlowCollector.md` ? as there could be more than one CRD

jpinsonneau: sure, here we go

mariomac (approved):

Makefile
```diff
@@ -87,6 +87,8 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	GOFLAGS="" go install fybrik.io/crdoc@latest
```
mariomac, line 90: Maybe we should do something like:
```
test -f $(go env GOPATH)/bin/crdoc || GOFLAGS="" go install fybrik.io/crdoc@latest
```

jpinsonneau: I guess `go install` already does this checks and also manage updates if available since I pointed @latest

PR #28, opened 2021-11-24
NETOBSERV-83 : Namespace configuration

OlivierCazade:
Added several fields in the CRD :
- namespace to deploy goflow-kube and console-plugin pods
- name and namespace of the CNO configmap

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -27,6 +27,11 @@ import (
 type FlowCollectorSpec struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
+	//+kubebuilder:default:=""
+	// Namespace  where console plugin and goflowkube pods are going to be deployed.
+	// If empty, the namespace of the operator is going to be used
+	Namespace string `json:"namespace,omitempty"`
+
 	// IPFIX contains IPFIX-related settings for the flow reporter
 	IPFIX FlowCollectorIPFIX `json:"ipfix,omitempty"`
 
@@ -38,6 +43,9 @@ type FlowCollectorSpec struct {
 
 	// ConsolePlugin contains settings related to the console dynamic plugin
 	ConsolePlugin FlowCollectorConsolePlugin `json:"consolePlugin,omitempty"`
+
+	// CNO contains settings related to the cluster network operator
+	Cno ClusterNetworkOperator `json:"cno,omitempty"`
 }
 
 // FlowCollectorIPFIX defines the desired IPFIX state of FlowCollector
@@ -192,6 +200,15 @@ type FlowCollectorConsolePlugin struct {
 	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"`
+}
+
+// CNO defines the desired configuration related to the Cluster Network Configuration
+type ClusterNetworkOperator struct {
+	// Important: Run "make generate" to regenerate code after modifying this file
+
+	//+kubebuilder:default:=openshift-network-operator
+	// Namespace  where the configmap is going to be deployed.
+	Namespace string `json:"namespace,omitempty"`
 }
 
 // FlowCollectorStatus defines the observed state of FlowCollector
--- a/controllers/consoleplugin/consoleplugin_reconciler.go
+++ b/controllers/consoleplugin/consoleplugin_reconciler.go
@@ -18,15 +18,14 @@ import (
 type Reconciler struct {
 	client.Client
 	SetControllerReference func(client.Object) error
-	OperatorNamespace      string
+	Namespace              string
 }
 
 const pluginName = "network-observability-plugin"
 
 // Reconcile is the reconciler entry point to reconcile the current plugin state with the desired configuration
 func (r *Reconciler) Reconcile(ctx context.Context, desired *flowsv1alpha1.FlowCollectorConsolePlugin) error {
-	// Check if goflow-kube already exists, as a deployment or as a daemon set
-	nsname := types.NamespacedName{Name: pluginName, Namespace: r.OperatorNamespace}
+	nsname := types.NamespacedName{Name: pluginName, Namespace: r.Namespace}
 
 	// Get existing objects
 	oldDepl, err := r.getObj(ctx, nsname, &appsv1.Deployment{})
@@ -40,32 +39,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, desired *flowsv1alpha1.FlowC
 
 	// First deployment, creating permissions and container plugin
 	if oldDepl == nil && oldSVC == nil {
-		rbac := buildRBAC(r.OperatorNamespace)
+		rbac := buildRBAC(r.Namespace)
 		for _, rbacObject := range rbac {
 			r.createOrUpdate(ctx, nil, rbacObject)
 		}
-		consolePlugin := buildConsolePlugin(desired, r.OperatorNamespace)
+		consolePlugin := buildConsolePlugin(desired, r.Namespace)
 		r.createOrUpdate(ctx, nil, consolePlugin)
 	}
 
 	// Check if objects need update
-	if oldDepl == nil || deploymentNeedsUpdate(oldDepl.(*appsv1.Deployment), desired) {
-		newDepl := buildDeployment(desired, r.OperatorNamespace)
+	if oldDepl == nil || deploymentNeedsUpdate(oldDepl.(*appsv1.Deployment), desired, r.Namespace) {
+		newDepl := buildDeployment(desired, r.Namespace)
 		r.createOrUpdate(ctx, oldDepl, newDepl)
 	}
-	if oldSVC == nil || serviceNeedsUpdate(oldSVC.(*corev1.Service), desired) {
-		newSVC := buildService(desired, r.OperatorNamespace)
+	if oldSVC == nil || serviceNeedsUpdate(oldSVC.(*corev1.Service), desired, r.Namespace) {
+		newSVC := buildService(desired, r.Namespace)
 		r.createOrUpdate(ctx, oldSVC, newSVC)
 	}
 	return nil
 }
 
-func deploymentNeedsUpdate(depl *appsv1.Deployment, desired *flowsv1alpha1.FlowCollectorConsolePlugin) bool {
+func deploymentNeedsUpdate(depl *appsv1.Deployment, desired *flowsv1alpha1.FlowCollectorConsolePlugin, ns string) bool {
+	if depl.Namespace != ns {
+		return true
+	}
 	return containerNeedsUpdate(&depl.Spec.Template.Spec, desired) ||
 		*depl.Spec.Replicas != desired.Replicas
 }
 
-func serviceNeedsUpdate(svc *corev1.Service, desired *flowsv1alpha1.FlowCollectorConsolePlugin) bool {
+func serviceNeedsUpdate(svc *corev1.Service, desired *flowsv1alpha1.FlowCollectorConsolePlugin, ns string) bool {
+	if svc.Namespace != ns {
+		return true
+	}
 	for _, port := range svc.Spec.Ports {
 		if port.Port == desired.Port && port.Protocol == "TCP" {
 			return false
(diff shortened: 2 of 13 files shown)


controllers/consoleplugin/consoleplugin_reconciler.go
```diff
@@ -145,3 +161,16 @@ func (r *Reconciler) getObj(ctx context.Context, nsname types.NamespacedName, ob
...
+func (r *Reconciler) getNamespaceName(desired *flowsv1alpha1.FlowCollectorConsolePlugin) string {
+	if desired.Namespace != "" {
+		return desired.Namespace
+	} else {
+		return r.OperatorNamespace
+	}
+}
+
+func (r *Reconciler) namespaceExist(ctx context.Context, nsName string) (bool, error) {
+	ns, err := r.getObj(ctx, types.NamespacedName{Name: nsName}, &corev1.Namespace{})
```
OlivierCazade, line 174: Is there a better way to query a namespace using the controller-runtime client ?

Using a NamespacedName type to query a Namespace feel weird but it works and I couldn't find anything better in the API doc.

api/v1alpha1/flowcollector_types.go
```diff
@@ -93,6 +96,11 @@ type FlowCollectorGoflowKube struct {
 	// ImagePullPolicy is the Kubernetes pull policy for the image defined above
 	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
 
+	//+kubebuilder:default:=""
+	// Namespace  where goflowkube pod is going to be deployed.
+	// If empty, the namespace of the operator is going to be used
+	Namespace string `json:"namespace,omitempty"`
```
jotak, line 102: I wasn't thinking there would be two separate fields for goflow-kube and plugin's namespaces, do you think there can be a real need? I'm asking, because it's one more field to configure for the user, and the easier it is to configure, the better is the user experience...

controllers/consoleplugin/consoleplugin_reconciler.go
mariomac: I don't see anything wrong with that implementation, unless someone else knows any way even better.

api/v1alpha1/flowcollector_types.go
```diff
@@ -194,6 +207,19 @@ type FlowCollectorConsolePlugin struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"`
 }
 
+// CNO defines the desired configuration related to the Cluster Network Configuration
+type ClusterNetworkOperator struct {
+	// Important: Run "make generate" to regenerate code after modifying this file
+
+	//+kubebuilder:default:=ovs-flows-config
+	// Namespace  where the configmap is going to be deployed.
+	ConfigmapName string `json:"configmapName,omitempty"`
```
jotak, line 216: Here also, not sure if there's a real need to configure the config map name that we generate (moreover if it's configurable here then it must also be configurable in the CNO resource, not sure if that's the case, @mariomac is that something you did?)

api/v1alpha1/flowcollector_types.go
```diff
@@ -194,6 +207,19 @@ type FlowCollectorConsolePlugin struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"`
 }
 
+// CNO defines the desired configuration related to the Cluster Network Configuration
+type ClusterNetworkOperator struct {
+	// Important: Run "make generate" to regenerate code after modifying this file
+
+	//+kubebuilder:default:=ovs-flows-config
+	// Namespace  where the configmap is going to be deployed.
```
jotak, line 215: btw (if we keep this field) the description is inaccurate

api/v1alpha1/flowcollector_types.go
```diff
@@ -38,6 +38,9 @@ type FlowCollectorSpec struct {
 
 	// ConsolePlugin contains settings related to the console dynamic plugin
 	ConsolePlugin FlowCollectorConsolePlugin `json:"consolePlugin,omitempty"`
+
+	// CNO contains settings related to the console dynamic plugin
```
jotak, line 42: bad copy-paste of the description :)

controllers/consoleplugin/consoleplugin_reconciler.go
jotak: I don't know, I guess this is fine

controllers/consoleplugin/consoleplugin_test.go
```diff
@@ -91,17 +95,22 @@ func TestServiceUpdateCheck(t *testing.T) {
 
 	//equals specs
 	serviceSpec, containerConfig := getServiceSpecs()
-	assert.Equal(serviceNeedsUpdate(&serviceSpec, &containerConfig), false)
+	// assert.Equal(serviceNeedsUpdate(&serviceSpec, &containerConfig, testNamespace), false)
```
jotak, line 98: shouldn't that pass?

controllers/flowcollector_controller.go
```diff
@@ -29,28 +29,29 @@ import (
...
-			operatorNamespace, cnoNamespace, ovsFlowsConfigMapName),
-		consoleEnabled: false,
+		Client:              client,
+		Scheme:              scheme,
+		ovsConfigController: nil,
+		consoleEnabled:      false,
 	}
 }
 
+//+kubebuilder:rbac:groups=core,resources=namespaces;configmaps,verbs=get;list;watch;create;update;patch;delete
```
jotak, line 54: there's already rbac on group `core` two lines below with the same verbs, i guess you can merge them

controllers/consoleplugin/consoleplugin_test.go
OlivierCazade: Yes, I commented it when debugging and forgot t uncomment it.

api/v1alpha1/flowcollector_types.go
OlivierCazade: I don't understand, was this constant referred in another repo ?

controllers/consoleplugin/consoleplugin_reconciler.go
jpinsonneau: can't find anything better in the client code, I guess this is the right way :+1:

api/v1alpha1/flowcollector_types.go
OlivierCazade: I thought the CRD would be more readable with namespace under each section than at the top level.

About the needs, what exactly is the needs of this task ? I mean, if we have need to run kubeflow and the plugin in a different namespace than the operator (while there can only be one kubeflow per operator), we may have needs to run both in differents namespaces.

api/v1alpha1/flowcollector_types.go
jotak: yes this config map is later watched by the CNO, see this story: https://issues.redhat.com/browse/NETOBSERV-66

jotak: so i guess the name matters here

api/v1alpha1/flowcollector_types.go
OlivierCazade: 🤦

api/v1alpha1/flowcollector_types.go
OlivierCazade: 🤦

jotak:
One thing I just realize while code reviewing.. what happens if we have a previous deployment then we update the CR to change the namespace? If I read correctly, there's nothing here to delete all previously deployed resources that were in the previous namespace, correct?
If so, I'm not sure how to fix that. Afaik we can't get the previous values of the CR (and we should keep operator pods stateless so than they can recover from a pod kill/restart), so we don't know what the previous namespace was (?). Any idea? (cc @mariomac @jpinsonneau )

api/v1alpha1/flowcollector_types.go
jotak: My bad, the story isn't clear about that. The initial need came after discussing with the operator framework team and I was told it could be seen as a bad practice to deploy stuff in the operator's namespace. Think about openshift for instance, where (I think) there a unique namespace where many operators are deployed, but we don't want this namespace to be polluted by workloads that are not operators themselves.
I should have made it clear in the task description, sorry about that.

But then, I don't think we need this level of detail of deploying each component in a separate namespace.

jpinsonneau:
> One thing I just realize while code reviewing.. what happens if we have a previous deployment then we update the CR to change the namespace? If I read correctly, there's nothing here to delete all previously deployed resources that were in the previous namespace, correct? If so, I'm not sure how to fix that. Afaik we can't get the previous values of the CR (and we should keep operator pods stateless so than they can recover from a pod kill/restart), so we don't know what the previous namespace was (?). Any idea? (cc @mariomac @jpinsonneau )

That's right ! I would say we need to create a secret or configmap keeping the namespace info and compare it on each CR value change.
If it's different, then try to delete ressources from the old namespace and update the new value

api/v1alpha1/flowcollector_types.go
OlivierCazade: Ok, thanks for the clarifications.

jotak:
about the cleanup: we can create a follow-up task or bug and not block this PR

controllers/consoleplugin/consoleplugin_test.go
OlivierCazade: Side question here, the reason I commented this check is because I wanted to see the result of the following part of the test.

Should we split this test into different one ?

jotak: yes go ahead if it's better :-)

api/v1alpha1/flowcollector_types.go
mariomac: Yeah I did this... maybe I was being too configurator. Feel free to remove it if you think it's better to just put a constant.

mariomac:
about the namespace cleanup: is this something we need to care of? I guess this same situation would happen with any other resource, derived from a bad practice: a customer not deleting first the old resource before recreating it with a new namespace.

... or maybe I misunderstood @jotak ?

api/v1alpha1/flowcollector_types.go
jotak: What I mean is, here https://github.com/openshift/cluster-network-operator/pull/1231/files#diff-993ef112f8af6a1b2aaf1b7f141a10306d0420db5b56a272d7365a8bacd82920R25 it's hardcoded, so if it's hardcoded on one side, there's no need to make it configurable on the other end, right?
It could give a wrong impression that the user can set whatever he wants, but in reality changing this would break on the CNO side

jotak:
@mariomac what the reconcile loop does here (what we've coded) is checking for existing object and, depending on the situation, delete, update or create objects. For instance when we modify the CR to use DaemonSet instead of Deployment, we will look for any existing Service and Deployment and delete them, before creating the DaemonSet. But to look for the existing object, we search them by name+namespace - previously using the operator namespace because it was where they were always deployed.

Now that the namespace is configurable, this will not work anymore if the namespace defined in CR is modified.

Another solution could be to make this field immutable (ie. once it's set, it cannot be modified; the user would have to delete the CR then re-create it; CR deletion cleans all associated objects, so that would work). (But I don't know how to make a field immutable, it doesn't seem to be supported (yet) out of the box by kube-builder, though workarounds seem to exist)

jotak:
> That's right ! I would say we need to create a secret or configmap keeping the namespace info and compare it on each CR value change. If it's different, then try to delete ressources from the old namespace and update the new value

Yes that's an option, although it adds some burden with yet another object to be created and managed. I'm still looking for something more direct if it's possible :-)

jotak:
bug created: https://issues.redhat.com/browse/NETOBSERV-95

jpinsonneau:
> > That's right ! I would say we need to create a secret or configmap keeping the namespace info and compare it on each CR value change. If it's different, then try to delete ressources from the old namespace and update the new value
> 
> Yes that's an option, although it adds some burden with yet another object to be created and managed. I'm still looking for something more direct if it's possible :-)

The configmap `goflow-kube-config` is unique per operator and should stay in the operator namespace right ? So we only configure goflow & plugin namespace in a single parameter.

The only issue I can see is if we plan to deploy multiple operators in the same namespace.

jotak:
> The configmap `goflow-kube-config` is unique per operator and should stay in the operator namespace right ? So we only configure goflow & plugin namespace in a single parameter.

Yes it's unique but no it doesn't stay in the oeprator namespace, it's actually bound to the goflow-kube deployment so it should be in goflow's namespace

But as said in chat, I imagine we should use the CR's status for that kind of thing (and the CR is unique + cluster-scope)

jotak:
/lgtm
thanks!

PR #29, opened 2021-11-25
NETOBSERV-90: enable deployment as Daemonset

mariomac:
- Adding a HostPort to the container specification in the DaemonSet
- Setting the proper RBAC permissions
- Enable the deployment in the master node

### DIFF
--- a/controllers/flowcollector_controller.go
+++ b/controllers/flowcollector_controller.go
@@ -59,6 +59,7 @@ func NewFlowCollectorReconciler(client client.Client, scheme *runtime.Scheme) *F
 //+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors/finalizers,verbs=update
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,resourceNames=hostnetwork,verbs=use
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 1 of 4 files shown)


controllers/flowcollector_controller.go
```diff
@@ -172,7 +173,6 @@ func (r *FlowCollectorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.DaemonSet{}).
-		Owns(&corev1.Service{}).
```
OlivierCazade, line 175: Why removing this ?

OlivierCazade (approved):
LGTM, my only comment is more a question than a comment

controllers/flowcollector_controller.go
mariomac: It is duplicated 2 lines below.

OlivierCazade: Thanks

controllers/goflowkube/goflowkube_objects.go
```diff
@@ -72,18 +71,33 @@ func buildDaemonSet(desired *flowsv1alpha1.FlowCollectorGoflowKube, ns string) *
...
 	}
 }
 
-func buildPodTemplate(desired *flowsv1alpha1.FlowCollectorGoflowKube) *corev1.PodTemplateSpec {
+func buildPodTemplate(desired *flowsv1alpha1.FlowCollectorGoflowKube, forDaemonSet bool) corev1.PodTemplateSpec {
 	cmd := buildMainCommand(desired)
-	return &corev1.PodTemplateSpec{
+	var ports []corev1.ContainerPort
+	var tolerations []corev1.Toleration
+	if forDaemonSet {
```
jpinsonneau, line 83: can't we use `desired.Kind == constants.DaemonSetKind` here instead of an additionnal parameter ?

jpinsonneau (changes requested):

controllers/goflowkube/goflowkube_objects.go
mariomac: Sure. Changing

jpinsonneau (approved):
LGTM

PR #32, opened 2021-11-29
NETOBSERV-88: restart pods on configmap change (bugfix)

mariomac:
It adds a new annotation to the pods that contains a digest of the
the goflow-kube-config configmap contents. When that configmap changes,
the Pod template inside the Daemonset or Deployment is updated so
Kubernetes will force the update of the goflow-kube pods.

Demo:

https://user-images.githubusercontent.com/939550/143897920-5a794abc-4e2b-48e5-9598-cc27f5d0191f.mp4

### DIFF
--- a/controllers/goflowkube/goflowkube_objects.go
+++ b/controllers/goflowkube/goflowkube_objects.go
@@ -3,6 +3,8 @@ package goflowkube
 import (
 	"encoding/json"
 	"fmt"
+	"hash/fnv"
+	"strconv"
 
 	flowsv1alpha1 "github.com/netobserv/network-observability-operator/api/v1alpha1"
 	"github.com/netobserv/network-observability-operator/controllers/constants"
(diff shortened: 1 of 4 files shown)


controllers/goflowkube/goflowkube_objects.go
```diff
@@ -20,6 +22,9 @@ const configVolume = "config-volume"
 const configPath = "/etc/goflow-kube"
 const configFile = "config.yaml"
 
+// annotation name to facilitate pod restart after any external configuration change
+const podConfigurationDigest = "goflow_kube_config_digest"
```
mariomac, line 26: Should we use some kind of qualified name for the annotation? e.g. `netobserv.openshift.org/goflow_kube_config_digest`

jotak: yes actually you could reuse the domain from the CRD: `flows.netobserv.io` or just `netobserv.io`

jotak (approved):

jpinsonneau (approved):

PR #33, opened 2021-11-30
Adding namespace field to example config file

OlivierCazade:


jotak:
thanks @OlivierCazade  can you also add the CNO namespace field?

jotak:
thanks @OlivierCazade !
/lgtm

PR #34, opened 2021-11-30
NETOBSERV-95 Cleaning up old namespaces

jotak:
https://issues.redhat.com/browse/NETOBSERV-95

Use Status to store current namespace used for deployments and detect
namespace change
Delete all namespace-scoped data previously deployed when namespace is
changed


jotak:
I'm opening as draft .. it seems it doesn't cover all the cases yet 
Also, there's some repeated patterns between the goflow reconciler and the plugin reconciler, I'll see if it makes sense to share more code between the two of them

jotak:
Pfuu that was a long refactoring (sorry for the reviewers! 😬 ) 

But I think it's worth it as it also fixes a couple of bugs and I added a bunch of new tests along that road. 

- Create a ClientHelper for shared Create/Update functions
- Create a NamespacedObjectManager that is like a helper to manage
  (fetch, delete...) namespace-scoped objects, taking care of deleting
everything needed when a namespace changes
- Many new tests (for namespace cleanup but also for console plugin)
- Discovered a couple of bugs that are now fixed: invalid updates on
  some objects like ConsolePlugin or Services...)
- A couple of errors were not reported back up to the controller, this
  is now normalized
- Remove Finalizer for the ovs configmap, it's now an owned resource,
  hence automatically garbage-collected

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/netobserv/network-observability-operator
-  newTag: main
+  newName: quay.io/jotak/network-observability-operator
+  newTag: v0.0.1-7
```
OlivierCazade: I think this change was not supposed to be commited.

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -3,15 +3,15 @@ kind: FlowCollector
...
   ipfix:
     cacheActiveTimeout: 60s
     cacheMaxFlows: 400
-    sampling: 100
+    sampling: 200
   goflowkube:
     kind: Deployment
     replicas: 1
-    port: 2055
+    port: 2056
```
OlivierCazade: Why this port changes ?
(Same question for the console plugin)

controllers/goflowkube/goflowkube_objects.go
```diff
@@ -183,21 +182,29 @@ func buildConfigMap(desiredGoflowKube *flowsv1alpha1.FlowCollectorGoflowKube,
...
+			Spec: corev1.ServiceSpec{
+				Selector: buildLabels(),
+				Ports: []corev1.ServicePort{{
+					Port:     desired.Port,
+					Protocol: corev1.ProtocolUDP,
+				}},
+			},
+		}
 	}
+	// In case we're updating an existing service, we need to build from the old one to keep immutable fields such as clusterIP
```
OlivierCazade, line 202: Is it safe to directly modify the old one ? Can we clone it before any modifications ?

As a user of the buildService function I would not expect the old one to be modified.

config/manager/kustomization.yaml
jotak: yes it was committed by mistake, same for the port changes in CR below, I reverted these changes

controllers/goflowkube/goflowkube_objects.go
jotak: good point, it's amended

OlivierCazade (approved):

jotak:
Fixed some issues after rebasing with conflicting PR (@mariomac loki URL config)

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -298,13 +298,18 @@ spec:
                 type: object
               namespace:
                 default: ""
-                description: Namespace  where console plugin and goflowkube pods are
+                description: Namespace where console plugin and goflowkube pods are
                   going to be deployed. If empty, the namespace of the operator is
                   going to be used
                 type: string
             type: object
           status:
```
jpinsonneau, line 306: Should we log a message explaining the state in the status ? 
Like "Creating pods to XXX namespace"
Or "Moving from XXX namespace to YYY namespace"

controllers/reconcilers/client_helper.go
```diff
@@ -0,0 +1,65 @@
...
+// CreateOwned is an helper function that creates an object, sets owner reference and writes info & errors logs
+func (c *ClientHelper) CreateOwned(ctx context.Context, obj client.Object) error {
+	log := log.FromContext(ctx)
+	err := c.SetControllerReference(obj)
+	if err != nil {
+		log.Error(err, "Failed to set controller reference")
+		return err
+	}
+	kind := reflect.TypeOf(obj).String()
+	log.Info("Creating a new "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
```
jpinsonneau, line 27: , kind ? 

It seems to be copy paste in a lot of logs

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
jotak: yes I can do that tomorrow

controllers/reconcilers/client_helper.go
jotak: I think it's as I intend to.. example of generated log:
```
2021-12-06T15:14:52.772+0100	INFO	controller-runtime.manager.controller.flowcollector	Creating a new *v1.ClusterRole	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "console-plugin-flowcollector", "namespace": "", "Namespace": "", "Name": "goflow-kube"}
```

jpinsonneau: I thought the output would be the same, sorry !

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
jpinsonneau: Perfect ! Thanks

jpinsonneau (approved):
LGTM

jotak:
Merged manually, QE will test post-merge

PR #37, opened 2021-12-14
Fix console proxy declaration

jotak:
This is fixing the console proxy declaration, so that it can work in "production mode" (ie. console not deployed on host with bridge).

Requires a fix in openshift-api (PR coming soon)


jotak:
(first commit in vendors update, second commit is code change)

jotak:
related fix in openshift-api: https://github.com/openshift/api/pull/1085

jotak:
/hold

Do not merge this one, merge this instead: https://github.com/netobserv/network-observability-operator/pull/38
We keep that one opened just for reference (it's referenced in BZ to illustrate the bug: https://bugzilla.redhat.com/show_bug.cgi?id=2032289)
My other PR includes an override of the API vendored

PR #38, opened 2021-12-14
Fix console proxy declaration, overriding vendor

jotak:
This is fixing the console proxy declaration, so that it can work in "production mode" (ie. console not deployed on host with bridge).

In second commit there's a vendor override so that it all works without having to wait for https://github.com/openshift/api/pull/1085


jpinsonneau (approved):
:heavy_check_mark:

jotak:
/approve

I'm merging fast, so that @stleerh can have the console plugin working

PR #40, opened 2021-12-17
Instructions to run bundle from a catalog source

mariomac:


README.md
```diff
@@ -56,10 +56,11 @@ This task should be automatically done by the CI/CD pipeline. However, if you wa
 bundle for local testing, you should execute the following commands:
 
 ```
-export USERNAME=<container-registry-username>
+export USER=<container-registry-username>
```
OlivierCazade, line 59: Just an idea here, may be we could make USER default to netobserv in the namespace and then having default value of IMG to quay.io/$USER/network-observability-operator

That way, users would only have to overload the USER var.

OlivierCazade (approved):
LGTM, my only comment is more an improvement idea not related to your PR.

Thanks !

README.md
mariomac: No strong opinion here, but I'm afraid that then we accidentally push images to the `netobserv` user in Quay.

OlivierCazade: I think, this is already the case if someone does a make image-push without any export.

Do we have permissions to push to this user ? I think only @jotak have enough permissions

mariomac: Me too (actually I pushed accidentally this morning 😓 ) but if it's only me, it's fine what you say.

jotak: Actually I'd like to keep admin rights on quay but remove write rights to prevent accidental push, I'll check if this is possible

PR #43, opened 2022-01-18
NETOBSERV-122: liveness probes for goflow-kube pod

mariomac:
Configures the `goflow-kube` deployment to use pod liveness probes.

This PR can't be merged until the related counterpart in goflow-kube is merged: https://github.com/netobserv/goflow2-kube-enricher/pull/19


config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -107,6 +107,14 @@ spec:
               goflowkube:
                 description: GoflowKube contains settings related to goflow-kube
                 properties:
+                  healthPort:
```
OlivierCazade, line 110: This changes should also be done to bundle/manifests/flows.netobserv.io_flowcollectors.yaml

OlivierCazade (approved):
LGTM, just a comment about crd definition, but may be we should create a test to ensure both files are equals.

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
mariomac: Do we need to add it manually? Didn't find any makefile target that regenerated those manifests

mariomac: Nevermind, I copied the additions manually.

PR #45, opened 2022-01-26
Update Grafana dashboard

jotak:
Add flows by namespaces + total flows metrics

Capture d’écran de 2022-01-26 15-53-41

### DIFF
--- a/config/samples/dashboards/Network Observability.json
+++ b/config/samples/dashboards/Network Observability.json
@@ -8,33 +8,297 @@
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
         "type": "dashboard"
       }
     ]
   },
   "editable": true,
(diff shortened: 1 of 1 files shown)


jotak:
cc @memodi I don't know if you use grafana, if so this might interest you

memodi:
> cc @memodi I don't know if you use grafana, if so this might interest you

@jotak Thanks,  I do use Grafana, having dashboard JSON will be very helpful. Just confirm, the flows/per second metric isn't reported yet, correct?

jotak:
> Just confirm, the flows/per second metric isn't reported yet, correct?

Actually yes it's already there, provided from Loki. It's not there as a prometheus metric though. But we can get it from Loki / LogQL to get the number of log entries (so, flows) per second

memodi:
> > Just confirm, the flows/per second metric isn't reported yet, correct?
> 
> Actually yes it's already there, provided from Loki. It's not there as a prometheus metric though. But we can get it from Loki / LogQL to get the number of log entries (so, flows) per second

that's awesome. Thanks for adding that in!

mariomac (approved):

jpinsonneau (approved):

PR #48, opened 2022-01-31
NETOBSERV-92 : Fixing tests coverage and adding coverage-report targets

OlivierCazade:
Without -coverpkg=./... option, code coverage is only done for the package beeing tested.

Before PR:
> total:														(statements)			41.5%

After:
> total:															(statements)			71.6%

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,13 @@ lint: prereqs
 	golangci-lint run ./...
 
 test: manifests generate envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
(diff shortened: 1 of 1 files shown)


OlivierCazade:
Could someone with Goland or Intellij check we have now the same coverage ?

May be the fix here is not enough.

mariomac (approved):

mariomac:
@OlivierCazade these are the reports for my Intellij IDEA run in master:

image


image

Strangely, my reported total coverage is now 41.5 🤦🏻 so I can't verify your numbers, but definitely they look much better and more consistent with the numbers I was getting some weeks ago.

mariomac:
Setting the coverpkg argument in intelliJ idea reports me more realistic coverage data:
image

OlivierCazade:
Thanks @mariomac  for the feedback.

I don't have the coverage per package but based on your screenshots, I have the same coverage for the following files :
- main.go
- flowcollector_controller.go
- flowcollector_objects.go

PR #51, opened 2022-02-04
NETOBSERV-107 Add version information

jotak:
JIRA https://issues.redhat.com/browse/NETOBSERV-107

version information is added in several places:
- via kustomize, as label on the operator (manager) itself
- through build ldflags, to the main package, logged for information
  (both in "make build" and "make image-build")
- in deployed components (goflow-kube and console plugin) as k8s label
  (note that the version here is taken from the images defined in CR, which can
be different from the operator version)
- lastly, I updated the helper target `make create-sample` so that it can use a custom version from env

Use the VERSION env in make targets to set a specific version
(e.g. `VERSION="0.1.0-test" IMAGE_TAG_BASE="quay.io/jotak/network-observability-operator" make image-build image-push deploy`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Build the manager binary
 FROM registry.access.redhat.com/ubi8/go-toolset:1.16.7-5 as builder
+ARG OPVERSION="unknown"
 
 WORKDIR /opt/app-root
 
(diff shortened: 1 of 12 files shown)


jotak:
FYI I had to upgrade kustomize but it seems that we need to manually remove it from our go bin path and `./bin/kustomize` to make it work

This PR is only a part of the changes needed for [NETOBSERV-107](https://issues.redhat.com/browse/NETOBSERV-107). More work is needed on CI side. Potentially also in goflow-kube / console plugin

PR #52, opened 2022-02-07
NETOBSERV-148 Added goflowkube port verifications

OlivierCazade:
My understanding is that we cannot check everything here with kubebuilder annotations.

So, I added a verification function.


OlivierCazade:
> /lgtm Should we add an "allow_port" or "deny_port" configuration list to override current list?

Yes, I thought about this but for now the only configurable channel we have is the flowcollector object, which in this case is not the right place because we would loose the benefit of the port restriction.

jpinsonneau (approved):
LGTM
I would suggest to keep port filtering static for now to avoid too much user config

PR #53, opened 2022-02-08
NETOBSERV-108 [4.10] Document publishing on operator hubs

jotak:
And update a few descriptions, and logo, in CSV

Note: I'm working on branch 4.10 now because we'll do the first release here, but then it will be cherry-picked to main


hack/release.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# NOTE: this release script is provided for help purpose, but you should carefully read it and understand what it does before running it.
+# Appropriate permissions will be required when pushing git tags / images
+# Adapt the variables below to match your local files, desired version and remote target.
+# Part of this workflow will be automated soon.
+
+path_noo="../network-observability-operator"
+path_gfk="../goflow2-kube-enricher"
+path_plg="../network-observability-console-plugin"
```
jpinsonneau, line 10: Maybe we can directly `git clone` these guys if folder doesn't exists and `pull` lasts changes otherwise ?

Who is going to use this script ?

jotak: yeah we could do that, but my intent was just to providing guidelines for releasing (I hesitate to include this script in the PR or just keep it for myself, but I thought it could be helpful if someone else needs to release before we move to a more automated process)

jotak: (so I didn't want to invest too much of my time on that :) )

jotak:
/hold
Something I need to fix: the default CR fields, which are used as default in the console form, aren't updated with the corresponding operator version (fields for goflow-kube and plugin images). I'll try to do something with kustomize.

jotak:
/unhold
$VERSION now inferred in default CRD values

hack/release.sh
jpinsonneau: sure, I understand. Thanks for the feedback !

bundle/manifests/network-observability-operator.clusterserviceversion.yaml
```diff
@@ -63,12 +63,57 @@ spec:
...
+
+    ```
+    GF_IP=`kubectl get svc goflow-kube -n network-observability -ojsonpath='{.spec.clusterIP}'` && echo $GF_IP
+    kubectl set env daemonset/ovnkube-node -c ovnkube-node -n ovn-kubernetes OVN_IPFIX_TARGETS="$GF_IP:2055"
+    ```
+
+    To turn it off, remove the `OVN_IPFIX_TARGETS` env from `daemonset/ovnkube-node`.
+
+    #### OpenShift 4.10 and above
+    No further action is required to enable IPFIX exports. If you want to turn off the exports, remove the `FlowCollector` cluster resource.
```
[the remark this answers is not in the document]
jotak: done

mariomac (approved):

PR #59, opened 2022-02-16
PoC Operators

jpinsonneau:
This PoC allow strimzi operator deployment + instance creation simply specifying operator name in FlowCollectorSpec.Operators

It create a Subscription using operator-framework api that will deploy the operator and then create the kafka instance using https://github.com/RedHatInsights/strimzi-client-go


controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,206 @@
...
+		},
+		Spec: lv1beta1.LokiStackSpec{
+			Size:              lv1beta1.SizeOneXExtraSmall,
+			ReplicationFactor: 1,
+			Storage: lv1beta1.ObjectStorageSpec{
+				Secret: lv1beta1.ObjectStorageSecretSpec{
+					Name: "test",
+				},
+			},
+			StorageClassName: "standard",
```
jpinsonneau, line 203: I'm not sure about that "standard" storage class name since I had it working on `cluster bot` and not on `openshift-install`

Anyway if you have `Pending` loki components you can either change that value in the code or create a "standard" storage class to the storage type you want like:
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: standard
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
reclaimPolicy: Retain
allowVolumeExpansion: true
volumeBindingMode: Immediate

```

OlivierCazade (commented):
Correct me if I am wrong here but my understanding is that each operator can only be deployed per cluster once.

Because of this, I think that the dependency is too strong :
- what happens if the operator is already deployed in the cluster
- if we deploy the operator and someone else create a second resource instance for it, this resource will get destroyed if we remove the flowcollector object

I think managing the instances like you did is great but the operator part raise many concerns.

And just for my understanding, did we try using [the operator dependency feature](https://docs.openshift.com/container-platform/4.9/operators/understanding/olm/olm-understanding-dependency-resolution.html)?

api/v1alpha1/flowcollector_types.go
```diff
@@ -45,6 +45,9 @@ type FlowCollectorSpec struct {
 
 	// CNO contains settings related to the cluster network operator
 	CNO ClusterNetworkOperator `json:"cno,omitempty"`
+
+	// Operators contains compatible operators names to install, comma separated E.g. "loki-operator,strimzi-kafka-operator,grafana-operator"
+	Operators string `json:"op,omitempty"`
```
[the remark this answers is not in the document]
jpinsonneau: yes sure, we can also add some extra config per operator like versions / endpoints / ports & so on ?

jpinsonneau:
> Correct me if I am wrong here but my understanding is that each operator can only be deployed per cluster once.

You can actually install multiple times the same operator in different namespaces. The only situation I didn't tested is if the operator is globally installed. Then maybe the subscription in our namespace will fail.

Here is an example with grafana operator:
image

> And just for my understanding, did we try using [the operator dependency feature](https://docs.openshift.com/container-platform/4.9/operators/understanding/olm/olm-understanding-dependency-resolution.html)?

That's look great ! Thanks for sharing. I will test that ASAP.
This could avoid managing `CatalogSource` and `Subscription` in our code but we will still need to create the instances of each operator right ?

OlivierCazade:
> > Correct me if I am wrong here but my understanding is that each operator can only be deployed per cluster once.
> 
> You can actually install multiple times the same operator in different namespaces. The only situation I didn't tested is if the operator is globally installed. Then maybe the subscription in our namespace will fail.
> 
> Here is an example with grafana operator: 

Ok, I did not know we could restrict an operator to some managed namespace. Thanks for this! This solve many of my concerns.

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
mariomac: In the meanwhile, maybe, instead of a comma-separated string, I'd use a string array.

Maybe I'd also use a more descriptive field name for the JSON/Yaml: `operators`? `dependencies`?

controllers/flowcollector_controller.go
```diff
@@ -120,6 +126,7 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	if r.consoleEnabled {
 		cpReconciler = consoleplugin.NewReconciler(clientHelper, ns, previousNamespace)
 	}
+	operatorsReconcilier := operators.NewReconcilier(ctx, clientHelper, ns)
```
mariomac, line 129: Nitpick:`Reconciler` instead of `Reconcilier` (for both variable and method names).

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,206 @@
...
+			Kind:       constants.OperatorGroup,
+			APIVersion: "operators.coreos.com/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.ObservabilityName,
+			Namespace: namespace,
+		},
+		Spec: ov1alpha1.CatalogSourceSpec{
+			DisplayName: constants.ObservabilityName,
+			Image:       "quay.io/jpinsonn/loki-operator-catalog:v0.0.1",
```
mariomac, line 45: I guess this will change but, should we anyway do it configurable?

mariomac (commented):
Pretty cool! I just dropped few comments for the moment when this PoC goes from PoC to product.

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
eranra: Maybe add a separate bool field for each operator (BTW: what are the defaults ?) 
If we deploy a stand-alone Loki, grafana etc. instance in network-observability the default should be probably to install the those in the namespace

controllers/flowcollector_controller.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net"
 	"strings"
 
+	//TODO replace ViaQ repo by grafana/loki/operator one when released
```
eranra, line 8: The operator is in https://github.com/grafana/loki/ ...  maybe this remark is not needed?

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,206 @@
+package operators
+
+import (
+	"encoding/json"
+	"fmt"
+
+	sv1b2 "github.com/RedHatInsights/strimzi-client-go/apis/kafka.strimzi.io/v1beta2"
+	lv1beta1 "github.com/ViaQ/loki-operator/api/v1beta1" //TODO replace ViaQ repo by grafana/loki/operator one when released
```
eranra, line 8: same ??? (use the grafana github place)

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,206 @@
...
+		Spec: gv1alpha1.GrafanaSpec{
+			Config: gv1alpha1.GrafanaConfig{
+				Security: &gv1alpha1.GrafanaConfigSecurity{
+					AdminUser:     "admin",
+					AdminPassword: "admin",
+				},
+			},
+		},
+	}
+}
```
eranra, line 183: make sense to add her at least data source for Loki ( maybe also Prometheus )

api/v1alpha1/flowcollector_types.go
eranra, line 50: @jpinsonneau can you add also Prometheus Operator so we will have the destination for the metrics?

@jotak @stleerh FYI ^^^

controllers/operators/operators_objects.go
eranra: @jpinsonneau maybe you can add this "as code" - this is similar to https://raw.githubusercontent.com/netobserv/documents/main/examples/zero-click-loki/1-storage.yaml to some level.

eranra (commented):
@jpinsonneau a question ... if I change the CR adding or removing an operator from the list ... with that operator also will be uninstalled ???

jpinsonneau:
> @jpinsonneau a question ... if I change the CR adding or removing an operator from the list ... with that operator also will be uninstalled ???

For now **no** since I wasn't sure about installing the operators in the same namespace.
If we all agree that we **force operator instances to be in the same namespace**, then yes we can remove them according to the list.

api/v1alpha1/flowcollector_types.go
jpinsonneau: sure, I'll take a look :+1:

controllers/flowcollector_controller.go
jpinsonneau: I'm looking for `grafana/loki/operator` not `loki` itself and it's [not yet released](https://pkg.go.dev/github.com/grafana/loki/operator)

controllers/operators/operators_objects.go
jpinsonneau: I would prefer to maintain our own `netobserv` catalog for dev / previews only and remove this for production

How do you see that ?

jotak: So maybe that should be commented out until it is released?

api/v1alpha1/flowcollector_types.go
jotak: One thing about loki, it may feel weird to the user to enable installing the operator but still having to populate the loki URL in a different section. BTW how does it work, if the user installs the loki operator and set a URL that doesn't match it, are we going to ignore the URL set and use the known one instead? Should we document that the URL field is ignored when the operator is installed as a dependency?

So what do you think, if the loki section looks like:

```yaml
  loki:
    installOperator: true
    # url: '' (ignored)
    # querierUrl: '' (ignored)
    batchWait: 1s
```

And then, we could introduce new similar section for grafana / kafka / prometheus with a field `installOperator` true or false.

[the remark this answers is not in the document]
jpinsonneau: will be managed by separated jsons
https://github.com/netobserv/network-observability-operator/pull/59#discussion_r814643994

jpinsonneau:
I tried to import all operators configs but there is [some limitations on CRD size](https://github.com/kubernetes/kubernetes/issues/82292) so we should add only needed configs

- Added prometheus (only `replicas` option for now)
- Added grafana `datasource` & allow dashboard from specified `configmap`
- Automatically create grafana route if `ingress: true` 
- Loki `storageClassName` configurable (`standard` for kubernetes / `gp2` for openshift) + secret name

The operators + instances will be created as soon as corresponding `instanceSpec` is set; else it will skip it. 
You can also specify extra objects like grafana datasource / dashboard.

Example:
```yaml
grafana:
  instanceSpec:
    config:
      security:
        admin_user: admin
        admin_password: admin
    ingress: true
  dashboardSpec:
      configMapRef:
        name: grafana-dashboard
        key: json
        optional: true
```

controllers/flowcollector_controller.go
eranra: ok

controllers/constants/constants.go
```diff
@@ -2,11 +2,25 @@
...
+	LokiOperator          = "loki-operator"
+	LokiOperatorNamespace = "network-observability"
+	OpenshiftOperatorNs   = "openshift-operators"
+	PrometheusOperator    = "prometheus"
+	SubscriptionKind      = "Subscription"
+	KafkaName             = "kafka"
+	GrafanaName           = "grafana"
+	LokiName              = "loki"
+	PrometheusName        = "prometheus"
+	ObservabilityName     = "network-observability"
```
jotak, line 20: Maybe use the operator name instead, "netobserv" ?

controllers/constants/constants.go
```diff
@@ -2,11 +2,25 @@
...
+	GoflowKubeName        = "goflow-kube"
```
jotak, line 21: goflow-kube is obsolete

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,290 @@
...
+func operatorGroup(name string, namespace string) *ov1.OperatorGroup {
+	return &ov1.OperatorGroup{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       constants.OperatorGroup,
+			APIVersion: "operators.coreos.com/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    map[string]string{"app": constants.ObservabilityName},
```
jotak, line 25: here and everywhere else the `constants.ObservabilityName` is used, I think `app` isn't the best label to use, maybe "part-of" would be better? cf https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels except we remove the `app.kubernetes.io/` prefix for shorter labels

jotak: (or `managed-by`)

config/samples/flows_v1alpha1_flowcollector_dev.yaml
```diff
@@ -0,0 +1,84 @@
...
+  prometheus:
+    instanceSpec:
+      replicas: 1
+  consolePlugin:
+    image: "quay.io/netobserv/network-observability-console-plugin:main"
+    imagePullPolicy: Always
+    port: 9001
+  cno:
+    namespace: "openshift-network-operator"
+  op: loki-operator,strimzi-kafka-operator,grafana-operator
```
jotak, line 84: `op` isn't necessary anymore, right?
Also, `cno` was renamed `clusterNetworkOperator` (as a general rule, we should avoid cryptic names here, as this is something exposed to the end users)

config/samples/flows_v1alpha1_flowcollector_dev.yaml
```diff
@@ -0,0 +1,84 @@
...
+kind: FlowCollector
+metadata:
+  name: cluster
+spec:
+  namespace: "network-observability"
+  ipfix:
+    cacheActiveTimeout: 60s
+    cacheMaxFlows: 400
+    sampling: 100
+  goflowkube:
```
jotak, line 11: s/goflowkube/flowlogsPipeline/

config/samples/flows_v1alpha1_flowcollector_dev.yaml
```diff
@@ -0,0 +1,84 @@
...
+    kind: Deployment
+    replicas: 1
+    port: 2055
+    image: "quay.io/netobserv/goflow2-kube:main"
+    imagePullPolicy: Always
+    logLevel: trace
+    printOutput: true
+    healthPort: 8080
```
jotak, line 19:
```suggestion
  flowlogsPipeline:
    kind: Deployment
    replicas: 1
    port: 2055
    image: 'quay.io/netobserv/flowlogs-pipeline:main'
    imagePullPolicy: IfNotPresent
    logLevel: info
    printOutput: false
    enableKubeProbes: false
    healthPort: 8080
```

jotak:
quick question, in the initial description you mentioned it is necessary to create an aws S3 bucket prior to deploying flowcollector, does that mean that:
- we won't be able to run that in KinD / CRC?
- would it work on non-aws cloud providers?

config/samples/flows_v1alpha1_flowcollector_dev.yaml
jpinsonneau: yes sorry it seems I missed that file when I merged :angel:

jpinsonneau:
> quick question, in the initial description you mentioned it is necessary to create an aws S3 bucket prior to deploying flowcollector, does that mean that:
> 
> * we won't be able to run that in KinD / CRC?
> * would it work on non-aws cloud providers?

You can run this on Kind / CRC updating `loki.instanceSpec.storageClassName` to `standard`:
```yaml
      storageClassName: standard
```

It works for any cloud provider but needs an [aws bucket to store logs](https://github.com/ViaQ/loki-operator/blob/71ec533741819defdc23670cb990fbe2d2fd2024/internal/handlers/lokistack_create_or_update.go#L43-L52)
I guess we can implement [storage alternatives](https://grafana.com/docs/loki/latest/storage/#implementations---chunks) for this operator in the future without too much effort

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,290 @@
...
+			Kind:       "KafkaTopic",
+			APIVersion: "kafka.strimzi.io/v1beta2",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      spec.TopicName, // kafka topic name is set from metadata name
+			Namespace: ns,
+			Labels:    map[string]string{"part-of": constants.ObservabilityName},
+		},
+		Spec: &sv1b2.KafkaTopicSpec{
+			Config:     spec.TopicSpec.Config,
```
jotak, line 165: getting segfault here .. I'm using the provided flowcollector_dev yaml, but it doesn't have a `TopicSpec` configured. I guess it doesn't have to be a mandatory configuration, does it?

Here's my logs:

```
1.6468341854308603e+09	INFO	controller.flowcollector	subscription strimzi-kafka-operator already exists	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "component": "Operators"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14ddb40]

goroutine 454 [running]:
github.com/netobserv/network-observability-operator/controllers/operators.kafkaTopic(...)
	/opt/app-root/controllers/operators/operators_objects.go:165
github.com/netobserv/network-observability-operator/controllers/operators.(*Reconciler).getInstances(0xc000ae5b88, {0x18f711d, 0x16})
	/opt/app-root/controllers/operators/operators_reconciler.go:202 +0x1400
github.com/netobserv/network-observability-operator/controllers/operators.(*Reconciler).manageInstance(0xc000ae5b88, {0x1b015b8, 0xc001db0a20}, {0x18f711d, 0x16})
	/opt/app-root/controllers/operators/operators_reconciler.go:165 +0x5c

```

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,290 @@
...
+ *  - auth has been disabled
+ */
+func catalogSource(namespace string) *ov1alpha1.CatalogSource {
+	return &ov1alpha1.CatalogSource{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       constants.OperatorGroup,
+			APIVersion: "operators.coreos.com/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.ObservabilityName,
```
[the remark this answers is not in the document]
jpinsonneau: sure :+1:

controllers/operators/operators_objects.go
jpinsonneau: good catch :+1: nothing should be mandatory. I will add some logic there
It seems there is a similar case for grafana `DashboardSpec`

mariomac (approved):

jotak:
I'm back on testing this, sorry for the delay!
I'm seeing this error in logs:
```
1.647941642174263e+09	ERROR	controller.flowcollector	Failed to reconcile required operators	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "error": "no matches for kind \"LokiStack\" in version \"loki.openshift.io/v1beta1\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227
```

Maybe it's just the time that the custom loki catalog gets deployed ?

jotak:
Similar for grafana:
```
1.6479417328308504e+09	ERROR	controller.flowcollector	Reconciler error	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "error": "no matches for kind \"Grafana\" in version \"integreatly.org/v1alpha1\""}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227
```

jotak:
/hold
putting on hold, just to have a discussion with the rest of the team: is it ok or not to ship this feature that relies on a temporary / alpha, self-hosted version of the Loki operator? I don't think it's a blocker, but I'd prefer to talk with all prior

jotak:
Seems like Loki fails to deploy for me.
What I did is just build / deployed the operator, then applied the provided CR with `_dev` suffix, changing `storageClassName to standard` (or does `standard` won't work on AWS, and creating a S3 bucket is mandatory?).
Anyway, the error I get doesn't look related to that, here it is, in the loki operator logs:

```
{"_ts":"2022-03-22T09:34:41.38305699Z","_level":"0","_component":"loki-operator_controller-runtime_manager_controller_lokistack","_message":"Reconciler error","_error":{"msg":"Operation cannot be fulfilled on lokistacks.loki.openshift.io \"loki\": the object has been modified; please apply your changes to the latest version and try again"},"name":"loki","namespace":"network-observability"}
```
My LokiStack:

```
$ oc get lokistack -o yaml
apiVersion: v1
items:
- apiVersion: loki.openshift.io/v1beta1
  kind: LokiStack
  metadata:
    creationTimestamp: "2022-03-22T09:34:41Z"
    generation: 1
    labels:
      part-of: netobserv
    name: loki
    namespace: network-observability
    ownerReferences:
    - apiVersion: flows.netobserv.io/v1alpha1
      blockOwnerDeletion: true
      controller: true
      kind: FlowCollector
      name: cluster
      uid: 37950a15-ab0b-4407-8e5c-73819b0061bc
    resourceVersion: "39218"
    uid: 77b0b01a-2059-4b6d-876f-df1d4c3b5615
  spec:
    managementState: Managed
    replicationFactor: 1
    size: 1x.small
    storage:
      secret:
        name: test
    storageClassName: standard
  status:
    components: {}
    conditions:
    - lastTransitionTime: "2022-03-22T09:34:41Z"
      message: Missing object storage secret
      reason: MissingObjectStorageSecret
      status: "False"
      type: Degraded
    - lastTransitionTime: "2022-03-22T09:34:41Z"
      message: All components ready
      reason: ReadyComponents
      status: "True"
      type: Ready
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""

```

Maybe that's due to the early state of the loki operator?

jotak:
Also, it seems like the created operators are not removed after removing the CR, or editing it to disable them

api/v1alpha1/flowcollector_types.go
```diff
@@ -144,16 +156,21 @@ type FlowCollectorHPA struct {
 	Metrics []ascv2.MetricSpec `json:"metrics"`
 }
 
-// FlowCollectorLoki defines the desired state for FlowCollector's Loki client
+// FlowCollectorLoki defines the desired state for FlowCollector's Loki configuration
 type FlowCollectorLoki struct {
+	// InstanceSpec is the Spec section of LokiStack instance
```
jotak, line 161: Could you add something along these lines: "Experimental feature. If provided, the Loki operator will be automatically installed. At this moment, the Loki operator itself is experimental and hasn't been officially released by Grafana." ?
(or said differently, but you get the point...)

api/v1alpha1/flowcollector_types.go
```diff
@@ -38,9 +42,18 @@ type FlowCollectorSpec struct {
...
 	Loki FlowCollectorLoki `json:"loki,omitempty"`
 
+	// Kafka contains settings related to the kafka configuration
+	Kafka FlowCollectorKafka `json:"kafka,omitempty"`
+
+	// Grafana contains settings related to the grafana configuration
+	Grafana FlowCollectorGrafana `json:"grafana,omitempty"`
+
+	// Prometheus contains settings related to the prometheus configuration
+	Prometheus FlowCollectorPrometheus `json:"prometheus,omitempty"`
```
jotak, line 55: For all these three, we should also mention that providing the configs will trigger installing the operators, and that it's experimental.

jpinsonneau:
> What I did is just build / deployed the operator, then applied the provided CR with _dev suffix, changing storageClassName to standard (or does standard won't work on AWS, and creating a S3 bucket is mandatory?).

On AWS using `openshift-install` command you will not have standard `StorageClasses`
image

> Anyway, the error I get doesn't look related to that, here it is, in the loki operator logs:
> ...
> Maybe that's due to the early state of the loki operator?

> I changed the storageClassName back to gp2 but it didn't make it to the LokiStack, it's still showing standard

Yes ! Error status are not revelent at all and it just crashs if your storage / bucket is not correctly set. You then need to manually uninstall `Operator` or `LokiStack` since most of changes / checks are not managed on their side.

> Also, it seems like the created operators are not removed after removing the CR, or editing it to disable them

My initial plan was to use global operators. I changed this to our operator namespace:
https://github.com/netobserv/network-observability-operator/pull/59#issuecomment-1050074446
I can implement deletes now I guess but it will be easy to loose data by mistake

controllers/operators/operators_reconciler.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+// Reconcile the required installed operators
+func (r *Reconciler) Reconcile(ctx context.Context, target *flowsv1alpha1.FlowCollector) error {
+	operators := []string{}
+
+	// install operators if main instance is required
+	// dependent objects will be ignored if InstanceSpec is nil
+	if target.Spec.Loki.InstanceSpec != nil {
```
jotak, line 47: Same comment than here https://github.com/netobserv/network-observability-operator/pull/78#discussion_r861599886 :-)
We need an explicit flag, due to restrictions when the CR is created from a console form (not yaml). Checking nil pointers won't work (they'll never be nil).

By the way, I haven't checked in deep if the Kafka config is consistent with what @OlivierCazade is doing here: https://github.com/netobserv/network-observability-operator/pull/78 , do you know? Things like TopicName will be defined in just one place, right?

jpinsonneau: There is not much difference between the configs. I'll adapt after https://github.com/netobserv/network-observability-operator/pull/78 merge

controllers/operators/operators_objects.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    map[string]string{"part-of": constants.ObservabilityName},
+		},
+	}
+}
+
+func grafanaSubscription(name string, namespace string) *ov1alpha1.Subscription {
```
jotak, line 48: As discussed in slack, I think it's necessary for productization to make these subscriptions optional, and configurable, in order to adapt upstream vs downstream.
I think a good option would be to externalize these as Subscription yaml resources, that will be easily adaptable between upstream and downstream.

As a consequence, it must be considered in code that subscriptions can potentially be missing (e.g.: no community loki operator), which should not be considered an error.

PR #60, opened 2022-02-16
flowlogs-pipeline integration

jotak:
Replacing the container with flowlogs-pipeline:latest, and adapting the
ConfigMap. Also it needs get/list/watch nodes.

Lot of renaming.

Needs plugin PR: https://github.com/netobserv/network-observability-console-plugin/pull/78


jotak:
For the timestamp issue, I can see that `TimeFlowStart` / `End` are 0. I remember that there was an issue like that previously in goflow-kube, too. Maybe just bumping goflow2 would fix it (fl2m is one version behind goflow-kube on that)

eranra (approved):
Based on additional changes in https://github.com/jotak/network-observability-operator/pull/1/files

controllers/goflowkube/goflowkube_objects.go
```diff
@@ -24,25 +23,53 @@ const configPath = "/etc/goflow-kube"
...
 const (
-	healthServiceName       = "health"
-	healthTimeoutSeconds    = 5
-	livenessPeriodSeconds   = 10
-	startupFailureThreshold = 5
-	startupPeriodSeconds    = 10
+	healthServiceName = "health"
+
+/* 	healthTimeoutSeconds    = 5
+livenessPeriodSeconds   = 10
```
eranra, line 29: I added https://github.com/netobserv/flowlogs2metrics/issues/86 to close this gap ... once this is completed we will be able to return those checks.

controllers/goflowkube/goflowkube_objects.go
```diff
@@ -24,25 +23,53 @@ const configPath = "/etc/goflow-kube"
...
+	Encode    Encode      `json:"encode"`
+	Extract   Extract     `json:"extract"`
+	Transform []Transform `json:"transform"`
+	Write     Write       `json:"write"`
+}
+type Ingest struct {
+	Collector Collector `json:"collector"`
+	Type      string    `json:"type"`
+}
+type Collector struct {
```
eranra, line 55: @jotak this works ... It is possible to use the struct from here:  https://github.com/netobserv/flowlogs2metrics/blob/051a608b1368f25986bcfb1880b416e7d57d943a/pkg/api/ingest_collector.go#L20 directly instead of reimplementing the API struct. The value is that we can be sure that we are aligned to some version of flowlogs2metrics.

jotak: Yes, but until now we've tried to keep the operator loosely coupled with other models to avoid dependency complexity and keep vendors light. We can revisit this choice if we feel it's useful.

jotak:
It's ready for review, along with https://github.com/netobserv/network-observability-console-plugin/pull/78
I keep it on hold in order to do a release before merging, as discussed with @memodi 

/hold

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -159,61 +242,56 @@ func (b *builder) podTemplate(configDigest string) corev1.PodTemplateSpec {
...
+							Type:   "add_kubernetes",
+						},
+						{
+							Input:  "DstAddr",
+							Output: "DstK8S",
+							Type:   "add_kubernetes",
+						},
+					},
+				}}},
+			Write: Write{Type: "loki", Loki: Loki{Labels: constants.LokiIndexFields}},
```
jpinsonneau, line 281: Should we expose `decode` `transform` & `write` in `FlowCollector` config if you simply want to collect and send to something else without using the console UI ?

jpinsonneau (approved):

controllers/flowlogspipeline/flp_objects.go
jotak: Yes that could be interesting. For now I prefer to focus on FLP integration with as few changes as possible in the current API, but we can expect more config exposition soon. We need also to draw the line between something opinionated or not. FLP is flexible, but the operator could be more opinionated (which means easier to configure). Not saying it has to be.. it's a discussion to have

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -33,16 +33,55 @@ const (
...
+}
+
+type Pipeline struct {
+	Ingest    Ingest      `json:"ingest"`
+	Decode    Decode      `json:"decode"`
+	Encode    Encode      `json:"encode"`
+	Extract   Extract     `json:"extract"`
+	Transform []Transform `json:"transform"`
+	Write     Write       `json:"write"`
+}
```
mariomac, line 55: This structure with change after this PR: https://github.com/netobserv/flowlogs-pipeline/pull/83

Anyway, I think this structure will remain constant. Am I right? In that case, for simplification, we can just bundle the configuration as a YAML text/file and embed it into the FLP Pod descriptor.

mariomac (approved):

controllers/flowlogspipeline/flp_objects.go
jotak: It's not entirely constant, in particular the Loki part is exposed through CR

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -159,61 +242,56 @@ func (b *builder) podTemplate(configDigest string) corev1.PodTemplateSpec {
...
 // returns a configmap with a digest of its configuration contents, which will be used to
 // detect any configuration change
 func (b *builder) configMap() (*corev1.ConfigMap, string) {
 	configStr := `{}`
-	config := &ConfigMap{
-		Listen:      fmt.Sprintf("netflow://:%d", b.desired.Port),
-		Loki:        LokiConfigMap{},
-		PrintInput:  false,
-		PrintOutput: b.desired.PrintOutput,
+	config := &Config{
```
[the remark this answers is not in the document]
jotak: that fine I've figured out

PR #61, opened 2022-02-17
hack: development using "make run" (with loki)

eranra:
1. Updating the Makefile to allow deploy and undeploy of sample cr (there was some unneeded | etch command)
2. Adding Development documentation inside hack folder
3. Adding simple Loki deployment 
4. Adding simple Loki Grafana deployment script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,9 @@ DATE=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-all: build
+NAMESPACE ?= network-observability
+
+all: help
 
 ##@ General
 
@@ -149,7 +151,7 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 	$(KUSTOMIZE) build config/crd | kubectl apply -f -
 
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl delete -f -
+	$(KUSTOMIZE) build config/crd | kubectl --ignore-not-found=true delete -f - || true
 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
@@ -157,7 +159,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/default | kubectl delete -f -
+	$(KUSTOMIZE) build config/default | kubectl --ignore-not-found=true delete -f - || true
+
+run: manifests generate fmt lint ## Run a controller from your host.
+	go run ./main.go
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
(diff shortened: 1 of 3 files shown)


jotak:
Currently we have some hacks in https://github.com/netobserv/documents , especially for loki & grafana, so I don't think we want to have that in two different places, however maybe it makes more sense to have it in the operator repo indeed.

Beside that, we haven't introduced much shell scripts yet, maybe we can keep using the makefile to add loki/grafana deploying targets? We could copy here the zero-click loki/grafana templates (https://github.com/netobserv/documents/tree/main/examples/zero-click-loki) and simply execute `kubectl` on that from a `deploy-loki| target, so that we keep the deployment logic all via make targets?

eranra:
@jotak when I asked in slack and you and @stleerh provided two solutions I said to myself that we need something agreed and common for testing and experience with the operator. In flowlogs2metrics we have something that looks like https://github.com/netobserv/flowlogs2metrics/blob/main/Makefile#L128 ... it also deploys Grafana and Prometheus. Once we get the work from @OlivierCazade to deploy the dependent operator, maybe the script to deploy Loki will not be needed. 

As for the hacking doc ... it was very helpful to me when I did the end-to-end tests with the console and flowlogs2metrics deployed with the operator  ... especially https://github.com/netobserv/network-observability-operator/pull/61/files#diff-88be51eadd2c2ccdf0a6f5bb8255bbe4f37f2b887a6839f638bc45f6cfe18bb4R12-R26 are missing and should be added somewhere.


Please advise what you prefer to have here ... 
(1) Close this PR and just save things locally 
(2) Keep part of the PR for Hacking and remove other parts
(3) Keep all ... 

From my PoV ./... working with `make run` i.e.  https://github.com/netobserv/network-observability-operator/pull/61/files#diff-88be51eadd2c2ccdf0a6f5bb8255bbe4f37f2b887a6839f638bc45f6cfe18bb4R20 is very important and efficient and I really think that this might help others when they start to work with the oprerator.

jotak:
I like the approach you have in FL2M, it's similar to our "zero-click-loki" wired in the Makefile, I think we could have the same here. And same for grafana.

Also +1 for documenting the `make run` approach, but it could actually be in the main readme and not in a separate hack doc.

jpinsonneau:
> +1 for `make run`

Also consider this new option for quick deploy https://github.com/netobserv/network-observability-operator/pull/59
At the end, there is no need to manipulate anything else than the [FlowCollector yaml](https://github.com/netobserv/network-observability-operator/pull/59/files#diff-f31afdbb36d4b659e2e2e5c9be4a7e73df49c16dd6d808aae9e65f03c83156e5)

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+	go run ./main.go &
+	@echo "### Waiting for goflow-kube pod to be ready"
+	while : ; do kubectl get deployment goflow-kube && break; sleep 1; done
+	kubectl wait --timeout=120s --for=condition=ready pod -l app=goflow-kube
+	@echo "### Enable network-observability-plugin in OCP console"
+	oc patch console.operator.openshift.io cluster --type='json' -p '[{"op": "add", "path": "/spec/plugins", "value": ["network-observability-plugin"]}]'
+	@echo "### Starting log-flows export into goflow-kube pod"
+	GF_IP=`oc get svc goflow-kube -o jsonpath='{.spec.clusterIP}'` && \
+	echo "goflow-kube pod ID: $$GF_IP" && \
+	oc patch networks.operator.openshift.io cluster --type='json' -p "[{'op': 'add', 'path': '/spec', 'value': {'exportNetworkFlows': {'ipfix': { 'collectors': ['$$GF_IP:2055']}}}}]"
```
memodi, line 321: You don't need to patch network.operator any more since CNO now watches ovs-flows-config configmap.

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+.PHONY: undeploy-sample-cr
+undeploy-sample-cr:
+	sed -e 's~:main~:$(VERSION)~' ./config/samples/flows_v1alpha1_flowcollector.yaml | kubectl --ignore-not-found=true delete -f -  || true
+
+##@ Development
+
+.PHONY: deploy-loki
+deploy-loki: ## Deploy loki.
+	@echo "### Deploying loki"
+	kubectl create namespace $(NAMESPACE)  --dry-run=client -o yaml | kubectl apply -f -
```
stleerh, line 260: You can just do this.

```
kubectl create namespace $(NAMESPACE)
```

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+	kubectl config set-context --current --namespace=$(NAMESPACE)
+	curl -S -L https://raw.githubusercontent.com/netobserv/documents/main/examples/zero-click-loki/1-storage.yaml | kubectl create -f -  || true
+	curl -S -L https://raw.githubusercontent.com/netobserv/documents/main/examples/zero-click-loki/2-loki.yaml	 | kubectl create -f -  || true
```
stleerh, line 263: Not sure I want `|| true` here.

Makefile
eranra: @memodi if I understood correctly ... this will be enabled only on OCP 4.10, so for anyone that uses an older version (e.g. 4.9 == current), this will not work. We can sure remove this once 4.10 GA's

Make sense ??? the idea is to allow quick and easy development experience

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
 	sed -e 's~<IMG>~$(CATALOG_IMG)~' ./config/samples/catalog/catalog.yaml | kubectl apply -f -
 
 # Deploy the sample FlowCollector CR
-.PHONY: create-sample
-create-sample:
-	sed -e 's~:main~:v$(VERSION)~' ./config/samples/flows_v1alpha1_flowcollector.yaml | kubectl apply -f -
+.PHONY: deploy-sample-cr
+deploy-sample-cr:
+	sed -e 's~:main~:$(VERSION)~' ./config/samples/flows_v1alpha1_flowcollector.yaml | kubectl apply -f - || true
```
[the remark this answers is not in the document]
eranra: @stleerh with regards to sed ... the first expression is very simple to understand  just replace `:main` with the version (if someone changes it from main) ... we do not have any other `:main` s in the file and I think we will never have so let's agree to keep the very simple statement. BTW; it was there like that I did not change it


@stleerh with regards to the || true ---> yes, the thinking is that if you re-run one of the make tagets and something already exists and you try to recreate and/or you try to cleanup and the object is not there we do not want to fail the make file (this is not production code) we want to alert (== this is what we see, some log of the failure) and skip to the next line in the target. 

This makes life very easy to re-run things like `make deploy-sample-cr` if you do that several times ./... totally fine.

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+
+# Undeploy the sample FlowCollector CR
+.PHONY: undeploy-sample-cr
+undeploy-sample-cr:
+	sed -e 's~:main~:$(VERSION)~' ./config/samples/flows_v1alpha1_flowcollector.yaml | kubectl --ignore-not-found=true delete -f -  || true
```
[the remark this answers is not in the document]
eranra: fixed :-)

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+	kubectl port-forward --address 0.0.0.0 svc/grafana 3000:3000 2>&1 >/dev/null &
+	@echo -e "\ngrafana ui is available (user: admin password: admin) on http://localhost:3000\n"
+
+.PHONY: undeploy-grafana
+undeploy-grafana: ## Undeploy grafana.
+	kubectl config set-context --current --namespace=$(NAMESPACE)
+	kubectl delete --ignore-not-found=true route grafana
+	kubectl delete --ignore-not-found=true service grafana
+	kubectl delete --ignore-not-found=true deployment grafana
+	kubectl delete --ignore-not-found=true configMap grafana-datasources
```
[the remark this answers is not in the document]
eranra: fixed :-)

Makefile
```diff
@@ -241,6 +243,82 @@ catalog-deploy:
...
+	-PID=$$(pgrep --oldest --full "main.go"); pkill -P $$PID; pkill $$PID
+
+.PHONY: ocp-run
+ocp-run: ocp-cleanup ocp-deploy   ## OCP-deploy + run the operator locally
+	@echo "### Running operator locally (in background process)"
+	-PID=$$(pgrep --oldest --full "main.go"); pkill -P $$PID; pkill $$PID
+	go run ./main.go &
+	@echo "### Waiting for goflow-kube pod to be ready"
+	while : ; do kubectl get deployment goflow-kube && break; sleep 1; done
+	kubectl wait --timeout=120s --for=condition=ready pod -l app=goflow-kube
```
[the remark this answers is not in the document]
eranra: @stleerh 

This was an hour of frustration for me :-) 

314 is waiting for the **deployment** to appear .... and 315 is waiting for the **pod** to be ready using the kubectl  `wait` functionality  ... 
We need 314 because without that the deployment doesn't exist so we can't even ask for the pods' condition ... and we get an error immediately 


I LOVE K8S :-)

Makefile
eranra: @stleerh Yes ... but this will fail if the namespace exists... so this is like the recommended way to make sure that we create the namespace only if it doesn't exist. Another option is to add the regular || true ... but this is in "the eyes of the beholder"  on how to implement this  :-)

Makefile
eranra: @stleerh this is for development ... I do not want to fail if this is already deployed .... just move on ... in production ... yes you are totally correct but for dev ??? the logic is let's go on and do the best we can ... if something exists let's use it and do not fail ... same with double delete.

PR #65, opened 2022-03-03
add development_kind

eranra:
This PR adds an end-to-end kind deployment for the operator to allow quick and efficient development of the operator locally. It deployed Kind with OVN-Kubernetes and all relevant components


eranra:
@jotak the `kind.sh` is not working for me --- I am using windows WSL as my "linux machine" and I get::

```

eranra@ERANRAICHS-1MM6-wsl [3059] ~
$ kubectl logs ovs-node-5fvl4 -n ovn-kubernetes
================== ovnkube.sh --- version: 3 ================
 ==================== command: ovs-server
 =================== hostname: ovn-worker
 =================== daemonset version 3
 =================== Image built from ovn-kubernetes ref: refs/heads/master  commit: fc51865676fd515ebaaa7b43ebeeb139f30a4ce6
Starting ovsdb-server.
Configuring Open vSwitch system IDs.
Enabling remote OVSDB managers.
modprobe: FATAL: Module openvswitch not found in directory /lib/modules/5.10.60.1-microsoft-standard-WSL2
Inserting openvswitch module ... failed!
rmmod: ERROR: could not open '/sys/module/bridge/holders': No such file or directory
rmmod: ERROR: Module unloading is not supported
removing bridge module ... failed!
```

.mk/development.mk
```diff
@@ -0,0 +1,66 @@
+##@ Development
+
+.PHONY: create-kind-cluster
+create-kind-cluster: $(KIND) ## Create kind cluster with OVN-k8s
+	ln -sf $$GOPATH/bin/kind-v0.11.1 $$GOPATH/bin/kind; \
```
jotak, line 5: shouldn't it be `$(GOPATH)` rather than `$$GOPATH` ? It fails for me, I don't have `$GOPATH` defined on my laptop

jotak:
Sorry Eran, this PR has been pending for a long time...
My fear is that it could create more maintenance burden, to deal with all the various environment that the team is using , when starting something like kind.
I have some failures when I try to run it, I'd need to investigate what's going on

.mk/development.mk
```diff
@@ -0,0 +1,66 @@
...
+	sudo yum -y install python-pip; \
+	mkdir -p $$GOPATH/src/github.com/ovn-org; \
+	cd $$GOPATH/src/github.com/ovn-org; \
+	git clone git@github.com:ovn-org/ovn-kubernetes.git || true; \
+	cd ovn-kubernetes/contrib; \
+	./kind.sh; \
```
jotak, line 11: This fixes the issue I have:
`export KIND_IMAGE=docker.io/kindest/node && ./kind.sh`

jotak:
After fixing the issues, I'm still getting a timeout error during control plane init :/
When I used kind previously, I was passing `-wk 1` to kind.sh which used to work better with my small 16GB laptop, but now even that isn't enough. 
Hopefully I should soon receive a better laptop :)

.mk/development.mk
eranra: removed that area

eranra:
@jotak after fighting with OVN-k8s for half a day I moved to the old get logs from simulator on kind approach. This is not 100% testing the operator but it does provide value ... let me know what do you think on the updated PR

.mk/local.mk
```diff
@@ -0,0 +1,39 @@
...
+create-kind-cluster: $(KIND) ## Create kind cluster
+	-$(KIND) create cluster --config config/kubernetes/kind.config.yaml
+	kubectl cluster-info --context kind-kind
+
+.PHONY: delete-kind-cluster
+delete-kind-cluster: $(KIND) ## Delete kind cluster
+	$(KIND) delete cluster
+
+.PHONY: local-deploy
+local-deploy: create-kind-cluster deploy-all  ## Local deploy (kind, loki, grafana and example-cr)
```
jotak, line 13: I ran `make local-deploy` expecting it would deploy everything, especially as I see a `deploy-all` target here, but it doesn't seem to deploy (run) the operator, is it as intended?

I get that it's hard to name everything correctly as there are many different targets.

Maybe add some doc in the README for these new targets (not super-detailed, but in the main lines...)

.mk/local.mk
```diff
@@ -0,0 +1,39 @@
...
+	-PID=$$(pgrep --oldest --full "main.go"); pkill -P $$PID; pkill $$PID
+	-kubectl delete namespace network-observability
+
+.PHONY: local-redeploy
+local-redeploy: clean-leftovers undeploy-all deploy-all  ## Local re-deploy (loki, grafana and example-cr)
+
+.PHONY: local-undeploy
+local-undeploy: clean-leftovers undeploy-all delete-kind-cluster  ## Local cleanup
+
+local-run: create-kind-cluster local-redeploy ## run the operator locally
```
jotak, line 26: FYI when running `make local-run`, it successfully deploys the stack but there's some errors about missing namespace `openshift-network-operator` => this is something I've fixed in my PR https://github.com/netobserv/network-observability-operator/pull/97

jotak:
Not sure exactly why the CI fails, but I can reproduce locally, if I delete my `bin` directory then run `make build`, it fails due to missing `bin/controller-gen`. It works on the `main` branch.

I guess it is related to `.bingo` ? By the way, could you explain to the team a little bit about `bingo`, why using it and what it changes for us? (I don't know if others are familiar with it, personally I'm not)

eranra:
> Not sure exactly why the CI fails, but I can reproduce locally, if I delete my `bin` directory then run `make build`, it fails due to missing `bin/controller-gen`. It works on the `main` branch.
> 
> I guess it is related to `.bingo` ? By the way, could you explain to the team a little bit about `bingo`, why using it and what it changes for us? (I don't know if others are familiar with it, personally, I'm not)

Hi @jotak it was because the list of MakeFile targets changed ... it was not because of bingo but because of the additional `.mk` files under the `.mk` folder. I changed the `PROJECT_DIR` variable to use firstword and not lastword so now it uses the first Makefile which is the main Makefile and now things are working as expected :-)

jotak (approved):
/lgtm

thanks @eranra !

PR #69, opened 2022-03-08
move to use service named flowlogs-pipeline

eranra:


Makefile
```diff
@@ -324,9 +324,9 @@ ocp-run: ocp-cleanup ocp-deploy   ## OCP-deploy + run the operator locally
 	kubectl wait --timeout=120s --for=condition=ready pod -l app=goflow-kube
```
jotak, line 324: same here: 

```
@echo "### Waiting for flowlogs-pipeline pod to be ready"
while : ; do kubectl get deployment flowlogs-pipeline && break; sleep 1; done
	kubectl wait --timeout=120s --for=condition=ready pod -l app=flowlogs-pipeline
```

PR #71, opened 2022-03-11
NETOBSERV-126: Allow optional mapping for ports to service names

OlivierCazade:
Added ports name configuration in our CRD. This create a configmap that is then used by the console plugin.


OlivierCazade:
/hold
Until the PR is also ready on the console plugin side.

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -187,3 +217,47 @@ func buildServiceAccount(ns string) *corev1.ServiceAccount {
...
+		lokiWrite["batchSize"] = b.desiredLoki.BatchSize
+		lokiWrite["batchWait"] = b.desiredLoki.BatchWait.ToUnstructured()
+		lokiWrite["maxBackoff"] = b.desiredLoki.MaxBackoff.ToUnstructured()
+		lokiWrite["maxRetries"] = b.desiredLoki.MaxRetries
+		lokiWrite["minBackoff"] = b.desiredLoki.MinBackoff.ToUnstructured()
+		lokiWrite["staticLabels"] = b.desiredLoki.StaticLabels
+		lokiWrite["timeout"] = b.desiredLoki.Timeout.ToUnstructured()
+		lokiWrite["url"] = b.desiredLoki.URL
+		lokiWrite["timestampLabel"] = b.desiredLoki.TimestampLabel
+	}
```
jotak, line 238: we don't need this `lokiWrite`, right?

api/v1alpha1/flowcollector_types.go
```diff
@@ -227,6 +227,10 @@ type FlowCollectorConsolePlugin struct {
 	// HPA spec of an horizontal pod autoscaler to set up for the plugin Deployment.
 	// +optional
 	HPA *FlowCollectorHPA `json:"hpa,omitempty"`
+
+	// Additional port name to use in the console
+	// +optional
+	PortsName map[string]string `json:"portsname,omitempty"`
```
jotak, line 233: `PortsName` or `PortNames` ? I think the latter is more correct(?)
By the way, what's the way to disable port naming? Providing an empty object, as opposed to an absent field (which would rather show default names) ?

OlivierCazade: Yes, I will change this to PortNames.

My understanding of the goal of this story is to add the possibility to add custom name for ports, a custom name defined here would have priority over default IANA naming.

Do you think we should also provide the possibility to disable the port naming feature?

OlivierCazade: Ok, I missed that part. I will add an option.

api/v1alpha1/flowcollector_types.go
```diff
@@ -224,9 +224,29 @@ type FlowCollectorConsolePlugin struct {
...
+
+	//+kubebuilder:default:={enable:true}
+	// Configuration of the port to service name translation
+	PortNaming ConsolePluginPortConfig `json:"portnaming,omitempty"`
+}
+
+// Configuration of the port to service name translation feature of the console plugin
+type ConsolePluginPortConfig struct {
+	//+kubebuilder:default:=true
+	// Should this featur be enabled
```
jotak, line 244: typo "feature"

api/v1alpha1/flowcollector_types.go
```diff
@@ -224,9 +224,29 @@ type FlowCollectorConsolePlugin struct {
...
+	// LogLevel defines the log level for the console plugin
+	LogLevel string `json:"logLevel,omitempty"`
+
 	// HPA spec of an horizontal pod autoscaler to set up for the plugin Deployment.
 	// +optional
 	HPA *FlowCollectorHPA `json:"hpa,omitempty"`
+
+	//+kubebuilder:default:={enable:true}
+	// Configuration of the port to service name translation
+	PortNaming ConsolePluginPortConfig `json:"portnaming,omitempty"`
```
jotak, line 238: can you just rename json to `portNaming` for consistency? And same below: `portNames`

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -187,3 +219,32 @@ func buildServiceAccount(ns string) *corev1.ServiceAccount {
 		},
 	}
 }
+
+// returns a configmap with a digest of its configuration contents, which will be used to
+// detect any configuration change
+func (b *builder) configMap() (*corev1.ConfigMap, string) {
+	config := map[string]interface{}{
+		"PortNaming": b.desired.PortNaming,
```
jotak, line 227: here also, `portNaming` ?

jotak (commented):
A couple of small comments, but overall code LGTM
I'll test it now

api/v1alpha1/flowcollector_types.go
```diff
@@ -224,9 +224,29 @@ type FlowCollectorConsolePlugin struct {
...
+
+// Configuration of the port to service name translation feature of the console plugin
+type ConsolePluginPortConfig struct {
+	//+kubebuilder:default:=true
+	// Should this featur be enabled
+	Enable bool `json:"enable,omitempty"`
+
+	// Additional port name to use in the console
+	// +optional
+	PortNames map[string]string `json:"portnames,omitempty"`
```
jotak, line 249: In the description doc, could you add an example of what is expected? something like:

```
// E.g. portNames: { "9090": "open-metrics" }
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -224,9 +224,29 @@ type FlowCollectorConsolePlugin struct {
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"`
 
+	//+kubebuilder:validation:Enum=trace;debug;info;warn;error;fatal;panic
+	//+kubebuilder:default:=info
+	// LogLevel defines the log level for the console plugin
+	LogLevel string `json:"logLevel,omitempty"`
```
jpinsonneau, line 230: Should we mention that is for console plugin `backend` side ?

Would it be also interesting to also have logs levels on frontend in the future ?

controllers/flowcollector_controller_test.go
```diff
@@ -116,6 +116,12 @@ var _ = Describe("FlowCollector Controller", func() {
 								},
 							}},
 						},
+						PortNaming: flowsv1alpha1.ConsolePluginPortConfig{
+							Enable: true,
+							PortNames: map[string]string{
+								"3100": "loky",
```
jpinsonneau, line 122:
```suggestion
								"3100": "loki",
```

:smile:

api/v1alpha1/flowcollector_types.go
OlivierCazade: I think we should have only one option to enable backend and frontend debug mode, but I can mention that for now this only affect backend.

controllers/consoleplugin/consoleplugin_reconciler.go
```diff
@@ -172,15 +194,20 @@ func pluginNeedsUpdate(plg *osv1alpha1.ConsolePlugin, desired *pluginSpec, ns st
...
 }
 
-func deploymentNeedsUpdate(depl *appsv1.Deployment, desired *flowsv1alpha1.FlowCollectorSpec, ns string) bool {
+func deploymentNeedsUpdate(depl *appsv1.Deployment, desired *flowsv1alpha1.FlowCollectorSpec, ns string, cmDigest string) bool {
 	if depl.Namespace != ns {
 		return true
 	}
 	return containerNeedsUpdate(&depl.Spec.Template.Spec, &desired.ConsolePlugin) ||
 		hasLokiURLChanged(depl, &desired.Loki) ||
+		configChanged(&depl.Spec.Template, cmDigest) ||
```
jotak, line 203: looks like we missed checking here the command line arguments .. so, changing the logLevel (or Loki URL) should trigger updating the deployment, which doesn't happen today.
As I'm testing it locally, I can do the changes and send a PR on top of yours

PR #75, opened 2022-04-06
Make DaemonSet mode the default

jotak:
Also remove now unused PrintOutput (use debug or trace log level to see
the output on FLP)


config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -9,13 +9,12 @@ spec:
     cacheMaxFlows: 400
     sampling: 100
   flowlogsPipeline:
-    kind: Deployment
-    replicas: 1
+    kind: DaemonSet
+#    replicas: 1
```
mariomac, line 13: I'd remove this line to avoid confusion, or maybe just add another commented line to indicate that this belongs to the Deployment type as alternative:
```yaml
#   kind: Deployment
#   replicas: 1
```

jotak: done

mariomac (approved):

jpinsonneau (approved):

PR #78, opened 2022-04-07
NETOBSERV-255: FLP multiple deployments

OlivierCazade:
This PR rework the FLP reconciler to two different ones:
- a single reconciler in charge of managing a single FLP deployment
- a parent reconciler in charge of managing the global resources (clusterRole) and managing the single reconcilers


api/v1alpha1/flowcollector_types.go
```diff
@@ -68,6 +68,19 @@ type FlowCollectorIPFIX struct {
...
+// FlowCollectorKafka defines the desired Kafka config of FlowCollector
+type FlowCollectorKafka struct {
+	// Important: Run "make generate" to regenerate code after modifying this file
+
+	//+kubebuilder:default:=""
+	// Address of the kafka server
+	Address string `json:"address"`
+
+	//+kubebuilder:default:=""
+	// Address of the kafka topic to use
```
jpinsonneau, line 80:
```suggestion
	// Kafka topic to use
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -124,6 +137,10 @@ type FlowCollectorFLP struct {
 	//+kubebuilder:default:=true
 	// EnableKubeProbes is a flag to enable or disable Kubernetes liveness/readiness probes
 	EnableKubeProbes bool `json:"enableKubeProbes,omitempty"`
+
+	// Kafka configurations, if empty the operator will deploy a all-in-one FLP
+	// +optional
+	Kafka *FlowCollectorKafka `json:"kafka,omitempty"`
```
jpinsonneau, line 143: Should we put this as `FlowCollectorSpec` param as we did for Loki ?

So if other components will use it, it will stay generic (I'm thinking of the PoC operators and maybe health checks from console plugin)

OlivierCazade: Done, thanks!

api/v1alpha1/flowcollector_types.go
```diff
@@ -46,6 +46,10 @@ type FlowCollectorSpec struct {
 	// Loki contains settings related to the loki client
 	Loki FlowCollectorLoki `json:"loki,omitempty"`
 
+	// Kafka configurations, if empty the operator will deploy a all-in-one FLP
+	// +optional
+	Kafka *FlowCollectorKafka `json:"kafka,omitempty"`
```
eranra, line 51: Kafka here is ingress or egress ... we support both, and we need two configurations one when we ingress data from Kafka into FLP and one when we export from FLP data into kafka. 

It will be good to be more explicit so users will understand exactly what is going on

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -1353,6 +1353,22 @@ spec:
...
+                    description: Address of the kafka server
+                    type: string
+                  topic:
+                    default: ""
+                    description: Kafka topic to use
+                    type: string
+                required:
+                - address
+                - topic
+                type: object
```
eranra, line 1371: For the area of using FLP as ingress ... we will have multiple FLP instances consuming from the same topic. Need to make sure that we have some configuration on the distribution that Kafka is creating when sending data into those FLP's going forward this is going to be very important after we implement the stateful connection tracking capabilities. There might be multiple different configurations based on Load/ number of consumers etc ... those might need to be exported and configurable to users.

api/v1alpha1/flowcollector_types.go
eranra: I see that you use the term "single" for this simple case and I would recommend reversing the logic and using the term `distributed` for kafka and the distributed approach ... this is just my way of thinking on things, starting from simple == simple and moving into the more complicated distributed approach and not vice versa

eranra: I do not remember if we have TLS support (or any other security) for Kafka ... but we need that support both in the operator and in the FLP code.

docs/FlowCollector.md
```diff
@@ -118,6 +118,13 @@ FlowCollectorSpec defines the desired state of FlowCollector
           IPFIX contains the settings of an IPFIX-based flow reporter. This section should not be defined if the ebpf section is already defined<br/>
         </td>
         <td>false</td>
+      </tr><tr>
+        <td><b><a href="#flowcollectorspeckafka">kafka</a></b></td>
+        <td>object</td>
+        <td>
+          Kafka configurations, if empty the operator will deploy a all-in-one FLP<br/>
```
eranra, line 125: So ... default configuration is "without" Kafka ... when we will support the deployment of the Kafka operator we might want to simplify and interlock between things so ... if someone requests to deploy Kafka ... we can "default" to the end point that we deploy (because we know that the customer is using Kafka) 

@jpinsonneau this is for you to look on

jpinsonneau: Yes we have [the same approach](https://github.com/netobserv/network-observability-operator/pull/59/files#diff-4d2762740f523111148b4e0e1f03cd3d907d3de8ea2dfc9d40fc194efabadd70L191) with Loki in the PoC

```go
// force loki url to loki's distributor from operator if requested
if b.desiredLoki.InstanceSpec != nil {
	lokiWrite["url"] = "http://loki-distributor-http-loki:3100"
} else {
	lokiWrite["url"] = b.desiredLoki.URL
}
```

We will keep that approach for Kafka too

api/v1alpha1/flowcollector_types.go
jotak, line 51: We need to have an explicit flag to enable/disable kafka, rather than relying on nil pointer. This is because of how the CRD is handled in OLD/console form, it doesn't allow to remove a section. We discovered it the hard way for eBPF: https://issues.redhat.com/browse/NETOBSERV-319

OlivierCazade: @eranra about your different comments here:
- the kafka configuration block here is more global, this is the kafka configuration as for kafka in flow collection, this block will also concern ebpf agent once we will implement kafka support in the agent
- FLPReconciler and singleFLPReconciler do not oppose,  the model here is to have one parent reconciler which create three reconciler, one for each possible FLP configuration (all-in-one, ingestor and transformer) both reconciler types are used whatever the configuration is. I would prefer to keep the FLPReconciler name for the parent, since this is the one that is getting exposed, are you fine if I rename singleFLPReconciler to singleDeploymentReconciler for more clarity (we can remove the FLP part since this is an internal object and we are already in the FLP package)?
- I did not dig much into TLS yet, I can create a dedicated task about kafka security, would you be fine with that ?

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
OlivierCazade: Yes, the kafka configuration block is meant to change, but I would prefer to change it again once we have the connection tracking so we can test the different configurations and chose what we need to expose.
Would you be fine with that approach?

api/v1alpha1/flowcollector_types.go
eranra: @OlivierCazade with regards to the configuration ... I agree this fits, I think that I learned more while reading the code, so this is totally ok. 

singleDeploymentReconciler -- it is good --- agree this will help to make things more understood :-) 

TLS -- yes we might need that for production

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
eranra: Yes, it doesn't have to be in this PR, agree ... maybe just open a ticket somewhere to remember that we need to improve that later.

docs/FlowCollector.md
OlivierCazade: Yes, the plan is to have the same logic for all dependencies.

api/v1alpha1/flowcollector_types.go
OlivierCazade: Good to knows, thanks. I will add an enable flag.

OlivierCazade: Enable flag added

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -32,40 +32,54 @@ const (
 	startupPeriodSeconds    = 10
 )
 
+const (
+	ConfSingle           = "allInOne"
+	ConfKafkaIngestor    = "kafkaIngestor"
```
jotak, line 37: actually, don't we say Ingester ? :)

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -32,40 +32,54 @@ const (
...
+	ConfKafkaTransformer = "kafkaTransformer"
+)
+
+var FlpConfSuffix = map[string]string{
+	ConfSingle:           "",
+	ConfKafkaIngestor:    "-kingestor",
+	ConfKafkaTransformer: "-ktransform",
```
jotak, line 44: I would remove the "k"
"-ingester"
"-transformer" (missing -er)

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -400,17 +414,17 @@ func (b *builder) autoScaler() *ascv2.HorizontalPodAutoscaler {
...
 
 func buildClusterRole() *rbacv1.ClusterRole {
 	return &rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   constants.FLPName,
-			Labels: buildAppLabel(),
+			Labels: buildAppLabel(""),
 		},
 		Rules: []rbacv1.PolicyRule{{
 			APIGroups: []string{""},
```
jotak, line 430: On cluster roles, actually, they need to be split, they are not the same for ingester and for transformer.
Only the transformer needs to watch pods/services/nodes/replicasets
And I guess only the ingester needs to use hostnetwork (I'm actually not sure why we need that, @mariomac can you remind me?)

We can create two clusterrole objects (even in the case of a single FLP if it makes it simpler), and deal with cluster role attribution via clusterrolebindings

jotak: Also I don't know why we need horizontalpodautoscalers here: it's not FLP itself who create the HPA ?

controllers/ovs/flowsconfig_reconciler.go
```diff
@@ -40,7 +40,7 @@ func NewFlowsConfigController(client reconcilers.ClientHelper,
 // Reconcile reconciles the status of the ovs-flows-config configmap with
 // the target FlowCollector ipfix section map
 func (c *FlowsConfigController) Reconcile(
-	ctx context.Context, target *flowsv1alpha1.FlowCollector) error {
+	ctx context.Context, target *flowsv1alpha1.FlowCollector, flpServiceName string) error {
```
jotak, line 43: I think it's possible to avoid OVS reconciliation here, if we keep the same service regardless if it's flp or flp-ingester, it would avoid the OVS reconfig.
As you prefer, if you think it can make things simpler.

controllers/reconcilers/namespaced_objects_manager.go
```diff
@@ -75,7 +75,7 @@ func (m *NamespacedObjectManager) CleanupNamespace(ctx context.Context) {
 		ref.SetNamespace(namespace)
 		log.Info("Deleting old "+obj.kind, "Namespace", namespace, "Name", obj.name)
 		err := m.client.Delete(ctx, ref)
-		if err != nil {
+		if client.IgnoreNotFound(err) != nil {
```
jotak, line 78: nice!

jotak (commented):
tested and looks good, a couple of comments mostly on naming and splitting cluster roles

controllers/flowlogspipeline/flp_objects.go
mariomac: As far as I know, `hostNetwork: true` is only required by the eBPF agent, so as long as it should be associated with another cluster role, the hostnetwork label here shouldn't be necessary.

OlivierCazade: Doesn't we need host network for daemonset deployment? So we can bind node port which is used by ovs?

jotak: We can do a try removing it, see if it works... Mario you're the one who added it :) (cf #29 )

OlivierCazade: For the record here is the error message without hostNetwork permission:
Forbidden: not usable by user or serviceaccount, spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 2055: Host ports are not allowed to be used, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]

jotak: btw the way, I'm dealing with issues with `hostNetwork` when running with KIND instead of openshift, for this PR https://github.com/netobserv/network-observability-operator/pull/97
Obviously this `hostnetwork` config is specific to openshift and we need to do the equivalent with vanilla kube .. sounds a bit similar to what @mariomac is doing with the ebpf agent.

Without it, FLP never receives the ipfix flows. If I add `hostNetwork: true` on FLP pods, then they start receiving the flows ... but then they're unable to reach Loki, I guess because Loki is on the "cluster" network, not the host network 😭
Anyone got an idea?

OlivierCazade: In openshift without this permission, the pod is not allowed to be created.

Here you are saying that the pod is created, is able to bind the host port, but does not receive ovs flows without the permission? This is strange...

mariomac: Maybe the error is because you removed the permission to use hostnetwork and did not remove the "hostnetwork" specification from the FLP daemonset. Here is an example of FLP running as daemonset and not requiring hostnetwork (only the agent requires it): https://github.com/netobserv/netobserv-ebpf-agent/blob/main/deployments/flp-daemonset.yml

mariomac: @jotak maybe changing the DNS policy fixes the problem:
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/deployments/flp-daemonset.yml#L18-L20
However I still think FLP shouldn't need the `hostnetwork: true` tag. In the linked deployment, FLP works without requiring it, both in OpenShift and Kubernetes.

OlivierCazade: > Maybe the error is because you removed the permission to use hostnetwork and did not remove the "hostnetwork" specification from the FLP daemonset.

Yes but the daemonset needs to bind the node port, this is how ovs reach it in daemonset mode :
https://github.com/netobserv/network-observability-operator/blob/d0a45a19c287e681e063483d1991d050b8573896/controllers/ovs/flowsconfig_reconciler.go#L123

OlivierCazade: I am not saying that FLP as a DaemonSet would not work.

But if you deploy FLP as a DaemonSet using container port like in your example then you need a service to make this port reachable. And if you do so, then you can not guarantee anymore that ovs flows will be send to the corresponding FLP container on the same node hence avoiding network communications.

To me, the point of using DaemonSet and host network instead of Deployment and service was to grantee that the communication between ovs and flp would be kept on the same node.

mariomac: Yeah, but please observe that in the example configuration, the eBPF agent is still sending to the FLP in its own host, that bounds the port to a host port:
https://github.com/netobserv/netobserv-ebpf-agent/pull/27/files#diff-d67b1319dcea073fb864a6f7d9e93ac5d56d34d531779dcf092e39364c72318dR31

OlivierCazade: I am missing something in your example then:
- ebpf-agent is sending to status.hostIP which correspond to the node IP and to the port 9999 
- flp is binding the container port 9999

I am missing the part in your example that bind the host port 9999 to the container port 9999.

My understanding is that the flp daemonset would also require hostNetwork to true so the container port getting mapped is the host port.

OlivierCazade: Yes, I know how hostNetwork feature works.

> it will be automatically bound to the host port 9999 unless you explicitly set another hostPort property.

I am very surprise about this.

Do you have documentation about it? What kind of cluster are you using for this test? The [k8s official documentation](https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#exposing-pods-to-the-cluster) does not mention surch automatic bind and at the oposite state the following:
> Note that the containers are not using port 80 on the node, nor are there any special NAT rules to route traffic to the pod.

mariomac: For more safety, I'd specify `hostPort` to the same value as `containerPort` but it still should work without setting `hostNetwork: true`

OlivierCazade:
I pushed a few new commits:
- ingestor to ingester name correction
- deployment suffix changes from -kingester to -ingestor and from -ktransform to -transformer (for the record, -ktransformer was hitting some length limit this why I move to -ktransform but -transformer also fit)
- refactoring cluster role to split into two different roles, please note that both roles have for now the same permissions, this is because for now, both ingester and transformer do the same thing, some permissions will be removed once different kafka configurations will be done
- flp-ingester and the single flp configuration now use the same service to avoid ovs unnecessary reconciliation

controllers/ovs/flowsconfig_reconciler.go
OlivierCazade: Both deployment now use the same service, but now that the refactoring was done, I kept using an argument in the ovs reconciliation instead of relaying on a constant.

Is it fine for you?

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -358,24 +372,27 @@ func (b *builder) service(old *corev1.Service) *corev1.Service {
 	}
 	// In case we're updating an existing service, we need to build from the old one to keep immutable fields such as clusterIP
 	newService := old.DeepCopy()
+	newService.Spec.Selector = b.selector
+	newService.Spec.SessionAffinity = corev1.ServiceAffinityClientIP
```
jotak, line 376: oh actually I guess this session affinity should be needed even without your work on kafka right?

controllers/flowlogspipeline/flp_objects.go
jotak: JIRA created, thanks for clearing that up! https://issues.redhat.com/browse/NETOBSERV-362

jotak:
While testing I noticed the transformer pod was stuck in a `pending` state, and it turns out it's because it is trying to listen on node port `2055`, which obviously it shouldn't (only the ingester should listen there).
But maybe that's something you plan to change in NETOBSERV-256 ?

jotak:
Also I'm seeing a couple of errors in the logs, which aren't too concerning (on first run, it's trying to delete stuff that doesn't exist), but we should try to fix, maybe via another PR / JIRA (I'm not sure if there's an easy fix or if it needs some refactoring)

jotak:
=> https://issues.redhat.com/browse/NETOBSERV-365

PR #81, opened 2022-04-19
NETOBSERV-201: Allow configuring Netobserv agent

mariomac:
You might observe that some default values differ from the IPFIX section:

* Sampling: disabled by default.
* Cache Active Timeout: decreased from 10s to 5s for a more "realtime" feeling.
* Cache Max Flows: increased from 100 to 1000 because the lack of sampling increased the rate of which samples are submitted (less agregation, more processing/traffic).

Related PR: https://github.com/netobserv/netobserv-ebpf-agent/pull/15


jpinsonneau (approved):

config/samples/flows_v1alpha1_flowcollector_ebpf.yaml
```diff
@@ -8,6 +8,13 @@ spec:
   ebpf:
     image: 'quay.io/netobserv/netobserv-ebpf-agent:main'
     imagePullPolicy: IfNotPresent
+    sampling: 0
+    cacheActiveTimeout: 5s
+    cacheMaxFlows: 1000
+    interfaces: []
+    excludeInterfaces: ["lo"]
+    buffersLength: 50
+    verbose: true
```
jotak, line 17: not necessarily for this PR (as it would also impact the agent), but can we change that to `logLevel` like we're doing for the other components, or is there a reason why it needs to be a bool ?

jotak (approved):
lgtm (not tested)

config/samples/flows_v1alpha1_flowcollector_ebpf.yaml
mariomac: It's fine. I'll address the observation into this PR. Thanks!

mariomac: Do'h! I approved before submitting the previous PR. The verbose-to-loglevel change has been added here: https://github.com/netobserv/network-observability-operator/pull/82

PR #83, opened 2022-04-20
NETOBSERV-221 Patch console CR from operator

jotak:


api/v1alpha1/flowcollector_types.go
```diff
@@ -324,6 +330,9 @@ type FlowCollectorStatus struct {
 
 	// Namespace where console plugin and flowlogs-pipeline have been deployed.
 	Namespace string `json:"namespace,omitempty"`
+
+	// Whether or not the plugin is already registered with the console operator
+	ConsolePluginRegistered bool `json:"consolePluginRegistered,omitempty"`
```
mariomac, line 335: I'd not expose this as a property that is visible to the customer. I'd maybe remove this property and dynamically verify the "consolepluginregistered" value on each reconciliation step: read the `cluster` Console Operator and verify that the `network-observability-plugin` value is contained in the `.spec.plugins` list.

jotak: yeah, you're right

mariomac:
I have been testing it. There are two issues I'd like to discuss:
* If I do an update of the Flow collector switching off the `register` property, the UI is not undeployed. I'd not fix this into this PR as this is a too-usual case in this operator and I created an specific issue for this: https://issues.redhat.com/browse/NETOBSERV-304
* If I undeploy the Flow Collector, the UI Plugin is also undeployed. Do we really want this behavior? I think the current behavior is consistent, but on the other side I'm thinking in a use case in which a customer collects all the flows during some time, then undeploys the flow collector to save resources, but still want to inspect the past flows. On the other side it's true that many other users would expect the plugin not being present anymore if they undeploy/disable flow collection.

jotak:
Pushed new commits to update doc and add some tests
I've split the big controller test to extract the console plugin apart.

jotak:
> 
>     * If I do an update of the Flow collector switching off the `register` property, the UI is not undeployed. I'd not fix this into this PR as this is a too-usual case in this operator and I created an specific issue for this: https://issues.redhat.com/browse/NETOBSERV-304
> 

But this is as intended: this field does not control whether or not the plugin is deployed, but just if it is registered against the console operator (just to avoid the extra-step of patching the console resource). I named it "register" rather than "enable" in the hope that it clarifies this point, but maybe it can be improved?

jotak:
> If I undeploy the Flow Collector, the UI Plugin is also undeployed. Do we really want this behavior? I think the current behavior is consistent, but on the other side I'm thinking in a use case in which a customer collects all the flows during some time, then undeploys the flow collector to save resources, but still want to inspect the past flows. On the other side it's true that many other users would expect the plugin not being present anymore if they undeploy/disable flow collection.

@mariomac for this second point, I don't think the user should remove the whole FlowCollector resource in that situation, because it's how operators work: the CR controls what is deployed; remove the CR and all managed resources are removed as well. So we would have to find an alternative, maybe something simple like a bool flag to switch on/off flow collection ?

mariomac:
@jotak good point. Maybe we should add a task to implement this "disablement" of flow collection in case you want to keep the rest of the components. At the moment I think this PR can be merged

mariomac (approved):

jotak:
Thanks @mariomac 
I've created this to track your remark: https://issues.redhat.com/browse/NETOBSERV-310 (with some open points inside, don't hesitate to chime in)
/approve

PR #84, opened 2022-04-21
Adding support for FLP operational metrics

eranra:
Use on each of the FLP instances: http://localhost:90990/metrics
For example::
```
1000650000@flowlogs-pipeline-hpjn4:/$ curl localhost:9090/metrics | grep "ingest_collector_flow_logs_processed"
# HELP ingest_collector_flow_logs_processed Number of log lines (flow logs) processed
# TYPE ingest_collector_flow_logs_processed counter
ingest_collector_flow_logs_processed 2899
```

We see that we processed 2899 from the point in time we started the pod

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -166,6 +166,12 @@ type FlowCollectorFLP struct {
 	// HealthPort is a collector HTTP port in the Pod that exposes the health check API
 	HealthPort int32 `json:"healthPort,omitempty"`
 
+	//+kubebuilder:validation:Minimum=1
+	//+kubebuilder:validation:Maximum=65535
+	//+kubebuilder:default:=9090
+	// PrometheusPort is the prometheus HTTP port: this port exposes prometheus metrics
+	PrometheusPort int32 `json:"prometheusPort,omitempty"`
+
 	//+kubebuilder:default:="quay.io/netobserv/flowlogs-pipeline:main"
 	// Image is the collector image (including domain and tag)
 	Image string `json:"image,omitempty"`
--- a/config/samples/flows_v1alpha1_flowcollector.yaml
+++ b/config/samples/flows_v1alpha1_flowcollector.yaml
@@ -18,6 +18,7 @@ spec:
     logLevel: info
     enableKubeProbes: true
     healthPort: 8080
+    prometheusPort: 9090
   loki:
     url: 'http://loki:3100/'
     batchWait: 1s
--- a/controllers/flowlogspipeline/flp_objects.go
+++ b/controllers/flowlogspipeline/flp_objects.go
@@ -25,6 +25,7 @@ const configFile = "config.yaml"
 
 const (
 	healthServiceName       = "health"
+	prometheusServiceName   = "prometheus"
 	healthTimeoutSeconds    = 5
 	livenessPeriodSeconds   = 10
 	startupFailureThreshold = 5
@@ -114,6 +115,11 @@ func (b *builder) podTemplate(configDigest string) corev1.PodTemplateSpec {
 		ContainerPort: b.desired.HealthPort,
 	})
 
+	ports = append(ports, corev1.ContainerPort{
+		Name:          prometheusServiceName,
+		ContainerPort: b.desired.PrometheusPort,
+	})
+
 	container := corev1.Container{
 		Name:            constants.FLPName,
 		Image:           b.desired.Image,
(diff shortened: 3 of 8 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -164,6 +164,12 @@ type FlowCollectorFLP struct {
 	// HealthPort is a collector HTTP port in the Pod that exposes the health check API
 	HealthPort int32 `json:"healthPort,omitempty"`
 
+	//+kubebuilder:validation:Minimum=1
+	//+kubebuilder:validation:Maximum=65535
+	//+kubebuilder:default:=9102
```
jotak, line 169: Default for prometheus is generally 9090, or if already used 9091 and so on , any special reason to use 9102 here?

api/v1alpha1/flowcollector_types.go
```diff
@@ -164,6 +164,12 @@ type FlowCollectorFLP struct {
...
+	// PromPort is the prometheus HTTP port: this port exposes prometheus metrics
+	PromPort int32 `json:"promPort,omitempty"`
```
jotak, line 171: Also could we rename to `PrometheusPort` ? As this is the end-user exposed config, we tend to make it more expressive

jotak: Side question: it's the same port used for pipeline metrics and internal metrics, right? We don't deploy 2 prometheus instances?

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -245,46 +308,22 @@ func (b *builder) configMap() (*corev1.ConfigMap, string) {
 			{"name": "enrich",
 				"follows": "decode",
 			},
-			{"name": "encode",
-				"follows": "enrich",
-			},
 			{"name": "loki",
-				"follows": "encode",
+				"follows": "enrich",
```
jotak, line 312: we don't need to have `encode` between `enrich` and `loki` ?

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -245,46 +308,22 @@ func (b *builder) configMap() (*corev1.ConfigMap, string) {
...
-				},
+			{"name": "prometheus",
+				"follows": "aggregate",
 			},
-			{"name": "loki",
-				"write": map[string]interface{}{
-					"type": "loki",
-					"loki": lokiWrite,
-				},
+			{"name": "write_none",
```
jotak, line 320: what does "write_none" mean? Is it not going to expose the metrics to prom endpoint? or is that performed  via `prometheus` stage?

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -18,6 +18,7 @@ spec:
     logLevel: info
     enableKubeProbes: true
     healthPort: 8080
+    promPort: 9201
```
jotak, line 21: it's 9201 here vs 9102 by default .. on purpose, or a mistake?

jotak (commented):
I wrote a few remarks related to config, but tested and working fine!

api/v1alpha1/flowcollector_types.go
eranra: Fixed, changed to 9090

(not sure why 9102 was used by @KalmanMeth in FLP ... @KalmanMeth any info to add here ???)

api/v1alpha1/flowcollector_types.go
eranra: fixed
Yes, one Prometheus end-point for all metrics

controllers/flowlogspipeline/flp_objects.go
eranra: It is empty, and not in use at all right now (we do not have any network metrics, so no enrichment or encoding) but I wanted to add those empty steps for completeness of the pipeline. This will change over time once we start to add metrics to the pipeline.

config/samples/flows_v1alpha1_flowcollector.yaml
eranra: Fixed, a mistake ... I changed all of them to 9090

controllers/flowlogspipeline/flp_objects.go
eranra: There is an anomaly between the code that @mariomac wrote and the code that @KalmanMeth wrote that does not allow Prometheus to be finalizing stage. So this is empty "don't nothing" termination stage in the pipeline. I talked with @KalmanMeth on that and once they fix that we will be able to remove that empty stage.

eranra: I added an issue in FLP repo

api/v1alpha1/flowcollector_types.go
KalmanMeth: @eranra @jotak Probably I ported this from code that already used port 9090 for the regular prometheus resource monitor, so another port was used for network metrics prometheus connector. Nothing sacred about the port number.

PR #85, opened 2022-04-21
NETOBSERV-304: fix update reconciliation of eBPF agent

mariomac:
replaces manual comparison of fields by `equality.Semantic.DeepDerivative`, which skips fields of the "actual" state that are not in the "desired" state (e.g. these fields that are added by Kubernetes as metadata)


jotak (approved):
lgtm (not tested)

So is that `DeepDerivative` the silver bullet that we should use everywhere to check for changes?

PR #86, opened 2022-04-22
Allow passing custom env vars to eBPF agent

mariomac:
It also removes the `buffersLength` option from the descriptor, as this is a very low-level internal parameter that shouldn't be exposed to the common users.

Any low-level internal detail can be still configured by means of the `env` argument.


jotak (approved):
lgtm

PR #91, opened 2022-04-28
Fix grafana dashboard

jotak:


memodi (commented):
/lgtm 
tested works with latest Grafana release, and should fix: https://issues.redhat.com/browse/NETOBSERV-261 

@jotak - just curious, how did you came up with uid: P982945308D3682D1

jotak:
@memodi I didn't do anything special, that generated by Grafana when I'm saving / exporting the dashboard. I've no idea if it has any impact, but since it works for you & others, I guess that's ok :-)

PR #93, opened 2022-05-02
FLP: use prometheus as terminal stage

mariomac:
This PR in FLP sets prometheus as terminal stage: https://github.com/netobserv/flowlogs-pipeline/pull/194

This means that we need to update our definition in order to avoid breaking the installation when we push the master image of FLP.


mariomac:
Unblocking it after internally discussing and make sure that this won't break the next NOO release.

jpinsonneau (approved):

PR #99, opened 2022-05-10
NETOBSERV-307: eBPF agent running with capabilities instead of full privileges

mariomac:
It also allows configuring the agent to run with full capabilities as a fallback option for some Kubernetes distributions that do not recognize the `BPF` and `PERFMON` capabilities (documented here https://github.com/netobserv/netobserv-ebpf-agent/pull/23)


jpinsonneau (approved):
LGTM

PR #100, opened 2022-05-11
Console plugin annotation

jpinsonneau:
Missing `console.openshift.io/plugins` annotation

> An operator can also give console a hint that it has a plugin available through the console.openshift.io/plugins annotations. The annotation tells the console to show a checkbox to enable plugins during operator install. Only users who can edit the console operator config will see the checkbox. The annotation value is a serialized JSON array of strings, where each item is the name of a ConsolePlugin resource the operator will create.
https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

### DIFF
--- a/config/manifests/bases/netobserv-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/netobserv-operator.clusterserviceversion.yaml
@@ -9,6 +9,7 @@ metadata:
     createdAt: ':created-at:'
     description: A network observability operator based on netflows (IPFIX) for OVN
     repository: https://github.com/netobserv/network-observability-operator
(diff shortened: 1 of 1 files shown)


jotak:
I think it's only relevant in the case our operator is installed before the console operator, which shouldn't happen too often, but why not :)

/lgtm

jpinsonneau:
> I think it's only relevant in the case our operator is installed before the console operator, which shouldn't happen too often, but why not :)

That's not my understanding of `The annotation tells the console to show a checkbox to enable plugins during operator install`

For me it should add an option to enable the plugin here:
image

Check:
https://github.com/openshift/console/pull/8175

PR #101, opened 2022-05-11
NETOBSERV-265 and NETOBSERV-348 timestamp indexing

jotak:
As a result to using millisecond-based time fields (TimeFlowEndMs,
TimeFlowStartMs), we also need to change the time indexed field to keep
consistent queries (TimeFlowEnd => TimeFlowEndMs)

Take that opportunity to stop exposing this as a config, as it could
create troubles


jotak:
/hold
requires https://github.com/netobserv/flowlogs-pipeline/pull/202

mariomac (approved):

PR #103, opened 2022-05-17
Removed goflow references

OlivierCazade:
I noticed this while working on kafka but did not want to overload the kafka PR.

We still have naming reference to goflow instead of flp, this PR change them.


jpinsonneau (approved):
:+1: looks good !

PR #105, opened 2022-05-18
NETOBSERV-365 Fix errors in operator logs

jotak:
The errors we see in the operator logs are due to trying to clean items from the previous namespace (in the case of a changing namespace), despite the context here is not a changing namespace.
So I just duplicated the cleanup function to allow cleaning the current namespace.


OlivierCazade (approved):
/lgtm thanks!

mariomac (approved):

PR #109, opened 2022-05-31
Remove netflow v5 load generator

jotak:
It was not officially supported by the operator, and now has been
removed from its config to avoid keeping unnecessary opened port


jotak:
Hey @mariomac @eranra ,
At the moment I've just removed the load generator (as it won't work anymore), but we can reintroduce something. I was thinking about creating a tool similar to `networkstatic/nflow-generator` but for ipfix, using [vmware's ipfix lib](https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/test/ipfix.go). That could also be useful for e2e tests if we struggle having real OVS flows.

mariomac (approved):

PR #111, opened 2022-06-01
add example workload (with syn attack)

eranra:
The workload is based on GoogleCloudPlatform/microservices-demo
Syn attack is based on docker and an example from https://github.com/bilalcaliskan/syn-flood

### DIFF
--- a/.mk/ocp.mk
+++ b/.mk/ocp.mk
@@ -1,19 +1,25 @@
 ##@ OCP
 
-.PHONY: ocp-expose
-ocp-expose:
+.PHONY: ocp-expose-infra
+ocp-expose-infra:
 	oc expose service grafana || true
 	@grafana_url=$$(oc get route grafana -o jsonpath='{.spec.host}'); \
 	echo -e "\nAccess grafana on OCP using: http://"$$grafana_url"\n"
 	oc expose service loki || true
 	@loki_url=$$(oc get route loki -o jsonpath='{.spec.host}'); \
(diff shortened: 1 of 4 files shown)


eranra:
> is that something where you can leverage netobserv to spot the attack? how would you demo that? (I'm interested if I can reuse that use-case in demo/talk :) )

Maybe this syn attack traffic is not "seen" by OVS/N ... I need to check more

eranra:
@jpinsonneau I see strange amount of packets (on the edges) when I change the view to show the number of packets::

image

I was expecting a very small amount. And BTW; I see the same in the table ... so something is not making sense here: 

image

jotak:
maybe with sampling = 1?

.mk/ocp.mk
```diff
@@ -8,6 +8,9 @@ ocp-expose:
 	oc expose service loki || true
 	@loki_url=$$(oc get route loki -o jsonpath='{.spec.host}'); \
 	echo -e "\nAccess loki on OCP using: http://"$$loki_url"\n"
+	oc expose -n sample-workload service frontend-external || true
```
jotak, line 11: I tend to use `make ocp-expose` just to create a grafana route. Can you split in two different targets then? (maybe `ocp-expose` and `ocp-expose-all`, similarly to the `deploy-all` that deploys more than just the infra ?

eranra: Fixed (@jotak  please review again that this fits your needs )

jotak (approved):
lgtm!

PR #113, opened 2022-06-02
Move FLP log level in the example CR to error

eranra:
No reason for the example CR's to be in info level for FLP. If needed this can be changed dynamically. To be able to use NOO similar to customers best is to use error level.

### DIFF
--- a/config/samples/flows_v1alpha1_flowcollector.yaml
+++ b/config/samples/flows_v1alpha1_flowcollector.yaml
@@ -26,7 +26,7 @@ spec:
     port: 2055
     image: 'quay.io/netobserv/flowlogs-pipeline:main'
     imagePullPolicy: IfNotPresent
-    logLevel: info
+    logLevel: error
     enableKubeProbes: true
     healthPort: 8080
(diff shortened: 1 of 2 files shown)


jotak:
Is it to avoid the messages like `Can't find kubernetes info for IP  ...` ?
I agree this is annoying ... but on the other hand I find the starting logs quite useful in general and good to have by default. So, I don't know what to think ...

eranra:
@jotak yes ... this message is filling up the log .... so moving this specific message to debug is also a totally reasonable solution.  BUT we will need a new release for this to effect, right?

eranra:
@jotak I need your view on what to do with this PR? we merged the FLP change  https://github.com/netobserv/flowlogs-pipeline/commit/8d711a2d4f139b1ae89f1c6589dd1789c6ce5d5a so need to decide if we drop or merge this one?

jotak:
I'd say now the FLP change is merged we can keep the Info logging by default, no?

eranra:
OK ... so closing this PR without merging. I hope this will not cause any more troubles ....

PR #114, opened 2022-06-03
Remove decode stages, rename some stages

jotak:
Requires https://github.com/netobserv/flowlogs-pipeline/pull/225

Decode stages should not be set after ingesting ipfix or grpc


jotak:
btw @memodi I'd be curious to know if there's a noticeable perf improvement with this PR (potentially on flow processing rate or FLP CPU usage)

memodi:
@jotak ack. We do have some numbers now with 0.1.2 release which will be good for comparison purposes for upcoming release with this change.

PR #115, opened 2022-06-07
Document FLP DaemonSet / Deployment kinds subtelties

jotak:


api/v1alpha1/flowcollector_types.go
```diff
@@ -192,7 +192,10 @@ type FlowCollectorFLP struct {
 
 	//+kubebuilder:validation:Enum=DaemonSet;Deployment
 	//+kubebuilder:default:=DaemonSet
-	// Kind is the workload kind, either DaemonSet or Deployment
+	// Kind is the workload kind, either DaemonSet or Deployment. When DaemonSet is used, each pod will receive
+	// flows from the node it is running on. When Deployment is used, the flows traffic received from nodes will
+	// be load-balanced. Note that in such a case, the number of replicas should be less or equal to the number of
+	// nodes, as extra-pods would be unused due to session affinity with the node IP.
```
OlivierCazade, line 198: Nice I did not think of that limitation.

OlivierCazade (approved):
/lgtm

jotak:
@nalhadef let me know if you would reformulate :)
The idea I want to express here, is that when the user configures flowlogs-pipeline as a Deployment, there's a limitation / known issue, in that it is worthless to have more replicas than the number of nodes, because those replicas would never be used

memodi (approved):
thanks! didn't know this.

PR #116, opened 2022-06-07
Kafka documentation improvement

OlivierCazade:


api/v1alpha1/flowcollector_types.go
```diff
@@ -174,7 +174,11 @@ type FlowCollectorKafka struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
 	//+kubebuilder:default:=false
-	// Should this feature be enabled
+	// Should this feature be enabled.
+	// If kafka feature is enabled flow collection will be done in two steps, an ingestion step and a transformation step.
+	// The first step is either done by a first FLP deployment or by the ebpf agent.
```
jotak, line 179: Since it is the user-facing documentation, we should avoid using non obvious acronyms like FLP. Can you change to "flowlogs-pipeline" ? (here and below)

OlivierCazade: Fixed, thanks!

PR #117, opened 2022-06-08
Perf dash

memodi:
Adding perf metrics dashboard

### DIFF
--- a/config/samples/dashboards/NetObserv_Perf_Metrics.json
+++ b/config/samples/dashboards/NetObserv_Perf_Metrics.json
@@ -1,0 +1,865 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": null,
+  "iteration": 1654711342423,
+  "links": [],
+  "panels": [
+    {
+      "aliasColors": {},
+      "bars": false,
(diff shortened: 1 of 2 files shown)


jotak:
How do you use it? Do you use the cluster grafana (but I think it is read-only and we cannot import dashboard, no?) Or do you use our script-deployed grafana (with "make deploy-grafana") ? In that case we need to configure it with TLS to fetch cluster prometheus, no? Or do you proceed differently?

memodi:
> How do you use it? Do you use the cluster grafana (but I think it is read-only and we cannot import dashboard, no?) Or do you use our script-deployed grafana (with "make deploy-grafana") ? In that case we need to configure it with TLS to fetch cluster prometheus, no? Or do you proceed differently?

@jotak - for this you need to have:
1. userWorkload to be enabled,  
2. service created for FLP
3. serviceMonitor to be created for FLP metrics. 
Manifests for those I have attached here: https://issues.redhat.com/browse/NETOBSERV-332 

After that an additional datasource is needed to be configured in grafana with Thanos Querier (because user workload metrics can't be queried to cluster prometheus) . Right now I have above steps scripted and use it with [dittybopper](https://github.com/cloud-bulldozer/performance-dashboards/blob/master/dittybopper/deploy.sh) script with custom template with additional datasource configured.

if you prefer I can add those manifests for above steps here and update deploy-grafana.sh script to set up with additional datasource, however I am not sure if that would be very OCP specific.

jotak:
hey @memodi 
I haven't tested it, but if this is helpful to you, don't hesitate to merge.
/lgtm

memodi:
Hi @jotak I think I raised this PR just so others can use this dashboard as well, since then I've updated our dashboards, I'll close this one for now and will continue to manage the dashboard in [QE Repo](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/netobserv-perf-tests/scripts/netobserv_dittybopper_ebpf.json)

PR #120, opened 2022-06-14
NETOBSERV-309 skip TLS checks & add TenantID

jpinsonneau:
Following conversation with @memodi, this PR force skip TLS checks on both FLP & console plugin until implementation and allow configuration of Loki tenant ID

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -103,6 +103,9 @@ func buildArgs(desired *flowsv1alpha1.FlowCollectorConsolePlugin, desiredLoki *f
 		"-key", "/var/serving-cert/tls.key",
 		"-loki", querierURL(desiredLoki),
 		"-loki-labels", strings.Join(constants.LokiIndexFields, ","),
+		"-loki-tenant-id", desiredLoki.TenantID,
+		//TODO: add loki tls config https://issues.redhat.com/browse/NETOBSERV-309
+		"-loki-skip-tls", "true",
 		"-loglevel", desired.LogLevel,
 		"-frontend-config", configPath + configFile,
 	}
@@ -137,6 +140,9 @@ func (b *builder) podTemplate(cmDigest string) *corev1.PodTemplateSpec {
 					"-key", "/var/serving-cert/tls.key",
 					"-loki", querierURL(b.desiredLoki),
 					"-loki-labels", strings.Join(constants.LokiIndexFields, ","),
+					"-loki-tenant-id", b.desiredLoki.TenantID,
+					//TODO: add loki tls config https://issues.redhat.com/browse/NETOBSERV-309
+					"-loki-skip-tls", "true",
 					"-loglevel", b.desired.LogLevel,
 					"-frontend-config", configPath + configFile,
 				},
--- a/controllers/flowlogspipeline/flp_objects.go
+++ b/controllers/flowlogspipeline/flp_objects.go
@@ -16,6 +16,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	promConfig "github.com/prometheus/common/config"
+
 	flowsv1alpha1 "github.com/netobserv/network-observability-operator/api/v1alpha1"
 	"github.com/netobserv/network-observability-operator/controllers/constants"
 	"github.com/netobserv/network-observability-operator/pkg/helper"
@@ -243,6 +245,13 @@ func (b *builder) addTransformStages(lastStage *config.PipelineBuilderStage) {
 		lokiWrite.URL = b.desiredLoki.URL
 		lokiWrite.TimestampLabel = "TimeFlowEndMs"
 		lokiWrite.TimestampScale = "1ms"
+		lokiWrite.TenantID = b.desiredLoki.TenantID
+		//TODO: set proper tls config https://issues.redhat.com/browse/NETOBSERV-309
+		lokiWrite.ClientConfig = &promConfig.HTTPClientConfig{
+			TLSConfig: promConfig.TLSConfig{
+				InsecureSkipVerify: true,
+			},
+		}
 	}
 	enrichedStage.WriteLoki("loki", lokiWrite)
 
--- a/vendor/github.com/prometheus/common/config/http_config.go
+++ b/vendor/github.com/prometheus/common/config/http_config.go
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build go1.8
 // +build go1.8
 
 package config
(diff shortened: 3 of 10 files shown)


controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -103,6 +103,9 @@ func buildArgs(desired *flowsv1alpha1.FlowCollectorConsolePlugin, desiredLoki *f
 		"-key", "/var/serving-cert/tls.key",
 		"-loki", querierURL(desiredLoki),
 		"-loki-labels", strings.Join(constants.LokiIndexFields, ","),
+		"-loki-tenant-id", desiredLoki.TenantID,
+		//TODO: add loki tls config https://issues.redhat.com/browse/NETOBSERV-309
+		"-loki-skip-tls", "true",
```
jotak, line 108: this could also be promoted right now as a CR config, like you did for TenantID.

jpinsonneau: I can do that if you prefer but the goal is to remove this after TLS implementation

memodi:
> New image: ["quay.io/netobserv/network-observability-operator:6536b7c"]. It will expire after two weeks.

@jpinsonneau - after deploying NOO off this PR and above image, I am no longer seeing cert error on console side, however for FLP post request I am seeing below error:
```
time=2022-06-14T15:47:23Z level=info component=client error=Post "https://lokistack-sample-ingester-http.network-observability:3100/loki/api/v1/push": proxyconnect tcp: dial tcp :0: connect: connection refused fields.level=warn fields.msg=error sending batch, will retry host=lokistack-sample-ingester-http.network-observability:3100 module=export/loki status=-1
```

controllers/consoleplugin/consoleplugin_objects.go
jotak: I don't think it has to be removed, it happens commonly that people want to just quick try software without securing everything, so this is still a good option to have IMO, especially as people can "bring their own Loki". What matters more is that it defaults to `false`.

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -243,6 +245,13 @@ func (b *builder) addTransformStages(lastStage *config.PipelineBuilderStage) {
 		lokiWrite.URL = b.desiredLoki.URL
 		lokiWrite.TimestampLabel = "TimeFlowEndMs"
 		lokiWrite.TimestampScale = "1ms"
+		lokiWrite.TenantID = b.desiredLoki.TenantID
+		//TODO: set proper tls config https://issues.redhat.com/browse/NETOBSERV-309
+		lokiWrite.ClientConfig = &promConfig.HTTPClientConfig{
```
memodi, line 250: Is there an authentication that happens with distributor/ingester when we use loki-operator without gateway mode?

memodi:
Following up on my comment above: https://github.com/netobserv/network-observability-operator/pull/120#issuecomment-1155386800:

I am getting same error when I try to work with zero-click loki server:
`time=2022-06-14T20:51:16Z level=info component=client error=Post "http://loki:3100/loki/api/v1/push": proxyconnect tcp: dial tcp :0: connect: connection refused fields.level=warn fields.msg=error sending batch, will retry host=loki:3100 module=export/loki status=-1` 

not sure if there needs to be some updates to be made to FLP to be able to work with NOO created off this PR

controllers/flowlogspipeline/flp_objects.go
jpinsonneau: Loki doesn't manage authentication. It only force you to send `TenantID` when using TLS. That's why the gateway is a custom component :smile_cat:

jpinsonneau:
> @jpinsonneau - after deploying NOO off this PR and above image, I am no longer seeing cert error on console side, however for FLP post request I am seeing below error: ` time=2022-06-14T20:12:01Z level=info component=client error=Post "http://lokistack-sample-distributor-http.network-observability:3100/loki/api/v1/push": proxyconnect tcp: dial tcp :0: connect: connection refused fields.level=warn fields.msg=error sending batch, will retry host=lokistack-sample-distributor-http.network-observability:3100 module=export/loki status=-1`

Hm sorry about that ! It seems `HTTPClientConfig` `ProxyURL` is used even if it's not set on our side. Since the proxy is not resolved, the query fails.
I have tried to set `nil` or `""` but it doesn't work as is.

We didn't had this in the PoC because we used a `map[string]interface{}`

I'm still looking for a solution

jotak:
@jpinsonneau Could it be related to @OlivierCazade 's https://github.com/netobserv/flowlogs-pipeline/pull/229 ? (missing omitempty in FLP)

jpinsonneau:
> @jpinsonneau Could it be related to @OlivierCazade 's [netobserv/flowlogs-pipeline#229](https://github.com/netobserv/flowlogs-pipeline/pull/229) ? (missing omitempty in FLP)

No it actually comes from `prometheus` [http_config file](https://github.com/netobserv/network-observability-operator/pull/120/commits/d0354231d91bd586d2000a9d12732cb0911a093c#diff-7ff3442278d7dcfd4c9874e8e2e0f543749ba126b032a990fc3cb8f8e70a5c03R192)

memodi:
@jpinsonneau - I don't know if this PR needs rebase after  [this PR](https://github.com/netobserv/network-observability-operator/pull/114) was merged, since I am running into below error on FLP side:

time=2022-06-16T21:41:14Z level=debug msg=findStageType: stage = decode
time=2022-06-16T21:41:14Z level=fatal msg=failed to initialize pipeline invalid stage type: unknown

jpinsonneau:
> @jpinsonneau - I don't know if this PR needs rebase after [this PR](https://github.com/netobserv/network-observability-operator/pull/114) was merged, since I am running into below error on FLP side

Yeah seems to ! Let's rebase

jpinsonneau:
@memodi I have tested today with latest `flowlogs-pipeline` version and it works fine :+1:

memodi:
> @memodi I have tested today with latest `flowlogs-pipeline` version and it works fine 👍

Yes, it works for me too. Thank you. Although I couldn't see any flows on Netflow Table and always ends up in 503 error after trying to load :D 

image

I had extra small LokiStack, not sure if Loki from Loki-operator has poor performance than what we install with zero-click Loki config,

jpinsonneau:
Performances are better on my side with `LokiStack` @memodi 
I use gp2 AWS s3 bucket witth extra small config. Do you have a running cluster so I can check what happens ?

jotak:
/lgtm
cc @OlivierCazade once this is merged we'll need to rebase and fix conflicts on the TLS branch work

vendor/github.com/prometheus/common/config/http_config.go
```diff
@@ -188,13 +189,13 @@ type HTTPClientConfig struct {
 	// Authorization.CredentialsFile.
 	BearerTokenFile string `yaml:"bearer_token_file,omitempty" json:"bearer_token_file,omitempty"`
 	// HTTP proxy server to use to connect to the targets.
-	ProxyURL URL `yaml:"proxy_url,omitempty" json:"proxy_url,omitempty"`
+	ProxyURL *URL `yaml:"proxy_url,omitempty" json:"proxy_url,omitempty"`
```
OlivierCazade, line 192: We should be very carefull with this kind of fix here :
- it can be easily overridden
- from other part of the code, while it looks like we are using the official library, we are using a custom one

OlivierCazade: May be we should add a a comment here to remind us that we did change it.

jpinsonneau: I have added a test to ensure Marshal / Unmarshal works as expected until we fork / update promConfig

OlivierCazade (approved):

PR #132, opened 2022-07-12
Workaround for sampling=1 bugs

jotak:
By default, prevent setting sampling=1 for IPFIX ~~(automatically raised
to 2 in that case)~~
Flag "forceSampleAll" can be use to force sampling=1 (used at their own risks)

See bugs:
- https://bugzilla.redhat.com/show_bug.cgi?id=2103136
- https://bugzilla.redhat.com/show_bug.cgi?id=2104943

Change ipfix cache defaults:

- timeout decreased to 20s
- max-flows increased to 400

### DIFF
--- a/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
+++ b/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
@@ -1444,23 +1444,23 @@ spec:
                     format: int32
                     minimum: 0
                     type: integer
-                  forceAllowSamplingAll:
+                  forceSampleAll:
                     default: false
                     description: It is not recommended to sample all the traffic with
                       IPFIX, as it may generate cluster instability. If you REALLY
                       want to do that, set this flag to true. Use at your own risks.
+                      When it is set to true, the value of "sampling" is ignored.
                     type: boolean
                   sampling:
                     default: 400
-                    description: 'Sampling is the sampling rate on the reporter. 100
+                    description: Sampling is the sampling rate on the reporter. 100
                       means one flow on 100 is sent. To ensure cluster stability,
-                      it is not possible to sample every packets by setting 0 or 1:
-                      in that case, the value will be raised to 2. If you really want
-                      to sample every packet, which may put cluster stability at risk,
-                      set "forceAllowSamplingAll" to "true". Alternatively, you can
-                      use the eBPF Agent instead of IPFIX.'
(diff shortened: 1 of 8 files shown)


jotak:
cc @memodi

jotak:
good point, yeah, probably better to return an error

mariomac (approved):

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -1432,24 +1432,35 @@ spec:
...
                     description: Sampling is the sampling rate on the reporter. 100
-                      means one flow on 100 is sent. 0 means disabled.
+                      means one flow on 100 is sent. To ensure cluster stability,
+                      it is not possible to set a value below 2. If you really want
+                      to sample every packet, which may impact the cluster stability,
+                      refer to "forceSampleAll". Alternatively, you can use the eBPF
+                      Agent instead of IPFIX.
                     format: int32
-                    minimum: 0
+                    minimum: 2
```
jpinsonneau, line 1463: There are no issues on `eBPF` side so I would not force this minimum here.

Moreover, I would prefer a realistic minimum based on perf tests for each agent more than a workaround here. Thoughts ?

jotak: Sampling for ebpf and ipfix are two different configs, so no problem on that side.

I don't think we want to introduce minimums in general, ideally, users should be able to set whatever they want and be careful about what they do. But the issue we found is critical enough (like, put a cluster down) and happened quite consistently with ipfix' sampling=1, so that's why we take extra measures to prevent it to happen in that specific situation. The stability really isn't the same between sampling=1 and sampling=2.

jpinsonneau: Oh sorry I though it was a common field. I agree the user should understand what he does and not able to put the cluster down.

What about adding some numbered recommendations in the descriptions based on performance results as a followup task ?

jotak: yes, that's something we need to document anyway. Captured here: https://issues.redhat.com/browse/NETOBSERV-493

jotak:
I am going to merge this as I don't see moves on OVS side at the moment. We will rollback if it ends up being unnecessary. I'll also start releasing v0.1.4 candidate after that.

/approve

PR #134, opened 2022-07-19
Hack script to "refresh" OVS

jotak:
ie. reset its configuration so that it sends again the ipfix templates
goal is to speed up ipfix template sync

### DIFF
--- a/.mk/ocp.mk
+++ b/.mk/ocp.mk
@@ -31,3 +31,7 @@ ocp-run: ocp-undeploy ocp-deploy   ## OCP-deploy + run the operator locally
 	@echo "====> Running the operator locally"
 	go run ./main.go
 
+.PHONY: ocp-refresh-ovs
+ocp-refresh-ovs:
+	@echo "====> Re-applying OVS configuration to speed-up templates sync"
+	./hack/refresh-ovs.sh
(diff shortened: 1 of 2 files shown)


memodi:
I am curious, in what scenarios we might use this hack?

jotak:
@memodi it's when for instance you change a config that triggers FLP reconciling/restarting, then you may need to wait for up to 10 minutes before seeing flows again (because ovs needs to sync with flp for sending ipfix definitions). This hack should force this sync hence avoiding the potential 10-minutes delay.

mariomac (approved):
Just by curiosity. Has this unsynced scenario happened very frequently to you? If that happened to me, I unnoticed it (maybe it just happened in one node and I didn't realize it because flows from other nodes were still going on)

jotak:
yes it does happen frequently to me, I'm surprised if it doesn't to you!
for instance, when changing FLP log level, or switching kafka on and off, or whatever that changes FLP without changing OVS...
we can also track the error metric, it tells when there's missing template definitions

jotak:
Yes, 10min is a maximum, for me as well it can be less, but it's common to have to wait more than 2-3 minutes. Unlike the eBPF agent that starts sending data immediately

PR #136, opened 2022-07-20
NETOBSERV-397 Kafka TLS

jotak:
- New TLS api based on CM / Secret
- Helper functions to generate volumes for certificates
- new "make deploy-kafka-tls"

makefile: create kafka user for TLS

Add Kafka TLS to agent & transformer

At the moment thre's a manual hack to do in order to copy kafka secrets
in the netobserv privileged namespace. Without it, ebpf pods can't mount
/bin/bash: line 1: q: command not found

Update vendor flp


.mk/development.mk
```diff
@@ -24,18 +24,39 @@ undeploy-loki: ## Undeploy loki.
...
+.PHONY: deploy-kafka-tls
+deploy-kafka-tls:
+	@echo -e "\n==> Deploy Kafka with mTLS. Get more help on https://github.com/netobserv/documents/blob/main/kafka.md"
+	kubectl create namespace $(NAMESPACE)  --dry-run=client -o yaml | kubectl apply -f -
+	kubectl apply -f "https://strimzi.io/install/latest?namespace="$(NAMESPACE) -n $(NAMESPACE)
+	kubectl apply -f "https://raw.githubusercontent.com/netobserv/documents/main/examples/kafka/metrics-config.yaml" -n $(NAMESPACE)
+	kubectl apply -f "https://raw.githubusercontent.com/netobserv/documents/main/examples/kafka/tls.yaml" -n $(NAMESPACE)
+	kubectl apply -f "https://raw.githubusercontent.com/netobserv/documents/main/examples/kafka/topic.yaml" -n $(NAMESPACE)
+	kubectl apply -f "https://raw.githubusercontent.com/netobserv/documents/main/examples/kafka/user.yaml" -n $(NAMESPACE)
+	kubectl wait --timeout=180s --for=condition=ready kafkauser flp-kafka -n $(NAMESPACE)
```
OlivierCazade, line 43: Waiting for kafka to be ready is nice.
For consistency, I think we should have both deploy-kafka target to wait.

OlivierCazade (approved):
LGTM

.mk/development.mk
jotak: yeah that was especially useful here, because if kafka isn't ready while FLP is already deployed with secrets configured, pods are stuck and won't start because secret volumes can't be mounted. Even after the secrets are created, the pods are still not created  - I had to explicitly delete them.
but I agree, I'll do the same for the non-tls kafka

jotak: done

OlivierCazade (approved):
I forgot to mention in the review comment that the PR was tested and worked as expected.

jotak:
/retest
we need to investigate those flaky tests...

PR #137, opened 2022-07-21
Fix issues with doc/code/manifest generation

jotak:
Sometimes we had to run "generate" and/or "manifest" make targets twice
in order to have the doc generated. This should fix it.

Also, use the existing go-get-tool function to get crdoc, to unify with
existing tooling download method


mariomac (approved):

jotak:
thanks @mariomac 
/approve

PR #138, opened 2022-07-22
NETOBSERV-488 Manage FlowCollector readiness / conditions

jotak:
Create readiness conditions that can report:
- Any error occuring
- Updating components
- Deployments in progress
- Ready

This information is accessible in FlowCollector status

Also add new printed columns for CLI description


jotak:
Example of use:

* Watch for status change:

```bash
$ oc get flowcollector -w
NAME      AGENT   KAFKA   STATUS
cluster   ipfix   false   
cluster   ipfix   false   
cluster   ipfix   false   Updating
cluster   ipfix   false   DeploymentInProgress
cluster   ipfix   false   Ready
```

* Check errors:

```bash
$ oc get flowcollector
NAME      AGENT   KAFKA   STATUS
cluster   ipfix   false   ReconcileFLPFailed

$ oc get flowcollector cluster -o=jsonpath='{.status.conditions[*].message}'
Failed to reconcile flowlogs-pipeline: flowlogs-pipeline port value is not authorized
```

controllers/reconcilers/client_helper.go
```diff
@@ -54,6 +59,26 @@ func (c *ClientHelper) UpdateOwned(ctx context.Context, old, obj client.Object)
...
+func (c *ClientHelper) DidChange() bool {
+	return c.changed
+}
+
+func (c *ClientHelper) IsInProgress() bool {
+	return c.deplInProgress
+}
+
+func (c *ClientHelper) CheckDeploymentInProgress(d *appsv1.Deployment) {
+	if d.Status.AvailableReplicas < d.Status.Replicas {
```
jpinsonneau, line 71: What about the case where AvailableReplicas > d.Status.Replicas ? 

```suggestion
	if d.Status.AvailableReplicas != d.Status.Replicas {
```

controllers/reconcilers/client_helper.go
```diff
@@ -54,6 +59,26 @@ func (c *ClientHelper) UpdateOwned(ctx context.Context, old, obj client.Object)
...
+		c.deplInProgress = true
+	}
+}
+
+func (c *ClientHelper) CheckDaemonSetInProgress(ds *appsv1.DaemonSet) {
+	if ds.Status.NumberAvailable < ds.Status.DesiredNumberScheduled {
```
jpinsonneau, line 77: Same here 
```suggestion
	if ds.Status.NumberAvailable != ds.Status.DesiredNumberScheduled {
```

jpinsonneau (approved):
LGTM, tested and works fine :+1:

jpinsonneau:
Is final goal to link this with https://github.com/netobserv/network-observability-operator/pull/132 to show current sampling and custom status on invalid sampling number ?

controllers/reconcilers/client_helper.go
jotak: It's debatable but I think we can consider the status is ready/deployed when there are more replicas than necessary (ie. it is being downscaled). Like, it's answering the question "do I have _at least_ the desired scaling ?"
I'm ok to change if you/others think it would be better otherwise.

jotak:
> Is final goal to link this with #132 to show current sampling and custom status on invalid sampling number ?

Ah, not at all, I really didn't had these two issues connected in mind

controllers/reconcilers/client_helper.go
jpinsonneau: I'm not aware enough on the impact of a pod being removed. Can it generate issues like killing a current connection ?

If yes then we should manage it in the status since it's a quick win. If no let's keep it as is :+1:

jotak: I guess there could be some data loss. Ok, let's  commit:)

PR #139, opened 2022-07-25
Try fixing flaky tests

jotak:


jotak:
I hope it can fix a race in tests, where CR was not completely deleted before other tests run.
Adding this `Eventually` should make sure the CR is actually deleted (and will retry a couple of times when it isn't)

jpinsonneau (approved):
LGTM, I don't get the error after several local tests

jotak:
/approve
thanks @jpinsonneau

PR #140, opened 2022-07-26
NETOBSERV-395: Tls loki

OlivierCazade:
Add tls configuration to loki using normal TLS (not mTLS) and also supporting a flag to skip TLS checks.


controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -25,6 +25,7 @@ const configMapName = "console-plugin-config"
 const configFile = "config.yaml"
 const configVolume = "config-volume"
 const configPath = "/opt/app-root/"
+const lokiCerts = "kafka-certs"
```
jpinsonneau, line 28:
```suggestion
const lokiCerts = "loki-certs"
```

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -112,6 +113,55 @@ func buildArgs(desired *flowsv1alpha1.FlowCollectorConsolePlugin, desiredLoki *f
...
+			ReadOnly:  true,
+		}}
+
+	args := []string{
+		"-cert", "/var/serving-cert/tls.crt",
+		"-key", "/var/serving-cert/tls.key",
+		"-loki", querierURL(b.desiredLoki),
+		"-loki-labels", strings.Join(constants.LokiIndexFields, ","),
+		"-loki-tenant-id", b.desiredLoki.TenantID,
+		"-loki-skip-tls", "true",
```
jpinsonneau, line 152:
```suggestion
```

jpinsonneau: Do you also experience [this issue](https://issues.redhat.com/browse/NETOBSERV-485) on args order ?

OlivierCazade: Either I was lucky and it kept the same order, or I did not have the issue.

Using this configuration I was able to reach a loki with tls:
`    tls:
      enable: true
      insecureSkipVerify: true
`

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -47,14 +47,20 @@ spec:
         certFile: user.crt
         certKey: user.key
   loki:
-    url: 'http://loki:3100/'
+    url: 'http://loki.network-observability.svc:3100/'
```
jotak, line 50: did you need this change / does it makes anything easier?
The problem I see by inferring namespace is that if you change the `spec.namespace`, then this url is likely invalid.
By sticking with just `http://loki` we assume loki is deployed next to the other components

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -112,6 +113,55 @@ func buildArgs(desired *flowsv1alpha1.FlowCollectorConsolePlugin, desiredLoki *f
...
+		"-loglevel", b.desired.LogLevel,
+		"-frontend-config", configPath + configFile}
+
+	if b.desiredLoki != nil && b.desiredLoki.TLS.Enable {
```
jotak, line 156: Maybe I'm wrong but I think the `InsecureSkipVerify` and the CA cert shouldn't be mutually exclusive. The loki server might expect TLS encryption (thus needs CA cert), but still, `InsecureSkipVerify` can be used (in this case it is the client that skips certificate verification, but that doesn't mean it wouldn't encrypt communications).
At least that's my understanding. So if that's correct, there shouldn't be an `else` clause here for CA.

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -338,11 +343,20 @@ func (b *builder) addTransformStages(lastStage *config.PipelineBuilderStage) {
...
-				InsecureSkipVerify: true,
-			},
+		if b.desiredLoki != nil && b.desiredLoki.TLS.Enable {
+			if b.desiredLoki.TLS.InsecureSkipVerify {
+				lokiWrite.ClientConfig = &promConfig.HTTPClientConfig{
+					TLSConfig: promConfig.TLSConfig{
+						InsecureSkipVerify: true,
+					},
+				}
+			} else {
```
jotak, line 353: Same remark here, I don't think there should be a `else`

config/samples/flows_v1alpha1_flowcollector.yaml
OlivierCazade: Yes, the certificate generated by Openshift is only valid for the full name.

controllers/consoleplugin/consoleplugin_objects.go
OlivierCazade: InsecureSkipVerify does not mean not using TLS it means no checking that the certificate used by the remote server is properly signed.
The CA cert is used for the same reason, it is used to validate the remote server certificate.

config/samples/flows_v1alpha1_flowcollector.yaml
jotak: Ah I see, ok then

controllers/consoleplugin/consoleplugin_objects.go
jotak: > InsecureSkipVerify does not mean not using TLS it means no checking that the certificate used by the remote server is properly signed.

That's what I say :)

>  The CA cert is used for the same reason, it is used to validate the remote server certificate. 

But the CA cert contains the server public key. So if TLS is enabled then I think you _need_ this cert for encryption, even if InsecureSkipVerify is true

jotak: (If this is wrong then I need to fix that on Kafka side, because that's how I did)

jotak: Ok, I've read a little more about TLS and I think you're right :)

jotak (approved):

PR #141, opened 2022-07-26
NETOBSERV-182 Drop unused fields

jotak:
When using IPFIX, it's now possible to avoid storing lots of garbage by
turning dropUnusedFields to true (it actually defaults to true)

I saw about 30% less space used with this method, uncompressed (it's probably less with compression)

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -262,6 +262,10 @@ type FlowCollectorFLP struct {
 	//+kubebuilder:default:=true
 	// EnableKubeProbes is a flag to enable or disable Kubernetes liveness/readiness probes
 	EnableKubeProbes bool `json:"enableKubeProbes,omitempty"`
+
+	//+kubebuilder:default:=true
+	// Set true to drop fields that are known to be unused by OVS, in order to save storage space.
+	DropUnusedFields bool `json:"dropUnusedFields,omitempty"`
 }
 
 type FlowCollectorHPA struct {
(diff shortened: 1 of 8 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -262,6 +262,10 @@ type FlowCollectorFLP struct {
 	//+kubebuilder:default:=true
 	// EnableKubeProbes is a flag to enable or disable Kubernetes liveness/readiness probes
 	EnableKubeProbes bool `json:"enableKubeProbes,omitempty"`
+
+	//+kubebuilder:default:=true
+	// Set true to drop fields that are known to be unused by OVS, in order to save storage space.
+	DropUnusedFields bool `json:"dropUnusedFields,omitempty"`
```
jotak, line 268: @mariomac I'm adding this field in FLP section as this is where it has an impact, however following the discussion on API review, maybe it makes more sense to have it under "ipfix"? (because it only has effect with ipfix, it would be ignored when agent is ebpf

mariomac: Ok, when this PR is merged, I'll rebase my branch and incorporate these changes

mariomac (approved):
It looks good to me. Have you found any noticeable impact in the FLP CPU consumption?

jotak:
Some quick measurements (on the left before 19:09, it's using current `main` images; on the right after 19:11, it's with this PR)

- FLP CPU stays stable:
Capture d’écran du 2022-07-29 19-17-41

- FLP memory minor increase (~about 4MB increase per FLP pod) :
Capture d’écran du 2022-07-29 19-17-57

- Flows rate stable (not reaching capacity anyway)
Capture d’écran du 2022-07-29 19-20-12

api/v1alpha1/flowcollector_types.go
mariomac: Thinking it twice, I think this field could still be in the FLP section, as:
1. It's a configuration for FLP
2. Despite it's useless in EBPF, it could still have generic functionality if in the future we use other technologies/agents.

PR #142, opened 2022-07-27
Use refactored confgen for metrics

jotak:
Requires: https://github.com/netobserv/flowlogs-pipeline/pull/267

Opened as draft until the above is merged

cc @KalmanMeth there's a couple of little changes on top of what you added :

- Interface with confgen is reduced to only what is needed here: calling "ParseDefinition" and "GenerateTruncatedConfig". So we don't need anymore the intermediate temporary files
- Propagate errors to caller


controllers/flowlogspipeline/flp_test.go
```diff
@@ -560,24 +580,21 @@ func TestPipelineTraceStage(t *testing.T) {
...
 
-//go:embed test_metrics_definitions
-var TestFlpMetricsConfig embed.FS
-
 func TestMergeMetricsConfigurationNoIgnore(t *testing.T) {
 	assert := assert.New(t)
 
 	flp := getFLPConfig()
-	FlpMetricsConfigDir = "test_metrics_definitions"
-	FlpMetricsConfig = TestFlpMetricsConfig
```
KalmanMeth, line 594: The idea here was to use a well-known (unchanging) set of metrics for testing, and not the current version in metrics_definitions, which might change from time-to-time. If we use the regular metrics_definitions, they whenever we make changes to metrics_definitions we may also have to update the test.

KalmanMeth (approved):

controllers/flowlogspipeline/flp_test.go
jotak: Ok, but on the other hand, there's also some value in testing the actual embedded metrics (making sure it's correctly formed, etc.). Could be a bit more maintenance , but it also increases coverage

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -437,15 +413,21 @@ func (b *builder) buildPipelineConfig() ([]config.Stage, []config.StageParam) {
...
-func (b *builder) configMap() (*corev1.ConfigMap, string) {
-	stages, parameters := b.buildPipelineConfig()
+func (b *builder) configMap() (*corev1.ConfigMap, string, error) {
+	stages, parameters, err := b.buildPipelineConfig()
+	if err != nil {
+		return nil, "", err
+	}
 
 	config := map[string]interface{}{
 		"log-level": b.desired.LogLevel,
```
ronensc, line 433: `config.ConfigFileStruct` can be used instead of `map[string]interface{}`

ronensc (approved):

mariomac (approved):

controllers/flowlogspipeline/flp_objects.go
jotak: hmm there are some weird things happening here: `ConfigFileStruct` doesn't have `health`. Seems like the output of confgen isn't exactly the input of flp ?

jotak: also LogLevel exists in `ConfigFileStruct` but it seems unused (see `config.ParseConfig` => it doesn't fill log-level).
I guess some more cleanup should be done on FLP

ronensc: Looking at the code of FLP's main, it seems to me that FLP accepts the health port only from the command line arguments.
https://github.com/netobserv/flowlogs-pipeline/blob/d7656a24bb2417bfd2fec25f75fb1ee077c1b86e/cmd/flowlogs-pipeline/main.go
If I'm correct, I think it's a bug in FLP that it doesn't allow the health port to be set from the config file.

jotak: I think it works, as it uses viper which accepts a config file as a substitute to command line args

jotak:
/approve
@ronensc I'm doing a second PR to try fixing some flaky tests, I'll use the filepath.Join as part of it

PR #144, opened 2022-08-01
NETOBSERV-448: Moving agent type, ipfix and ebpf config into their own common subsection

mariomac:
Addressing this API review comment: https://github.com/openshift/api/pull/1218#discussion_r927657821


api/v1alpha1/flowcollector_types.go
```diff
@@ -23,11 +23,39 @@ import (
...
+	// work but require manual configuration.
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum:="IPFIX";"EBPF"
+	// +kubebuilder:validation:Required
+	Type AgentType `json:"type"`
+
+	// Settings related to IPFIX-based flow reporter when the "agent" property is set
+	// to "ipfix".
+	// +optional
+	IPFIX FlowCollectorIPFIX `json:"ipfix,omitempty"`
```
mariomac, line 51: Discriminated union documentations state that both IPFIX and EBPF fields should be optional pointers. However, I had to set them as values because otherwise we would loose all the default values and users would be forced to define either ipfix of ebpf sections.

https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions

config/samples/flows_v1alpha1_flowcollector_versioned.yaml
```diff
@@ -4,21 +4,22 @@ metadata:
...
-    logLevel: info
-    privileged: false
+  agent:
+    type: IPFIX
+    ipfix:
+      cacheActiveTimeout: 60s
+      cacheMaxFlows: 100
+      sampling: 400
+    ebpf:
+      image: 'quay.io/netobserv/netobserv-ebpf-agent:v0.1.3'
```
OlivierCazade, line 14: Are we checking the backward compatibility with this tagged image?
I just tested and I had errors, FLP being unable to connect to loki.

config/samples/flows_v1alpha1_flowcollector_versioned.yaml
```diff
@@ -4,21 +4,22 @@ metadata:
...
-    image: 'quay.io/netobserv/netobserv-ebpf-agent:v0.1.1'
-    imagePullPolicy: IfNotPresent
-    sampling: 0
-    cacheActiveTimeout: 5s
-    cacheMaxFlows: 1000
-    interfaces: []
-    excludeInterfaces: ["lo"]
-    logLevel: info
-    privileged: false
+  agent:
```
OlivierCazade, line 7: I think you also need to update flows_v1alpha1_flowcollector.yaml file

OlivierCazade (commented):
When modifying the sample files the PR is working correctly

config/samples/flows_v1alpha1_flowcollector_versioned.yaml
mariomac: Yeah we check backwards-compatibility but the image tag is wrong and we'll fix it.

Anyway, an error in the agent should not affect FLP connectivity.

Makefile
```diff
@@ -199,7 +205,7 @@ TMP_DIR=$$(mktemp -d) ;\
 cd $$TMP_DIR ;\
 go mod init tmp ;\
 echo "Downloading $(2)" ;\
-GOBIN=$(PROJECT_DIR)/bin go get $(2) ;\
+GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
```
jotak, line 208: this was addressed in @stleerh 's PR that is now merged

api/v1alpha1/flowcollector_types.go
```diff
@@ -482,7 +495,7 @@ type FlowCollectorStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:scope=Cluster
-//+kubebuilder:printcolumn:name="Agent",type="string",JSONPath=`.spec.agent`
+//+kubebuilder:printcolumn:name="Agent",type="string",JSONPath=`.spec.agent.type`
```
jotak, line 498: Do you know if, with discrimated union, there is a way to add a "sampling" column that shows either ebpf sampling or ipfix sampling?
I wanted to do that before, but couldn't find a way.

config/manager/kustomization.yaml
```diff
@@ -13,6 +13,7 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/netobserv/network-observability-operator
-  newTag: 0.1.3
+  newTag: main
 commonLabels:
   app: network-observability-operator
+  version: main
```
jotak, line 19: can you revert this file change ? I don't know why t added "version: main" on labels, it shouldn't. Anyway, that's a generated file.

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -29,7 +30,7 @@ spec:
     logLevel: info
     enableKubeProbes: true
     healthPort: 8080
-    prometheusPort: 9102
+    prometheusPort: 9090
```
jotak, line 33: I remember you mentioned this inconsistency.
I've just opened a PR to fix it, but the other way around, using 9102 in kubebuilder: https://github.com/netobserv/network-observability-operator/pull/149

jotak:
You will have merge conflict on all the generated files, I would advise to not commit them (I think, more or less, everything in `bundle`)

api/v1alpha1/flowcollector_types.go
mariomac: Hmmm i don't know the way. Even if they are name "union", at the end they follow different paths: `.spec.agent.ipfix` and `.spec.agent.ebpf`

config/manager/kustomization.yaml
mariomac: I know... that's probably because I used a more recent version of the generatior. I'll downgrade to the previous version.

mariomac:
Yes I'll expect until definitive version of 0.1.4 is merged then I'll fix all the conflicts and regenerate the manifests.

config/manager/kustomization.yaml
jotak: or i can upgrade mine, if this is due to operator-sdk version

mariomac:
Added do-not-merge label as we will remove support for ipfix

mariomac:
Closing this PR in favor of https://github.com/netobserv/network-observability-operator/pull/162

PR #145, opened 2022-08-03
Try fix flaky tests, + use filepath.Join

jotak:
Flaky tests are due to the new status conditions recently added to the CRD: the CR kinda auto-updates its status and could cause a race when tests also update the CR in the meantime, resulting in this error message: "the object has been modified; please apply your changes to the latest version and try again"

To fix that, this patch provide a new idiomatic way to update CR in
tests, where it will first fetch the current CR until it is ready
(assuming than its status won't change "by itself" once it's ready)


mariomac (approved):
Pretty elegant. Thanks!

OlivierCazade (approved):

ronensc (approved):

PR #146, opened 2022-08-03
NETOBSERV-505 NOO fails to build with go1.18 and above

stleerh:
Use 'go install' to install the executable. This stopped working in  
go1.18 because of https://go.dev/doc/go-get-install-deprecation.

In making this change, kustomize broke and no longer installed. The problem  
is documented in kubernetes-sigs/kustomize#3618.  
The solution is to use at least v4.5.2. kustomize was upgraded to 4.5.7  
from 3.8.8 and everything still worked.


jotak:
+1 but we need to make sure this major upgrade of kustomize doesn't have any wrong side effect on building bundle

jotak (approved):
LGTM, thanks @stleerh

PR #147, opened 2022-08-04
NETOBSERV-504 Fix console plugin bool flags usage

jotak:
Also remove duplicated code for plugin args


jotak:
@OlivierCazade there was some duplicated code left in arguments building, not sure if it was left on purpose? I merged them.

OlivierCazade (approved):
LGTM, and no that was not left on purpose, thanks for this

PR #153, opened 2022-08-11
NETOBSERV-497 Use direct-prom pipeline

jotak:
- Remove aggregation stage
- Update FLP / vendor

Requires https://github.com/netobserv/flowlogs-pipeline/pull/266


jotak:
/hold
requires https://github.com/netobserv/flowlogs-pipeline/pull/266

controllers/flowlogspipeline/metrics_definitions/bandwidth_per_network_service_per_namespace.yaml
```diff
@@ -11,22 +11,13 @@ tags:
...
   type: prom
   prom:
     metrics:
       - name: bandwidth_per_network_service_per_namespace
         type: counter
-        filter: {key: name, value: bandwidth_network_service_namespace}
-        valuekey: recent_op_value
+        valuekey: Bytes
         labels:
           - Service
```
KalmanMeth, line 22: If we no longer do the summation in the extract_aggregate, where is the sum performed? In prometheus counter? So we have to call Prometheus.Add() on each record instead of on the aggregates. OK.

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -376,14 +373,13 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 	}
 
 	// obtain extract_aggregate and encode_prometheus stages from metrics_definitions
-	aggregates, promMetrics, err := b.obtainMetricsConfiguration()
+	promMetrics, err := b.obtainMetricsConfiguration()
 	if err != nil {
 		return err
```
KalmanMeth, line 378: Fix comment that we add only encode_prometheus stage but not extract_aggregate.

KalmanMeth (approved):

KalmanMeth:
@jotak @jpinsonneau @OlivierCazade We have these metrics now in prometheus. How do we expose them to a user? Do we send them to our own prometheus instance or to the existing Openshift prometheus? Then, how do we show the graphs?

controllers/flowlogspipeline/metrics_definitions/bandwidth_per_network_service_per_namespace.yaml
jotak: exactly, it's built-in prom client

jotak:
> @jotak @jpinsonneau @OlivierCazade We have these metrics now in prometheus. How do we expose them to a user? Do we send them to our own prometheus instance or to the existing Openshift prometheus? Then, how do we show the graphs?

It depends on the operator flavour, if it's upstream or downstream. Upstream, the user needs to install prometheus (or configure cluster monitoring to scrape user workload targets). It's not documented in the operator yet because it's still kinda WIP (cf epic https://issues.redhat.com/browse/NETOBSERV-391 ), but there's some documentation there: https://github.com/netobserv/documents/blob/main/openshift.md#metrics
Also, if I remember correctly you created some make targets to deploy prometheus, so that's an option for users.

In the productized/downstream version, metrics will be available in cluster monitoring, hence visible from the openshift console.

In both case that's not something we're actively working on right now as it's not in 4.12 roadmap, it will certainly be in 4.13. But that doesn't mean we can't make progress on it if we have time for it.

And at last, we will create new dashboards in our console plugin (epic https://issues.redhat.com/browse/NETOBSERV-139 )

PR #154, opened 2022-08-18
NETOBSERV-101 - loki statusURL

jpinsonneau:
Following https://github.com/observatorium/api/pull/350 we will not have access to loki ready, config, metrics and build infos endpoints using the gateway.

This PR allow to configure a `statusURL` in the CRD and send it to the console plugin

Related PR: https://github.com/netobserv/network-observability-console-plugin/pull/186

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -300,8 +300,14 @@ type FlowCollectorLoki struct {
 	//+kubebuilder:validation:optional
 	// QuerierURL specifies the address of the Loki querier service, in case it is different from the
 	// Loki ingester URL. If empty, the URL value will be used (assuming that the Loki ingester
-	// and querier are int he same host).
+	// and querier are in the same host).
 	QuerierURL string `json:"querierUrl,omitempty"`
+
+	//+kubebuilder:validation:optional
+	// StatusURL specifies the address of the Loki /ready /metrics /config endpoints, in case it is different from the
+	// Loki querier URL. If empty, the QuerierURL value will be used.
+	// This is usefull to show error messages and some context in the frontend
+	StatusURL string `json:"statusUrl,omitempty"`
 
 	//+kubebuilder:default:="netobserv"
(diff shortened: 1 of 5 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -300,9 +300,15 @@ type FlowCollectorLoki struct {
...
 	// QuerierURL specifies the address of the Loki querier service, in case it is different from the
 	// Loki ingester URL. If empty, the URL value will be used (assuming that the Loki ingester
-	// and querier are int he same host).
+	// and querier are in the same host).
 	QuerierURL string `json:"querierUrl,omitempty"`
 
+	//+kubebuilder:validation:optional
+	// StatusURL specifies the address of the Loki /ready /metrics /config endpoints, in case it is different from the
+	// Loki querier URL. If empty, the QuerierURL value will be used.
+	// This is usefull to show error messages and some context in the frontend
```
jotak, line 309: typo: useful

api/v1alpha1/flowcollector_types.go
```diff
@@ -300,9 +300,15 @@ type FlowCollectorLoki struct {
 	//+kubebuilder:validation:optional
 	// QuerierURL specifies the address of the Loki querier service, in case it is different from the
 	// Loki ingester URL. If empty, the URL value will be used (assuming that the Loki ingester
-	// and querier are int he same host).
+	// and querier are in the same host).
```
jotak, line 303: nitpicking, but "same host" is not totally correct IMO, I would rephrase as "same server", no? (they can be on same host (node), but different servers / deployments)

jotak (approved):
almost lgtm, just a typo + small remark

jotak:
btw: what would be the URL to set when using the loki-operator with the gateway? is it documented?

jpinsonneau:
> btw: what would be the URL to set when using the loki-operator with the gateway? is it documented?

You need to point query frontend in that case. Check https://github.com/netobserv/documents/pull/22/files#diff-7f51acaa5430f57ab824677d12e76ad94983d7dd45c949b7cb1335991a019996R97

PR #155, opened 2022-08-22
NETOBSERV-435 - Set default req/limits resources for eBPF agent

jpinsonneau:
I set some minimalist resources spec for eBPF agent as already defined for others components.
This works fine on my dev cluster generating some spam on a httpd deployment:
- memory is around 30MiB
- max cpu at 84m 

Another discussion around what limits should be set for downstream is available in [NETOBSERV-380](https://issues.redhat.com/browse/NETOBSERV-380)
Requests / limits per cluster size should be documented for upstream in [NETOBSERV-493](https://issues.redhat.com/browse/NETOBSERV-493)

### DIFF
--- a/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
+++ b/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
@@ -136,6 +136,12 @@ spec:
                           PERFMON, NET_ADMIN, SYS_RESOURCE.'
                         type: boolean
                       resources:
+                        default:
+                          limits:
+                            memory: 100Mi
+                          requests:
+                            cpu: 100m
+                            memory: 50Mi
                         description: 'Compute Resources required by this container.
                           Cannot be updated. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                         properties:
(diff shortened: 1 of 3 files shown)


config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -772,6 +772,12 @@ spec:
                       SYS_RESOURCE.'
                     type: boolean
                   resources:
+                    default:
+                      limits:
+                        memory: 100Mi
+                      requests:
+                        cpu: 100m
+                        memory: 50Mi
```
memodi, line 780: I think these defaults for memory would be too low for eBPF. It might be okay for RSS but I have seen container_working_set_bytes for ebpf pods in the order of around 250-300 Mb.  It gets OOMKilled when container_working_set_bytes reaches limits.

memodi (commented):
default memory seems low for ebpf.

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
jpinsonneau: I guess we can raise that a bit but as said in the scrum today I think we should keep these values as low as possible upstream since we don't know the cluster size.
We need to add documentation to explain expected values per components per expected amount of flows so the user will be able to override this wisely.

Amoghrd:
Verified, works as expected!

PR #157, opened 2022-08-24
Workflows update

memodi:
This should help by creating subscription of NOO catalog image developed from main branch as an alternative to make deploy target

/cc @jotak @jpinsonneau


jpinsonneau (approved):

memodi:
@jotak I've some confusion around this workflows, when I create a catalog using the NOO image bundle created by this workflow, the PackageManifests still uses 0.1.4 i.e. version published to OperatorHub:

Screen Shot 2022-08-29 at 1 54 52 PM

Screen Shot 2022-08-29 at 1 55 08 PM

 any ideas why it uses 0.1.4 as version in PackageManifests? is it just some config issue and it actually would use the NOO image tagged as `main` ? Just want to make sure that when use catalog image bundle using that's pushed using this workflow actually using correct images.

memodi:
> As far as I can tell, that's because you call `make bundle-build` without prior call of `make bundle`, so the version that is used is still whatever is in `bundle/manifests/netobserv-operator.clusterserviceversion.yaml`.

Okay, does it only use the version name? does it actually deploy NOO image with that version, in this case 0.1.4?

jotak:
> Okay, does it only use the version name? does it actually deploy NOO image with that version, in this case 0.1.4?

Yes it will use 0.1.4 in the end. Running `VERSION=main make bundle-build` will only set the version of the bundle docker image but it doesn't affect the version of underlying components, so you will actually get the last release and not the current `main` build.

jotak:
@memodi I sent you a patch suggestion on your branch

jotak:
> thanks @jotak. I tried to run the workflow in my forked repo, however it failed for `make bundle-for-test` step since operator-sdk is not installed. Any ideas how we can include through GH Actions? I tried to setup using [this GH Action ](https://github.com/marketplace/actions/setup-operator-sdk-action) available at marketplace, but it seems buggy where it would not find newer operator-sdk versions.

oh, correct, I didn't think about that.. we should probably add a make target that installs operator-sdk

memodi:
> oh, correct, I didn't think about that.. we should probably add a make target that installs operator-sdk

alright, I added operator-sdk install to Makefile and also tested out on [my forked repo](https://github.com/memodi/network-observability-operator/runs/8138848009?check_suite_focus=true) , PTAL. thanks!

jotak (approved):

memodi:
thanks @jotak , I think it needs approved label for it to be able to merge?

PR #158, opened 2022-08-26
NETOBSERV-396: Flp prom tls

OlivierCazade:
Adding TLS to the FLP prometheus configuration.
This provide two options:
- using an automatic mode using openshift annotations
- letting the user provide its own certificates


config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -1404,14 +1404,87 @@ spec:
...
+                            type: object
+                        type: object
+                      port:
+                        default: 9090
+                        description: the prometheus HTTP port
+                        format: int32
+                        maximum: 65535
+                        minimum: 1
+                        type: integer
+                      tlsType:
```
jpinsonneau, line 1476: What's the difference between `tlsType: DISABLED` and `enable: false` ?

api/v1alpha1/flowcollector_types.go
```diff
@@ -201,6 +201,37 @@ type FlowCollectorKafka struct {
...
+	// and "AUTO" to use Openshift auto generated certificate using annotations
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum:="DISABLED";"MANUAL";"AUTO"
+	// +kubebuilder:validation:Required
+	//+kubebuilder:default:="DISABLED"
+	TLSType PrometheusTLSConfigType `json:"tlsType,omitempty"`
+
+	// TLS configuration.
+	// +optional
+	ManualTLS ClientTLS `json:"manualTls"`
```
jotak, line 232: Maybe we should not reuse `ClientTLS` here, and create a new struct, I believe it is not 100% relevant when used to configure tls on server side (or at least it should be renamed).
As @jpinsonneau pointed, `ClientTLS.enable` setting can be confusing as we have `TLSType` here. Also I believe InsecureSkipVerify is not relevant on server side?

api/v1alpha1/flowcollector_types.go
```diff
@@ -201,6 +201,37 @@ type FlowCollectorKafka struct {
 	TLS ClientTLS `json:"tls"`
 }
 
+const (
+	PrometheusTLSDiasbled = "DISABLED"
```
jotak, line 205: nit: typo "Disabled"

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -508,6 +539,48 @@ func (b *builder) service(old *corev1.Service) *corev1.Service {
 	return newService
 }
 
+func (b *builder) promService(old *corev1.Service) *corev1.Service {
```
jotak, line 542: I already didn't like much this pattern of "if service doesn't exist do ... else do ..." (I blame myself for that), now it spreads in the code, it's ok for this PR but at some point I'd like to see if we can have something more concise | readable | elegant ...

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
OlivierCazade: Thanks, I changed it and manualtls only contain cert configuration now.

api/v1alpha1/flowcollector_types.go
OlivierCazade: Yes, after Julien comment, I changed the code to use CertificateReference instead of ClientTLS

controllers/flowlogspipeline/flp_objects.go
OlivierCazade: I agree, but as your comment state this come from the immutable cluster ip field in the service.

I am curious to know how the kubectl cli works when updating a service. Does it first check existing service and enrich the new one? This would be strange since it add a lot of service specific code to kubectl.

api/v1alpha1/flowcollector_types.go
```diff
@@ -201,6 +201,37 @@ type FlowCollectorKafka struct {
...
+	//+kubebuilder:validation:Minimum=1
+	//+kubebuilder:validation:Maximum=65535
+	//+kubebuilder:default:=9102
+	// the prometheus HTTP port
+	Port int32 `json:"port,omitempty"`
+
+	// Select the type of TLS configuration
+	// "DISABLED" (default) to not configure TLS for the endpoint, "MANUAL" to manually provide cert file and a key file,
+	// and "AUTO" to use Openshift auto generated certificate using annotations
+	// +unionDiscriminator
```
jotak, line 224: cc @mariomac is that how `unionDiscriminator` works? Looking at [your PR](https://github.com/netobserv/network-observability-operator/pull/144/files#diff-8b68f9600ecc25d88f615075d5135fca8f7e520634d6ae84027efdbc023b643a), I would think that the `enum` values need to match the discriminated field names; in other words, here, the enum for `MANUAL` would have to be `ManualTLS` ?
or this is pure coincidence in your code (with IPFIX/EBPF) and it is not required?

jotak: PS I could find any doc about this `unionDiscriminator` annotation, do you have a link?

controllers/flowlogspipeline/flp_objects.go
jotak: i'm not sure either

api/v1alpha1/flowcollector_types.go
OlivierCazade: According to the documentation [here](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions)

> Discriminant values should be PascalCase and should be equivalent to the camelCase field name (json tag) of one member of the union

So having the enum and the field name matching is a recommendation, I will change the name here.

controllers/flowlogspipeline/flp_objects.go
```diff
@@ -510,6 +540,48 @@ func (b *builder) service(old *corev1.Service) *corev1.Service {
...
+	if old == nil {
```
mariomac, line 544: I see that later, `promService` method is invoked with a non-nil object and also explicitly `nil`.

Could be this method explicitly divided into two? e.g.:
```
func (b *builder) newPromService() *corev1.Service
func (b *builder) updatedPromService(old *corev1.Service) *corev1.Service
```

controllers/flowlogspipeline/flp_reconciler.go
```diff
@@ -123,8 +126,8 @@ func (r *FLPReconciler) GetServiceName(kafka *flowsv1alpha1.FlowCollectorKafka)
 }
 
 func (r *FLPReconciler) Reconcile(ctx context.Context, desired *flowsv1alpha1.FlowCollector) error {
-	for _, singleFlp := range r.singleReconcilers {
-		err := singleFlp.Reconcile(ctx, desired)
+	for i := 0; i < len(r.singleReconcilers); i++ {
```
mariomac, line 129: Maybe it's more "idiomatic":
```go
for i := range r.singleReconcilers
```
but do it as you prefer.

mariomac (commented):
Mostly LGTM. Just few comments open for discussion

api/v1alpha1/flowcollector_types.go
mariomac: yeah, in the PR @jotak  mention, it is done different,[ following the suggestion of an API review comment](https://github.com/openshift/api/pull/1218#discussion_r927636099), but double-checking the documentation examples, it seems the correct way is the way Olivier did.

controllers/flowlogspipeline/flp_reconciler.go
OlivierCazade: Yes, thanks

api/v1alpha1/flowcollector_types.go
```diff
@@ -201,6 +201,44 @@ type FlowCollectorKafka struct {
...
+	// +kubebuilder:validation:Enum:="DISABLED";"MANUAL";"AUTO"
+	// +kubebuilder:validation:Required
+	//+kubebuilder:default:="DISABLED"
+	Type PrometheusTLSConfigType `json:"type,omitempty"`
+
+	// TLS configuration.
+	// +optional
+	Manual *CertificateReference `json:"manual"`
```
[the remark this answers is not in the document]
OlivierCazade: According to the recommendation this field name should match with the name in the enum discriminator.
I will also set it to provided to match your other comment.

api/v1alpha1/flowcollector_types.go
```diff
@@ -201,6 +201,44 @@ type FlowCollectorKafka struct {
...
+
+type PrometheusTLSConfigType string
+
+// PrometheusTLS define the TLS configuration of Prometheus
+type PrometheusTLS struct {
+	// Select the type of TLS configuration
+	// "DISABLED" (default) to not configure TLS for the endpoint, "MANUAL" to manually provide cert file and a key file,
+	// and "AUTO" to use Openshift auto generated certificate using annotations
+	// +unionDiscriminator
+	// +kubebuilder:validation:Enum:="DISABLED";"MANUAL";"AUTO"
```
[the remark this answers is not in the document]
OlivierCazade: Changed, thanks!

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
mariomac: could it be `CERTIFICATE` is a valid option for the `tls` type field instead of `MANUAL` or `PROVIDED`?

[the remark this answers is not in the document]
OlivierCazade: I think having only CERTIFICATE would be a bit ambiguous since AUTO mode also use certificate, it is just that they are generated and not provided by the user.

jotak:
/lgtm

Tested with this ServiceMonitor (in "AUTO" mode) :
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: flowlogs-pipeline
  namespace: network-observability
spec:
  endpoints:
  - interval: 15s
    port: prometheus
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
  namespaceSelector:
    matchNames:
    - network-observability
  selector:
    matchLabels:
      app: flowlogs-pipeline
```

Two remarks:
1. We must figure out what's required to get rid of the `insecureSkipVerify: true` in ServiceMonitor and document it. I haven't dug much, but had a host name mismatch issue without it.
2. I just noticed we use the old "alpha" version of the cert annotation, whereas a "beta" now exists and is recommended. We should update everywhere we use it.

jotak:
Quick update: I'm trying with this ServiceMonitor and it doesn't work, I don't know why:

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: flowlogs-pipeline
  namespace: network-observability
spec:
  endpoints:
  - interval: 15s
    port: prometheus
    scheme: https
    tlsConfig:
#      insecureSkipVerify: true
      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
      serverName: flowlogs-pipeline-prom.network-observability.svc
  namespaceSelector:
    matchNames:
    - network-observability
  selector:
    matchLabels:
      app: flowlogs-pipeline
```

Whereas this works:
```bash
oc exec -n openshift-monitoring prometheus-k8s-0 -- curl https://flowlogs-pipeline-prom.network-observability.svc:9102/metrics --cacert /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
```

Which I believe should be an equivalent call...

OlivierCazade:
The CA cert file was already present in the prometheus pod?

jotak:
Yeah I expected the curl command would fail, but it worked :)
See also how other servicemonitors are configured: e.g. from `monitor-ovn-node`

```yaml
    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
      interval: 30s
      port: metrics
      scheme: https
      tlsConfig:
        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
        serverName: ovn-kubernetes-node.openshift-ovn-kubernetes.svc
```

They all use the same `service-ca.crt`, and adapt `serverName`.

OlivierCazade:
I got the service monitor working with this config:

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: flowlogs-pipeline
  namespace: network-observability
spec:
  endpoints:
  - interval: 15s
    port: prometheus
    scheme: https
    tlsConfig:
      # caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
      ca:
        configMap:
          name: "openshift-service-ca.crt"
          key: "service-ca.crt"
      serverName: flowlogs-pipeline-prom.network-observability.svc
  namespaceSelector:
    matchNames:
    - network-observability
  selector:
    matchLabels:
      app: flowlogs-pipeline
```

My understanding is that user workload metrics are handled by different pods in a different namespace. The contrary to the openshift monitoring pod the ca cert may not be already there.

PR #161, opened 2022-09-02
Add default limits to CR

jotak:
(it makes it easier to tune these limits by modifying the CR)

side question: should we change these limit? E.g. raise to, maybe, 800 Mi ?

### DIFF
--- a/config/samples/flows_v1alpha1_flowcollector.yaml
+++ b/config/samples/flows_v1alpha1_flowcollector.yaml
@@ -34,6 +34,12 @@ spec:
       port: 9102
     ignoreMetrics: []
     dropUnusedFields: true
+    resources:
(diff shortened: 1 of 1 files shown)


jpinsonneau:
If we do that for FLP, should we do it for all components ? 

We also need to keep synced this with defaults values from `api/v1alpha1/flowcollector_types.go`
FYI there is also a PR pending for missing config of eBPF agent: https://github.com/netobserv/network-observability-operator/pull/155

jotak:
@jpinsonneau I added the ebpf limits following what you did in your PR. We should merge them +/- together for consistency.
(I haven't tested mine yet, fussy cluster this morning :/ )

jpinsonneau (approved):

PR #163, opened 2022-09-06
NETOBSERV-448: reviewing API to follow conventions

mariomac:
* Modified fields' documentation to follow JSON naming, according to the [Openshift API Conventions' "Use JSON fields name in Godoc"](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#use-json-field-names-in-godoc)
* **Breaking change**: changed eBPF Agents' cacheMaxFlows default to 5000, according to the new default in the eBFP agent (1000 demonstrated to be too small for an idle 6-nodes cluster)
* **Breaking change**: in the `FlowCollector` spec, renamed `flowlogsPipeline` to `processor`


bundle/manifests/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -52,31 +52,29 @@ spec:
...
                     properties:
                       cacheActiveTimeout:
                         default: 5s
-                        description: CacheActiveTimeout is the max period during which
+                        description: cacheActiveTimeout is the max period during which
                           the reporter will aggregate flows before sending
                         pattern: ^\d+(ns|ms|s|m)?$
                         type: string
                       cacheMaxFlows:
-                        default: 1000
```
jpinsonneau, line 70: :+1:

api/v1alpha1/flowcollector_types.go
```diff
@@ -39,40 +39,42 @@ const (
...
 
+	// agent for flows' extraction.
 	// +kubebuilder:default:={type:"EBPF"}
 	Agent FlowCollectorAgent `json:"agent"`
 
-	// Settings related to the flowlogs-pipeline component, which collects and enriches the flows, and produces metrics.
-	FlowlogsPipeline FlowCollectorFLP `json:"flowlogsPipeline,omitempty"`
+	// processor defines the settings of the component that receives the flows from the agent,
+	// enriches them, and forwards them to the Loki persistence layer.
+	Processor FlowCollectorFLP `json:"processor,omitempty"`
```
jpinsonneau, line 53: :+1: sounds clearer for users

jpinsonneau (approved):
LGTM. Should we ping everyone in slack about the breaking changes before merge ?

api/v1alpha1/flowcollector_types.go
```diff
@@ -39,40 +39,42 @@ const (
 type FlowCollectorSpec struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
-	// Namespace where NetObserv pods are deployed.
+	// namespace where NetObserv pods are deployed.
 	// If empty, the namespace of the operator is going to be used.
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 
+	// agent for flows' extraction.
```
jotak, line 47: nit / typo: I don't think the quote is needed here? Should be just "gent for flows' extraction" ?
(If a native english speaker could confirm, cc @nalhadef / @stleerh ... )

jotak:
just a minor comment (cf above), else lgtm

api/v1alpha1/flowcollector_types.go
mariomac: Not 100% sure but I think you are right. Fixing it and we can revert it back if Neal/Steven or anyone else says it's required.

nalhadef: @mariomac @jotak I agree. The quotation marks should be removed.

PR #164, opened 2022-09-07
NETOBSERV-433 and NETOBSERV-373 - unify names

jotak:
- Use "netobserv" instead of "network-observability" (e.g. as default namespace)

Other required PRs:

- console: https://github.com/netobserv/network-observability-console-plugin/pull/191
- docs: https://github.com/netobserv/documents/pull/26

### DIFF
--- a/controllers/ebpf/internal/permissions/permissions.go
+++ b/controllers/ebpf/internal/permissions/permissions.go
@@ -72,7 +72,7 @@ func (c *Reconciler) reconcileNamespace(ctx context.Context) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: c.privilegedNamespace,
 			Labels: map[string]string{
-				"app":                                "network-observability-operator",
+				"app":                                constants.OperatorName,
 				"pod-security.kubernetes.io/enforce": "privileged",
 				"pod-security.kubernetes.io/audit":   "privileged",
 			},
--- a/controllers/flowcollector_controller_ebpf_test.go
+++ b/controllers/flowcollector_controller_ebpf_test.go
@@ -116,11 +116,11 @@ func flowCollectorEBPFSpecs() {
 				fmt.Sprintf("expected FLOWS_TARGET_HOST env var in %+v", spec.Containers[0].Env))
 
 			ns := v1.Namespace{}
-			By("expecting to create the network-observability-privileged namespace")
+			By("expecting to create the netobserv-privileged namespace")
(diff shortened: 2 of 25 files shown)


controllers/ebpf/internal/permissions/permissions.go
```diff
@@ -72,7 +72,7 @@ func (c *Reconciler) reconcileNamespace(ctx context.Context) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: c.privilegedNamespace,
 			Labels: map[string]string{
-				"app":                                "network-observability-operator",
+				"app":                                constants.OperatorName,
```
jpinsonneau, line 75: Should it be `netobserv-operator` or `netobserv` ?

controllers/flowcollector_controller_ebpf_test.go
```diff
@@ -116,11 +116,11 @@ func flowCollectorEBPFSpecs() {
 				fmt.Sprintf("expected FLOWS_TARGET_HOST env var in %+v", spec.Containers[0].Env))
 
 			ns := v1.Namespace{}
-			By("expecting to create the network-observability-privileged namespace")
+			By("expecting to create the netobserv-privileged namespace")
 			Expect(k8sClient.Get(ctx, nsKey, &ns)).To(Succeed())
 			Expect(ns.Labels).To(Satisfy(func(labels map[string]string) bool {
 				return helper.IsSubSet(ns.Labels, map[string]string{
-					"app":                                "network-observability-operator",
+					"app":                                constants.OperatorName,
```
jpinsonneau, line 123: Same here

jpinsonneau (approved):

controllers/ebpf/internal/permissions/permissions.go
jotak: Actually I never realized the privileged namespace had an app label ... @mariomac do you remember why is that? Is it needed?

jpinsonneau: As far as I remember it was just to follow [k8s recommendations](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)

We also talked about adding `app.kubernetes.io/part-of` & `app.kubernetes.io/version` in the past.

Maybe it's the right time to do it ?

mariomac: Yeah as part as I remember it was just for self-documentation metadata. Something like "this namespace belongs to this app", even if it is clear enough in the namespace name.

PR #165, opened 2022-09-08
NETOBSERV-549 Setup initial e2e tests

jotak:
- Copy & adapt tests from ebpf-agent e2e tests


jotak:
Draft / To do: deploy components from the operator, not individually

PR #167, opened 2022-09-15
Allow enabling profiler port in FLP

mariomac:
Related PR: https://github.com/netobserv/flowlogs-pipeline/pull/308

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -295,6 +295,12 @@ type FlowCollectorFLP struct {
 	// healthPort is a collector HTTP port in the Pod that exposes the health check API
 	HealthPort int32 `json:"healthPort,omitempty"`
 
+	//+kubebuilder:validation:Minimum=0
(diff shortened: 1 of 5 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -295,6 +295,12 @@ type FlowCollectorFLP struct {
 	// healthPort is a collector HTTP port in the Pod that exposes the health check API
 	HealthPort int32 `json:"healthPort,omitempty"`
 
+	//+kubebuilder:validation:Minimum=0
+	//+kubebuilder:validation:Maximum=65535
+	//+optional
+	// profilePort allows setting up a Go pprof profiler listening to this port
```
jotak, line 301: Maybe add: "Use only for debugging", or something along that line?

mariomac: As mentioned face-to-face last week, leaving this port open does not affect performance. It could have some impact just during the profile session, which is at most few seconds.

PR #168, opened 2022-09-21
Fix eBPF agent's sampling: 0 value to disabled

mariomac:
An explicit `sampling: 0` (disabled) value caused that the NOO interpreted this value as "unset", so it defaulted it to 50.


jotak:
Just to clarify: "disabled" means "monitor all packets" or the opposite, "turn off monitoring"?
If the former (which I guess it is), it is equivalent than setting sampling to 1, so another approach could be to set 1 as a minimum, and update the documentation accordingly ... what would you prefer?

mariomac:
@jotak good appreciation. By "disabled" i mean "disable the sampling", this is "monitor all packets". Same as setting it to 1.

I have no strong preferences, but I'd be fine keeping both 0 and 1 as the same behavior.

PR #169, opened 2022-09-21
NETOBSERV-555 NETOBSERV-552 CRD updates - many breaking changes

jotak:
Main items:
- NETOBSERV-555: "spec.flowlogsPipeline.kind" is removed, assuming
  DaemonSet
- processor.replicas/hpa are now
  kafka.consumerReplicas/consumerAutoscaler (hpa renamed autoscaler,
also for console)
- NETOBSERV-552: "ovn-kubernetes" and "cno" options moved into IPFIX section
- IPFIX documented as a legacy option
- NETOBSERV-544: remove kafka.enable option. Instead, use a top-level
  enum for deployment mode: DIRECT or KAFKA

Smaller changes:
- "flp.prometheus" option renamed to "flp.metricsServer", trying to be
  more agnositc (it's actually openmetrics/opentelemetry standard) and
not make user thinks it actually communicates with prometheus
- Refactored the main controller tests to adapt to new deployment
  options
- Make "FetchAll" logs less verbose


jotak:
You can check just the first commit (9599fd4) to review separately the FLP reconciler refactoring.

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -4,12 +4,19 @@ metadata:
   name: cluster
 spec:
   namespace: netobserv
+  deploymentType: DIRECT
```
mariomac, line 7: The name `deploymentType` sounds like a bit ambiguous to me. Maybe `communication`?

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -4,12 +4,19 @@ metadata:
...
   agent:
     type: EBPF
```
mariomac, line 9: Not 100% related to this task, but I was thinking in the case where a customer would like to override some properties of the eBPF agent, which is the only supported agent for downstream.

They should provide a yaml fragment like:

```yaml
agent:
  ebpf:
    sampling: 1
```

Being `ebpf` the only section they should use, wouldn't it be cleaner just to:
- remove agent.type
- move all the ebpf stuff as properties of the `agent` section
- try to find a "hidden" place for `ipfix` section that would allow overriding the eBPF agent without making our users to "pay" the price of a discriminated union.

Another option could be to move as properties of `agent` these properties that are common to all the agents (sampling, cacheMaxLenght, cacheMaxTimeout...) and keep some implementation details as subsections of `ebpf` (like the image, the image pull policy, etc...)

mariomac (commented):
Overall, great job! I have few comments

api/v1alpha1/flowcollector_types.go
```diff
@@ -215,34 +225,45 @@ type FlowCollectorKafka struct {
 	// tls client configuration.
 	// +optional
 	TLS ClientTLS `json:"tls"`
+
+	//+kubebuilder:validation:Minimum=0
+	//+kubebuilder:default:=1
+	// consumerReplicas defines the number of replicas (pods) to start for flowlogs-pipeline-transformer, which consumes Kafka messages.
+	// This setting is ignored when Kafka is disabled.
+	ConsumerReplicas int32 `json:"consumerReplicas,omitempty"`
```
OlivierCazade, line 233: I understand the logic of moving the field here, but I think it would be better to let it in the FLP section:
- while this fields only apply when kafka is enabled, this field apply to FLP. May be we could name it better in FLP like "KafkaConsumerReplicas"
- we would be more consitant with the console plugin section which also has this kind of field

api/v1alpha1/flowcollector_types.go
```diff
@@ -215,34 +225,45 @@ type FlowCollectorKafka struct {
...
+
+	// consumerAutoscaler spec of a horizontal pod autoscaler to set up for flowlogs-pipeline-transformer, which consumes Kafka messages.
+	// This setting is ignored when Kafka is disabled.
+	// +optional
+	ConsumerAutoscaler *FlowCollectorHPA `json:"consumerAutoscaler,omitempty"`
```
OlivierCazade, line 238: Same comment here

OlivierCazade (commented):
The feedback we had during the CRD API review was also to not use boolean in the API, are we going to change this part?

I only reviewed the CRD change not the impacted code. I will continue the review tomorrow.

config/samples/flows_v1alpha1_flowcollector.yaml
OlivierCazade: If sampling is common to both agent, can we move this part upper?

Something like :
```
agent:
  sampling: 1
  ebpf:
      ...
  ipfix:
      ...
```

OlivierCazade: I agree that this would make more sense.

OlivierCazade: (Commenting here for the third time because I am reading this too fast...)

My understanding is we are finally keeping both agent downstream but we are going to document clearly that only the ebpf agent is officially supported.

But may be we could move to the agent section the field that are relevent to both agent:
```
agent:
   sampling: 1
   ebpf:
      ...
   ipfix:
      ...
```

jotak: they are not shared fields, and I don't think we want make them shared, for instance because we want to have different default values (eBPF supports higher sampling rate than IPFIX for instance)

config/samples/flows_v1alpha1_flowcollector.yaml
jotak: I'm not fan of `deploymentType`, but maybe not `communication` either .. let's get more ideas :)

jpinsonneau: what about `optionnalComponents` where we can list `Kafka` but also `Prometheus` and even `Loki` in the future if it become optional ? 
It will be very flexible and would allow any new options.

If we really want to be explicit, we can simply use `components` with values like `Agent-Processor` and `Agent-Kafka-Processor`

mariomac: Some ideas: `wiring`, `connect`, `messaging`, `linking`...

mariomac: `transfer`...

api/v1alpha1/flowcollector_types.go
```diff
@@ -24,10 +24,16 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 const (
-	AgentIPFIX = "IPFIX"
-	AgentEBPF  = "EBPF"
+	AgentIPFIX           = "IPFIX"
+	AgentEBPF            = "EBPF"
+	DeploymentTypeDirect = "DIRECT"
+	DeploymentTypeKafka  = "KAFKA"
```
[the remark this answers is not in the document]
jotak: I think we will need to organize a vote, because everyone has different ideas :)

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
jotak: I can rollback it, but it was to avoid acronyms that maybe not everybody knows. and the full size "horizontalPodAutoscaler" is.. very long :/

jotak:
I'll review the boolean fields as @OlivierCazade and @stleerh pointed out

api/v1alpha1/flowcollector_types.go
jotak: done

config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -1658,7 +1428,251 @@ spec:
...
+                              or secret'
+                            enum:
+                            - configmap
+                            - secret
+                            type: string
+                        type: object
+                      enable:
+                        default: false
+                        description: enable TLS
+                        type: boolean
```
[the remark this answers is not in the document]
jotak: i'll check the remaining booleans in a follow-up

OlivierCazade (approved):
LGTM, thanks!

mariomac (approved):

PR #170, opened 2022-09-23
Expose new Kafka parameters + tweak defaults

jotak:
Follow-up on https://github.com/netobserv/flowlogs-pipeline/pull/309

- PullQueueCapacity: capacity of the kafka consumer client internal
  queue
- PullMaxBytes: indicates to the broker the maximum batch size that the consumer will accept.

**Update**
[Add producerBatchSize, add metric, tweak defaults](https://github.com/netobserv/network-observability-operator/pull/170/commits/3fc9186b02f61ca405f036b590cb7f0b9f364b1d)
[3fc9186](https://github.com/netobserv/network-observability-operator/pull/170/commits/3fc9186b02f61ca405f036b590cb7f0b9f364b1d)

New metric "bandwidth_per_namespace" allows to get flows sizes (ie.
extracted Bytes sum) per source and dest namespace
That proves useful for self-monitoring, eg. to get the overhead of
netobserv's traffic compared to the rest of the traffic.

Also, setting a much higher cacheMaxSize for the ebpf agent, to reduce
traffic overhead.

Also raising limits for agent+FLP to 800MB

And tweaking loki client default to decrease number of retries and increase batch size

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -155,7 +155,7 @@ type FlowCollectorEBPF struct {
 	// imagePullPolicy is the Kubernetes pull policy for the image defined above
 	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
 
-	//+kubebuilder:default:={requests:{memory:"50Mi",cpu:"100m"},limits:{memory:"100Mi"}}
+	//+kubebuilder:default:={requests:{memory:"50Mi",cpu:"100m"},limits:{memory:"800Mi"}}
 	// resources are the compute resources required by this container.
 	// Cannot be updated.
 	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
@@ -175,7 +175,7 @@ type FlowCollectorEBPF struct {
 
 	// cacheMaxFlows is the max number of flows in an aggregate; when reached, the reporter sends the flows
 	//+kubebuilder:validation:Minimum=1
-	//+kubebuilder:default:=5000
+	//+kubebuilder:default:=100000
 	CacheMaxFlows int32 `json:"cacheMaxFlows,omitempty"`
 
 	// interfaces contains the interface names from where flows will be collected. If empty, the agent
@@ -208,6 +208,11 @@ type FlowCollectorEBPF struct {
 	// BPF, PERFMON, NET_ADMIN, SYS_RESOURCE.
 	// +optional
 	Privileged bool `json:"privileged,omitempty"`
+
+	//+kubebuilder:default:=10485760
+	// +optional
+	// kafkaBatchSize limits the maximum size of a request in bytes before being sent to a partition. Ignored when not using Kafka. Default: 10MB.
+	KafkaBatchSize int `json:"kafkaBatchSize"`
 }
 
 // FlowCollectorKafka defines the desired Kafka config of FlowCollector
@@ -310,7 +315,7 @@ type FlowCollectorFLP struct {
 	// logLevel of the collector runtime
 	LogLevel string `json:"logLevel,omitempty"`
 
-	//+kubebuilder:default:={requests:{memory:"100Mi",cpu:"100m"},limits:{memory:"300Mi"}}
+	//+kubebuilder:default:={requests:{memory:"100Mi",cpu:"100m"},limits:{memory:"800Mi"}}
 	// resources are the compute resources required by this container.
 	// Cannot be updated.
 	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
@@ -335,6 +340,16 @@ type FlowCollectorFLP struct {
 	// This setting is ignored when Kafka is disabled.
 	// +optional
 	KafkaConsumerAutoscaler *FlowCollectorHPA `json:"kafkaConsumerAutoscaler,omitempty"`
+
+	//+kubebuilder:default:=1000
+	// +optional
+	// kafkaConsumerQueueCapacity defines the capacity of the internal message queue used in the Kafka consumer client. Ignored when not using Kafka.
+	KafkaConsumerQueueCapacity int `json:"kafkaConsumerQueueCapacity"`
+
+	//+kubebuilder:default:=10485760
+	// +optional
+	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
+	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
 }
 
 type FlowCollectorHPA struct {
--- a/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
+++ b/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
@@ -66,7 +66,7 @@ spec:
                         pattern: ^\d+(ns|ms|s|m)?$
                         type: string
                       cacheMaxFlows:
-                        default: 5000
+                        default: 100000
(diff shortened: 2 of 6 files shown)


jotak:
@mariomac @OlivierCazade @stleerh  here I just expose the new parameters from the kafka client that Olivier added to FLP, maybe we should expose more, what do you think?

There's for instance BatchMaxLen, which impacts on batching within FLP pipeline. I'm not sure if it's interesting to tweak, as these batches are not sent over the network (bigger batches on that won't decrease network usage as far as I can tell). Also, if we decide to expose that, I wonder if we could then tie kafka's `PullMaxBytes` with that (the former counts in number of flows, the latter in number of bytes, flows have approx. a constant size so we may make a correlation)

jotak:
Also, this is only for the consumer. What about the producer? Any tweaks to expose?

api/v1alpha1/flowcollector_types.go
```diff
@@ -215,6 +215,14 @@ type FlowCollectorKafka struct {
 	// tls client configuration.
 	// +optional
 	TLS ClientTLS `json:"tls"`
+
+	//+kubebuilder:default:=100
+	// pullQueueCapacity defines the capacity of the internal message queue used in the Kafka consumer client.
+	PullQueueCapacity int `json:"pullQueueCapacity"`
```
jotak, line 221: maybe we should rather name it `ConsumerQueueCapacity` ? (that would be more consistent with the other CRD update PR)

api/v1alpha1/flowcollector_types.go
```diff
@@ -215,6 +215,14 @@ type FlowCollectorKafka struct {
...
+
+	//+kubebuilder:default:=1000000
+	// pullMaxBytes indicates to the broker the maximum batch size that the consumer will accept. Default: 1MB.
+	PullMaxBytes int `json:"pullMaxBytes"`
```
jotak, line 225: Same here: `ConsumerMaxBytes` ?

mariomac:
`BatchMaxBytes` is certainly confusing. It could currently have an effect in the Loki load (as it acts as a retainer in the pipeline). In the [PR I'm working](https://github.com/netobserv/flowlogs-pipeline/pull/313), this argument won't have any effect when you combine Kafka+Protobuf, as messages already come batched from the Agent.

As they mention, it's important to be able to setup the Kafka's `batch.size` (I initially thought it was set by `BatchMaxBytes` 😅) and the `linger.ms`. However I think linger can impact more in the producer side. I double checked that the [eBPF agent allows setting the batch size but not the linger](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/agent/config.go) so I will add this property as part of my current kafka optimizations there.

mariomac:
Apart of these values being configured at the broker side, as currently specified, we'd need to allow specifying these per producer and consumer, as different producer/consumption patterns would require different batchSize/linger to optimize CPU and consumption.

api/v1alpha1/flowcollector_types.go
mariomac: I think we need to specify QueueCapacity, BatchSize and LingerMS as a subsection of each component: for the agent and for FLP.

BTW I think LingerMS in only for Producers (the agent, in this case)

mariomac:
**UPDATE**: I reminded that our Golang library, Segmentio's Kafka-Go, doesn't allow setting the linger value. So we will have to play only with the max batch size.

jotak:
About linger: it seems indeed to be lacking in segmentio's api, however, is that really an issue since we already have our own batch/timeout implemented in the agent for grouping flows?

mariomac:
@jotak I think we can just avoid linger configuration.

api/v1alpha1/flowcollector_types.go
```diff
@@ -215,6 +215,21 @@ type FlowCollectorKafka struct {
 	// tls client configuration.
 	// +optional
 	TLS ClientTLS `json:"tls"`
+
```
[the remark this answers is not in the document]
jotak: ok, sounds like a good argument, especially as we plan to allow kafka exports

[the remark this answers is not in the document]
jotak: @mariomac done

stleerh:
> Also, this is only for the consumer. What about the producer? Any tweaks to expose?

See https://issues.redhat.com/browse/NETOBSERV-593.

vendor/github.com/netobserv/flowlogs-pipeline/pkg/config/stage_params.go
```diff
@@ -53,6 +53,10 @@ func NewConnTrackParams(name string, ct api.ConnTrack) StageParam {
 	return StageParam{Name: name, Extract: &Extract{Type: api.ConnTrackType, ConnTrack: &ct}}
 }
 
+func NewTimbasedParams(name string, ct api.ExtractTimebased) StageParam {
```
stleerh, line 56: Misspelling: NewTimebasedParams

vendor/github.com/netobserv/flowlogs-pipeline/pkg/confgen/flowlogs2metrics_config.go
```diff
@@ -78,6 +83,8 @@ func (cg *ConfGen) GenerateTruncatedConfig() []config.StageParam {
 			parameters[i] = config.NewTransformNetworkParams("transform_network", *cg.config.Transform.Network)
 		case "extract_aggregate":
 			parameters[i] = config.NewAggregateParams("extract_aggregate", cg.aggregateDefinitions)
+		case "extract_timebased":
+			parameters[i] = config.NewTimbasedParams("extract_timebased", cg.timebasedTopKs)
```
stleerh, line 87: Misspelling: NewTimebasedParams

stleerh:
It would be good if there was a generic way to pass any config settings to Kafka as an advanced feature.  Then it could expose the more common ones only.

vendor/github.com/netobserv/flowlogs-pipeline/pkg/confgen/flowlogs2metrics_config.go
jotak: this is in vendors, so that's actually in FLP repo

jotak:
> It would be good if there was a generic way to pass any config settings to Kafka as an advanced feature.  Then it could expose the more common ones only.

yeah, I don't know exactly what's the best way to do that, at some point I was wondering if we shouldn't have a second CRD for advanced configuration .. the point being to try keeping things simple for the average user.

mariomac:
With respect to your conversation @jotak , @stleerh : would it be possible then just to have an "extraConfig" field that is just backed by a map that passes all its entries as properties to Kafka? Something similar to the [env](https://github.com/netobserv/network-observability-operator/blob/521445bf4bad6b9bf3c0fcb1fafb5cbed6ae3b0f/api/v1alpha1/flowcollector_types.go#L188) property of the eBPF agent.

jotak:
PR rebased

@mariomac are you thinking exposing that *instead of* having these new fields in the CR? (hence we would reject this PR?) 
That's probably a topic we should discuss in meeting

We can also take a look at how the loki-operator works, seems it also allow overriding some settings, I'm curious if their approach differs

mariomac (approved):

mariomac:
@jotak I'm not against exposing the values via CRD, but I think that at the mid term it's better to just allow overriding some internals, so we don't need to create a patch and release a new version every time we want to tweak some internals for a given customer.

However, allow overriding the FLP configuration might be complex, as it is not just a properties map but a structured and connected graph

PR #172, opened 2022-09-26
NETOBSERV-473 - Loki and strimzi operator installation

jpinsonneau:
This PR adds an option for both loki & kafka to install dependent operators automatically using:
```yaml
  operatorsAutoInstall: ["kafka", "loki"]
  kafka:
    autoInstallSpec:
      replicas: 3
      storage:
        type: persistent-claim
        size: 200Gi
        class: gp2
      zooKeeperReplicas: 3
      zooKeeperStorage:
        type: persistent-claim
        size: 20Gi
        class: gp2
      partitions: 24
      topicReplicas: 3
  loki:
    autoInstallSpec:
      secretName: loki-secret
      objectStorageType: s3
      size: 1x.extra-small
      storageClassName: gp2
      retentionDays: 1
```

It create subscription based on the [environment](https://github.com/netobserv/network-observability-operator/pull/172/files#diff-30add0122a3cbfd6e0941c85cb0ca2c39b2f5af1a21682ac75504044aa1d4a47R24) currently forced to `openshift` for testing.
Then it apply related yaml from `controllers/operators/embed/` folder overriding optional configuration provided in `instanceSpec`.

image

/!\ you still need to manage some manual tasks:
- create `loki-secret` for storage access
- ~~apply loki role ([currently returning an error](https://github.com/netobserv/network-observability-operator/pull/172/files#diff-30add0122a3cbfd6e0941c85cb0ca2c39b2f5af1a21682ac75504044aa1d4a47R150-R152) since controller service account doesn't have these roles)~~ 
loki roles are automatically applied https://github.com/netobserv/network-observability-operator/pull/172/commits/552e3f55b32fac82da0d83bd7dcb5123fcb632b1
- ~~[copy kafka secrets](https://issues.redhat.com/browse/NETOBSERV-639) in `netobserv-privileged` namespace (make fix-ebpf-kafka-tls)~~ 
kafka secret are automatically copied https://github.com/netobserv/network-observability-operator/pull/172/commits/55dbebfc81b34dece983868891b29cd3f9b102db
- use custom catalog for `loki-operator` until release 5.6.x (Nov 15th)

Create the catalog using the following yaml:
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: loki
  namespace: openshift-marketplace
spec:
  displayName: Loki Operator Catalog
  image: 'quay.io/jpinsonn/loki-operator-catalog:v0.0.1'
  publisher: jpinsonn
  sourceType: grpc
```
Then specify it in our CRD:
```yaml
  loki:
    autoInstallSpec:
      source: loki
```


KalmanMeth:
We should keep in mind that we will also probably need a similar prometheus operator installation to expose the flp metrics.

api/v1alpha1/flowcollector_types.go
```diff
@@ -219,19 +222,63 @@ type FlowCollectorEBPF struct {
 type FlowCollectorKafka struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
+	// InstanceSpec is the Spec section of LokiStack instance.
```
mariomac, line 225: To follow the API guidelines:
```suggestion
	// instanceSpec is the Spec section of LokiStack instance.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -369,20 +416,30 @@ type FlowCollectorHPA struct {
 
 // FlowCollectorLoki defines the desired state for FlowCollector's Loki client
 type FlowCollectorLoki struct {
+	// InstanceSpec is the Spec section of LokiStack instance.
```
mariomac, line 419:
```suggestion
	// instanceSpec is the Spec section of LokiStack instance.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
 	// tls client configuration.
+	// it will be ignored if instanceSpec is specified
 	// +optional
 	TLS ClientTLS `json:"tls"`
 }
 
+// LokiInstanceSpec defines the desired state for LokiStack instance configuration
+type LokiInstanceSpec struct {
+	//+kubebuilder:default:=false
+	// Enable automatic loki-operator installation
```
mariomac, line 496:
```suggestion
	// enable automatic loki-operator installation
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	// This will create an OperatorHub subscription and instance according to your platform
+	Enable bool `json:"enable,omitempty"`
+
+	// Name of a secret in the same namespace as the LokiStack custom resource
```
mariomac, line 500:
```suggestion
	// secretName is the name of a secret in the same namespace as the LokiStack custom resource
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	SecretName string `json:"secretName"`
+
+	// Type of object storage that should be used
```
mariomac, line 506:
```suggestion
	// objectStorageType is the type of object storage that should be used
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=azure;gcs;s3;swift
+	ObjectStorageType string `json:"objectStorageType"`
+
+	// ManagementState defines the type for CR management states.
```
mariomac, line 513:
```suggestion
	// managementState defines the type for CR management states.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	// Default is managed.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=Managed
+	// +kubebuilder:validation:Enum=Managed;Unmanaged
+	ManagementState string `json:"managementState,omitempty"`
+
+	// Size defines one of the support Loki deployment scale out sizes.
```
mariomac, line 522:
```suggestion
	// size defines one of the support Loki deployment scale out sizes.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum="1x.extra-small";"1x.small";"1x.medium"
+	Size string `json:"size"`
+
+	// Storage class name defines the storage class for ingester/querier PVCs.
```
mariomac, line 529:
```suggestion
	// storageClassName defines the storage class for ingester/querier PVCs.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	StorageClassName string `json:"storageClassName"`
+
+	// ReplicationFactor defines the policy for log stream replication.
```
mariomac, line 535:
```suggestion
	// replicationFactor defines the policy for log stream replication.
```

api/v1alpha1/flowcollector_types.go
```diff
@@ -427,10 +485,70 @@ type FlowCollectorLoki struct {
...
+	//
+	// +required
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum:=1
+	ReplicationFactor int32 `json:"replicationFactor"`
+
+	// RetentionDays contains the number of days logs are kept.
```
mariomac, line 543:
```suggestion
	// retentionDays contains the number of days logs are kept.
```

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -71,6 +77,13 @@ spec:
         certFile: user.crt
         certKey: user.key
   loki:
+    instanceSpec:
+      enable: false
+      secretName: loki-secret
+      objectStorageType: s3
+      size: 1x.extra-small
+      storageClassName: gp2
+      retentionDays: 1
```
mariomac, line 86: Must it be in days? maybe if the storage is too high, some users might want to set the retention to hours.

config/manager/kustomization.yaml
```diff
@@ -12,7 +12,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/netobserv/network-observability-operator
-  newTag: 0.1.4
+  newName: quay.io/jpinsonn/network-observability-operator
+  newTag: dev
```
mariomac, line 16:
```suggestion
  newName: quay.io/netobserv/network-observability-operator
  newTag: 0.1.4
```

mariomac (commented):
I tried but the following pods are stuck in `ContainerCreating` status:

* netobserv-ebpf-agent
* flowlogs-pipeline-transformer
* netobserv-plugin

Due to this errors in the description's events list:

```
  Normal   Scheduled    2m53s                default-scheduler  Successfully assigned netobserv-privileged/netobserv-ebpf-agent-99tzk to ip-10-0-144-117.ec2.internal by ip-10-0-169-224
  Warning  FailedMount  50s                  kubelet            Unable to attach or mount volumes: unmounted volumes=[kafka-certs-ca kafka-certs-user], unattached volumes=[kube-api-access-2zbnc kafka-certs-ca kafka-certs-user]: timed out waiting for the condition
  Warning  FailedMount  46s (x9 over 2m53s)  kubelet            MountVolume.SetUp failed for volume "kafka-certs-ca" : secret "kafka-cluster-cluster-ca-cert" not found
  Warning  FailedMount  46s (x9 over 2m53s)  kubelet            MountVolume.SetUp failed for volume "kafka-certs-user" : secret "flp-kafka" not found

  Warning  FailedMount  4m10s (x7 over 4m41s)  kubelet            MountVolume.SetUp failed for volume "kafka-certs-ca" : secret "kafka-cluster-cluster-ca-cert" not found
  Warning  FailedMount  3m38s (x8 over 4m41s)  kubelet            MountVolume.SetUp failed for volume "loki-certs-ca" : configmap "lokistack-ca-bundle" not found
```

To be addressed in another PR/task: There are some ugly stacktraces in the manager logs due to trying to reconcile elements while the related CRD is still not applied:


```
1.6649732503408134e+09	INFO	controller.flowcollector	checking for lokistack in ns netobserv ...	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "component": "ClientHelper", "function": "ApplyWithNamespaceOverride"}
1.6649732503408515e+09	ERROR	controller.flowcollector	Can't apply embed/loki_instance.yaml yaml	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "component": "OperatorsController", "function": "manageOperator", "error": "no matches for kind \"LokiStack\" in version \"loki.grafana.com/v1\""}
github.com/netobserv/network-observability-operator/controllers/operators.(*Reconciler).Reconcile
	/opt/app-root/controllers/operators/operators_reconciler.go:124
github.com/netobserv/network-observability-operator/controllers.(*FlowCollectorReconciler).Reconcile
	/opt/app-root/controllers/flowcollector_controller.go:165
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2




1.66497323313885e+09	ERROR	controller.flowcollector	Can't apply embed/loki_instance.yaml yaml	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "component": "OperatorsController", "function": "manageOperator", "error": "no matches for kind \"LokiStack\" in version \"loki.grafana.com/v1\""}
github.com/netobserv/network-observability-operator/controllers/operators.(*Reconciler).Reconcile
	/opt/app-root/controllers/operators/operators_reconciler.go:124
github.com/netobserv/network-observability-operator/controllers.(*FlowCollectorReconciler).Reconcile
	/opt/app-root/controllers/flowcollector_controller.go:165
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
```

config/samples/flows_v1alpha1_flowcollector.yaml
jpinsonneau: Yes it is in days on loki-operator side. We can sync with logging team if we really need less

api/v1alpha1/flowcollector_types.go
```diff
@@ -219,19 +222,63 @@ type FlowCollectorEBPF struct {
 type FlowCollectorKafka struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
+	// InstanceSpec is the Spec section of LokiStack instance.
+	// Experimental feature. If provided, the Loki operator will be automatically installed
+	// using operator lifecycle manager (https://olm.operatorframework.io/)
+	// It will create a "Subscription" using "operators.coreos.com/v1alpha1"
+	// and "Kafka" instance, "KafkaTopic", "KafkaUser" using "kafka.strimzi.io/v1beta2"
+	InstanceSpec *KafkaInstanceSpec `json:"instanceSpec,omitempty"`
```
mariomac, line 230: `instanceSpec` seems something too generic and it's difficult to know what refers to.

I'd add something more related with the intention of let the NOO to deploy it, like: `autoDeploy`, `autoCreate`, `fromOperator`....

api/v1alpha1/flowcollector_types.go
```diff
@@ -219,19 +222,63 @@ type FlowCollectorEBPF struct {
...
 	// tls client configuration.
+	// it will be ignored if instanceSpec is specified
 	// +optional
 	TLS ClientTLS `json:"tls"`
 }
 
+type KafkaInstanceSpec struct {
+	//+kubebuilder:default:=false
+	// Should this feature be enabled
+	Enable bool `json:"enable,omitempty"`
```
mariomac, line 251: Thinking on the API recommendations, maybe we should remove this field (also in LokiInstanceSpec) and consider:

* if instanceSpec != null --> autoDeploy is enabled
* if instanceSpec == null --> autoDeploy is disabled

api/v1alpha1/flowcollector_types.go
```diff
@@ -32,7 +32,10 @@ const (
 
 func (spec *FlowCollectorSpec) UseEBPF() bool  { return spec.Agent.Type == AgentEBPF }
 func (spec *FlowCollectorSpec) UseIPFIX() bool { return spec.Agent.Type == AgentIPFIX }
-func (spec *FlowCollectorSpec) UseKafka() bool { return spec.DeploymentModel == DeploymentModelKafka }
+func (spec *FlowCollectorSpec) UseKafka() bool {
```
jotak, line 35: I'm not sure about this ... does it mean that the user can set "DeploymentModel" to "DIRECT" and still use Kafka? I don't think it should be allowed

api/v1alpha1/flowcollector_types.go
jotak: Remember that relying on pointers may not work when installing from the OLM form, iirc.
In the previous PoC there was an idea to list explicitly the operators we want to install (I think it was Julien's idea)
It could be a good design choice..

```yaml
  operatorsAutoInstall: ['loki', 'kafka']
  kafka:
    autoInstallSpec:
       ...
  loki:
    autoInstallSpec:
       ...
```

That way we comply with the API guidelines (avoid bool) and we still make it very explicit to install them

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -57,6 +57,12 @@ spec:
     kafkaConsumerQueueCapacity: 1000
     kafkaConsumerBatchSize: 10485760
   kafka:
+    instanceSpec:
+      enable: false
+      replicas: 3
+      zooKeeperReplicas: 3
+      partitions: 10
```
jotak, line 64: previously, @mariomac  recommended something bigger, like 24

api/v1alpha1/flowcollector_types.go
jpinsonneau: If everyone agree we can move to `autoDeploy` :+1: 

^^^ 
@jotak @OlivierCazade

api/v1alpha1/flowcollector_types.go
jpinsonneau: check [DeploymentModel suggestion](https://github.com/netobserv/network-observability-operator/pull/172/files/1c54d69d9016b4426d4d9f2fa04aacb8478c020b#r988744399) above

controllers/operators/embed/kafka_instance.yaml
```diff
@@ -0,0 +1,37 @@
...
+        authentication:
+          type: tls
+    config:
+      offsets.topic.replication.factor: 1
+      transaction.state.log.replication.factor: 1
+      transaction.state.log.min.isr: 1
+      default.replication.factor: 1
+      min.insync.replicas: 1
+      inter.broker.protocol.version: "3.1"
+    storage:
```
jotak, line 29: we need to go with persistent storage (ephemeral is ok for demo, not for production)
Similar to what @memodi did here: https://github.com/netobserv/documents/pull/30/files

By the way, the storage class is something we can try to get dynamically, but we also need a fallback if we can't find a suitable one, which means we need to expose that in the CR

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -71,6 +77,13 @@ spec:
         certFile: user.crt
         certKey: user.key
   loki:
+    instanceSpec:
+      enable: false
+      secretName: loki-secret
+      objectStorageType: s3
+      size: 1x.extra-small
```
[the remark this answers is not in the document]
jpinsonneau: I put extra small for testing.

Do you mean `1x.medium` ? 
Available values are `Enum="1x.extra-small";"1x.small";"1x.medium"`

api/v1alpha1/flowcollector_types.go
jpinsonneau: sure, let's try this approach for first version

api/v1alpha1/flowcollector_types.go
jpinsonneau: moved to `operatorsAutoInstall` array + `autoInstallSpec` sections

api/v1alpha1/flowcollector_types.go
jpinsonneau: thanks @jotak

jpinsonneau:
Loki roles are now automatically set https://github.com/netobserv/network-observability-operator/pull/172/commits/552e3f55b32fac82da0d83bd7dcb5123fcb632b1

jpinsonneau:
Kafka secrets are now automatically sync using watcher https://github.com/netobserv/network-observability-operator/pull/172/commits/55dbebfc81b34dece983868891b29cd3f9b102db

jpinsonneau:
FlowCollector will now get conditions updated with dependent operators updated https://github.com/netobserv/network-observability-operator/pull/172/commits/f7b6b861b058462a760ab36f43b580708dbc6f20
image

This overlap a bit https://issues.redhat.com/browse/NETOBSERV-610. We still need to improve [ClientHelper.GetStatus](https://github.com/netobserv/network-observability-operator/pull/172/files#diff-e46a46b91172361d386fb98d0fc9dad946821c6af258e7df000fb594341868d8R182) to be more precise about pods states after creation

mariomac (commented):
Kafka auto-deployment apparently worked properly but I haven't been able to deploy Loki. I tried it twice:

## First attempt

1. Manually created the `netobserv` namespace
2. Manually created the `loki-secret`
3. Deployed the operator --> Loki secret was removed and I didn't notice it until later
4. Deployed the flowcollector --> status is `MissingLokiSecret`
5. Manually recreated the loki-secret
6. FlowCollector status is still MissingLokiSecret
7. Redeployed FlowCollector
8. FlowCollector Status is always "Unknown"

Checking manager logs, it continuously says: 

```
NFO	controller.flowcollector	custom resource definition: 'lokistacks.loki.grafana.com' doesn't exist yet. waiting for it	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "component": "OperatorsController", "function": "manageOperator"}
```

The FLP and plugin pods can't neither start. Checking their status:

```
  Normal   Scheduled    5m18s                 default-scheduler  Successfully assigned netobserv/flowlogs-pipeline-transformer-78c5785cc9-cvwk4 to ip-10-0-137-45.ec2.internal by ip-10-0-163-59
  Warning  FailedMount  3m16s                 kubelet            Unable to attach or mount volumes: unmounted volumes=[loki-certs-ca], unattached volumes=[config-volume kafka-certs-ca kafka-certs-user loki-certs-ca loki-token kube-api-access-fktfl]: timed out waiting for the condition
  Warning  FailedMount  73s                   kubelet            Unable to attach or mount volumes: unmounted volumes=[loki-certs-ca], unattached volumes=[kafka-certs-user loki-certs-ca loki-token kube-api-access-fktfl config-volume kafka-certs-ca]: timed out waiting for the condition
  Warning  FailedMount  69s (x10 over 5m19s)  kubelet            MountVolume.SetUp failed for volume "loki-certs-ca" : configmap "lokistack-ca-bundle" not found
```

## Second attempt
I had to undeploy the whole operator and redeploying it operator in the "correct" steps:

1. Deploy the operator
2. Create the loki-secret
3. Deploy the flowcollector --> Now the status is "Ready"

However, the FLP pods are in "containerCreating status" after 10 minutes:

```
  Warning  FailedMount  4m27s (x6 over 4m42s)  kubelet            MountVolume.SetUp failed for volume "kafka-certs-ca" : secret "kafka-cluster-cluster-ca-cert" not found
  Warning  FailedMount  3m38s (x8 over 4m42s)  kubelet            MountVolume.SetUp failed for volume "kafka-certs-user" : secret "flp-kafka" not found
  Warning  FailedMount  2m39s                  kubelet            Unable to attach or mount volumes: unmounted volumes=[kafka-certs-user loki-certs-ca], unattached volumes=[kube-api-access-ltrfw config-volume kafka-certs-ca kafka-certs-user loki-certs-ca loki-token]: timed out waiting for the condition
  Warning  FailedMount  2m34s (x9 over 4m42s)  kubelet            MountVolume.SetUp failed for volume "loki-certs-ca" : configmap "lokistack-ca-bundle" not found
  Warning  FailedMount  35s                    kubelet            Unable to attach or mount volumes: unmounted volumes=[loki-certs-ca], unattached volumes=[kube-api-access-ltrfw config-volume kafka-certs-ca kafka-certs-user loki-certs-ca loki-token]: timed out waiting for the condition
```

The manager logs still show the `custom resource definition: 'lokistacks.loki.grafana.com' doesn't exist yet. waiting for it` message.

## Actions

I think we should fix the Loki creation (or at least document if I'm missing some step or doing something wrong). And after merging the feature, we should file some bugs/improvements that can be handled after the feature freeze:

* While Kafka and Loki are being created, the FlowCollector status is `Ready` while it is not. It should show status like `CreatingLoki` or `CreatingKafka`.

* We need to see if there is any standard way to not wipe the `netobserv` namespace if we deploy the operator.

* If status is `MissingLokiSecret` and the Loki secret is created, it should detect it.

jpinsonneau:
- Ready will no longuer be "true" while `LokiStackPending` or `KafkaNotReady` are.
- You can now specify `source` and `startingCSV` in CRD
- Improved conditions
https://github.com/netobserv/network-observability-operator/pull/172/commits/0517e2cd8974b2a4b5d1ee7e21a6b87f3aa4f685

controllers/flowcollector_controller.go
```diff
@@ -352,15 +450,17 @@ func (r *FlowCollectorReconciler) newClientHelper(desired *flowsv1alpha1.FlowCol
...
-	meta.SetStatusCondition(&fc.Status.Conditions, *cond)
+	r.updateConditions(ctx, &[]metav1.Condition{*cond}, fc)
+}
+
+func (r *FlowCollectorReconciler) updateConditions(ctx context.Context, conds *[]metav1.Condition, fc *flowsv1alpha1.FlowCollector) {
+	conditions.SetNewConditions(&fc.Status.Conditions, conds)
+
 	if err := r.Status().Update(ctx, fc); err != nil {
 		log.FromContext(ctx).Error(err, "Set conditions failed")
 		// Do not propagate this update failure: if update failed it's generally because it was modified concurrently:
```
jpinsonneau, line 466: @jotak have you tried to use `patch` here instead ? 

Since I have implemented the watchers, I get more concurrent modifications than before and it miss some conditions updates.
The worst situation would be missing `namespace` update here since it's stored in the `Status` and we update everything once.

OlivierCazade:
I am still testing this PR, but here is a first feedback:
- our upstream release should point to the upstream release of the other operator, my understanding is we are not pointing the downstream ones

I first had some issue because I missed the comment about manually adding the catalog source which resulted in operator beeing in a wrong state. After cleaning everything, both loki and kafka operator were installed successfully and the flowcollecting stack was working normally.

controllers/flowcollector_controller.go
```diff
@@ -67,6 +74,10 @@ func NewFlowCollectorReconciler(client client.Client, scheme *runtime.Scheme) *F
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,resourceNames=hostnetwork,verbs=use
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;create;update;watch
 //+kubebuilder:rbac:groups=apiregistration.k8s.io,resources=apiservices,verbs=list;get;watch
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=operatorgroups;subscriptions,verbs=create;delete;patch;update;get;watch;list
+//+kubebuilder:rbac:groups=kafka.strimzi.io,resources=kafkas;kafkatopics;kafkausers,verbs=create;delete;patch;update;get;watch;list
+//+kubebuilder:rbac:groups=loki.grafana.com,resources=lokistacks;network,verbs=create;delete;patch;update;get;watch;list
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=list;get;watch
```
OlivierCazade, line 81: More a general comment here, but we should try commenting the reason for each permission.

controllers/flowcollector_controller.go
```diff
@@ -101,6 +113,40 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	clientHelper := r.newClientHelper(desired)
+
+	// reconcile operator content only when CRD / Owned object has changed
+	// this avoid watched objects like secret to call unecessary reconcile steps
```
OlivierCazade, line 118: May be we should start to have multiple goroutine.
Each one watching and reconciling independent part.
So we would avoid this kind of check.

controllers/reconcilers/client_helper.go
```diff
@@ -138,3 +146,40 @@ func (c *ClientHelper) ReconcileClusterRole(ctx context.Context, desired *rbacv1
...
+func (c *ClientHelper) Apply(ctx context.Context, u *unstructured.Unstructured) error {
+	return c.ApplyWithNamespaceOverride(ctx, u, "")
+}
+
+// Apply a given manifest with an optional namespace to override
+func (c *ClientHelper) ApplyWithNamespaceOverride(ctx context.Context, u *unstructured.Unstructured, namespaceOverride string) error {
+	log := log.FromContext(ctx, "component", "ClientHelper", "function", "ApplyWithNamespaceOverride")
+
+	if namespaceOverride != "" {
+		u.SetNamespace(namespaceOverride)
```
OlivierCazade, line 160: This modify the original object right?
I think we should avoid to do that.

pkg/helper/maps.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"reflect"
+)
+
+var (
+	MaxDepth = 16
+)
+
+func Merge(dst, src map[string]interface{}) map[string]interface{} {
```
OlivierCazade, line 11: We may add a todo here once we move to go 1.18
There is a new maps copy function that would do the job:
https://pkg.go.dev/golang.org/x/exp/maps#Copy

OlivierCazade (commented):
I did not finish to review yet, I will continue later and I need to do more testing (I did not try to update instanceSpecs after the first deployment for example)

pkg/helper/maps.go
mariomac: Being picky, you don't need to return `dst` as maps are treated as references.

controllers/flowcollector_controller.go
jpinsonneau: Sure I can put some effort on this. Also I guess we can reduce some permissions :+1:

OlivierCazade:
I tested again this morning, and while first deployment worked well, then I tried to make some change in the deployment config.

I changed the kafka replicas values from 3 to 1 but nothing happened, there were no errors in the operator logs and the 3 replicas were still there for both kafka and zookeeper.

OlivierCazade:
> According to the previous discussions, we decided to not update / delete existing content (subscriptions / instances etc) so for now this is the expected behavior. It create content once and then never affect it.

Ok, I understand the choice and the reasons behind but we need to document this because this is not the usual way and may be not intuitive for the users.

controllers/flowcollector_controller.go
jpinsonneau: This would help to split the code and avoid checking `req.Name` all the time, however, how can we manage status & conditions with this approach ?

Let's say we have two controllers, `flowcollector_controller` (for our components) & `dependencies_controller` (for secrets & dependent operators). 
If a secret or a dependent operator is pending on `dependencies_controller,` we should not mark status as `Ready` from `flowcollector_controller`.

Does this mean we should store some states from each controller in `FlowCollectorStatus` as same as Namespace and rely on these to set the final conditions ?

jpinsonneau:
PR has been rebased. No more changes on my side.

`loki operator 5.6` is still not yet available as far as I see in my cluster bot however @memodi had it in a CI cluster :thinking:

mariomac:
@jpinsonneau I tried to install it but I had some problems with the PersistentVolume claims:

```
  Warning  FailedScheduling  3m34s  default-scheduler  0/6 nodes are available: 6 pod has unbound immediate PersistentVolumeClaims. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling.
```

That's ok, I just changed the size of the storage and reapplied the Flowcollector (also decreasing the instances for Kafka and Zookeper) but the changes didn't take effect.

Then I tried to deleting the FlowCollector but the reconciliation of kafka was still blocked:

```
NAME                                                    READY   STATUS    RESTARTS   AGE
amq-streams-cluster-operator-v2.2.0-2-bfdb5f487-7z9xb   1/1     Running   0          6m50s
kafka-cluster-zookeeper-0                               0/1     Pending   0          6m33s
kafka-cluster-zookeeper-1                               0/1     Pending   0          6m33s
kafka-cluster-zookeeper-2                               0/1     Pending   0          6m33s
netobserv-controller-manager-746d858486-rwbc8           2/2     Running   0          14m
```

I had to completely undeploy the operator to be able to remove the pods.

mariomac:
@jpinsonneau I redeployed with smaller persistent volume sizes and the status of the flowcollector is `ReconcileDependentOperatorsFailed`. The manager logs continuously this message:

```
1.6680775737424061e+09	ERROR	controller.flowcollector	Failed to reconcile dependent operators: LokiStack.loki.grafana.com "lokistack" is invalid: spec.tenants.mode: Unsupported value: "openshift-network": supported values: "static", "dynamic", "openshift-logging"	{"reconciler group": "flows.netobserv.io", "reconciler kind": "FlowCollector", "name": "cluster", "namespace": "", "error": "LokiStack.loki.grafana.com \"lokistack\" is invalid: spec.tenants.mode: Unsupported value: \"openshift-network\": supported values: \"static\", \"dynamic\", \"openshift-logging\""}
```

controllers/flowcollector_controller.go
jotak: no I haven't tried patch, I guess it's worth investigating

controllers/secrets/secrets_reconciler.go
```diff
@@ -0,0 +1,110 @@
...
+			if err := c.reconcileSecret(ctx, kafkaTLS.CACert.Name); err != nil {
+				return fmt.Errorf("reconciling kafka ca secret: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func (c *Reconciler) getSecrets(ctx context.Context, name string) (*v1.Secret, *v1.Secret, error) {
+	rlog := log.FromContext(ctx, "component", "deployLoki")
```
jotak, line 61: Why "deployLoki" here? mistake?

jotak:
I've finished reviewing the code. I have some remarks but I think they can be addressed in a follow-up (we should probably try to go ahead and iterate from there, in order to have it more tested before GA, at least for non-regressions. I just hope this feature will not bring too much maintenance cost for the benefit, because it is quite complex and depends on external factors).

I haven't tested, except some non-regression tests. I leave it to people who test to put the lgtm mark :)

What I'd like to see in a follow-up:
- Having kubebuilder validation on AutoInstall enum (I think we can do that by declaring a new string type, like `type AutoInstall string`, with the appropriate kubebuilder annotation on it)
- Use upper case for new enum, consistently with other enums
- Nitpicking, but wouldn't it be nice to use a similar API that we did on Spec (like `spec.UseKafka()`) for these auto-install things?
- Could we have the secrets watcher being run as a separate controller? Maybe also the CRD watcher? In the same idea than what Olivier suggested with different goroutines, I guess...
- If I read correctly, Kafka secrets, when present, are always copied to privileged namespace. But there's cases where we don't have such a namespace, e.g. when agent is ipfix, and I wonder if that's also possible with ebpf (I don't remember, I guess @mariomac can answer). In that case we should not try to copy secrets.

controllers/secrets/secrets_reconciler.go
jpinsonneau: yes it is ! thanks

controllers/operators/embed/operator_group.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: netobserv-dependend-operators
```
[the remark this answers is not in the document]
jpinsonneau: It's a [mandatory resource](https://github.com/netobserv/network-observability-operator/pull/172/files#diff-30add0122a3cbfd6e0941c85cb0ca2c39b2f5af1a21682ac75504044aa1d4a47R93-R94) generating RBACs
https://olm.operatorframework.io/docs/tasks/install-operator-with-olm/#prerequisites
https://docs.openshift.com/container-platform/4.8/operators/understanding/olm/olm-understanding-operatorgroups.html

You can't create a subscription in a particular namespace without this

controllers/flowcollector_controller_test.go
```diff
@@ -630,7 +630,7 @@ func GetReadyCR(key types.NamespacedName) *flowsv1alpha1.FlowCollector {
 			return err
 		}
 		cond := meta.FindStatusCondition(cr.Status.Conditions, conditions.TypeReady)
-		if cond.Status == metav1.ConditionFalse {
+		if cond != nil && cond.Status == metav1.ConditionFalse {
```
jotak, line 633: @jpinsonneau maybe the problem is here: if cond is nil, then it should return an error ?

jpinsonneau: thanks it seems to work as expected :+1:

PR #173, opened 2022-09-27
NETOBSERV-579 Use netobserv prefix, add 4 new metrics, remove one

jotak:
New metrics: source/dest bytes traffic per node / per owner
Removed `bandwidth_per_network_service_per_namespace` (per namespace traffic can be obtained from the new metrics)
That also removed the per-service discrimination, but I'm not sure we actually need it. In case we do, we should introduce a new metric.

cc @memodi @Amoghrd 

Detailed changes:

The 4 following metrics show traffic per source/dest workload/namespace, or node for those with "_node_".
- Added **netobserv_node_received_bytes_total**
- Added **netobserv_node_sent_bytes_total**
- Added **netobserv_received_bytes_total**
- Added **netobserv_sent_bytes_total**

They allow to more accurately follow traffic occuring in a given namespace, e.g.

`sum(rate(netobserv_received_bytes_total{SrcK8S_Namespace="netobserv"}[1m]))`
and
`sum(rate(netobserv_received_bytes_total{DstK8S_Namespace="netobserv",SrcK8S_Namespace!="netobserv"}[1m]))`

will cover all traffic occuring inside netobserv and from/to other namespaces/nodes

The "send" and "received" words here account for traffic going in or out from the observation point of view, it's basically the same thing as the "Reporter" field in the UI (or FlowDirection field) ; so they don't add up (because that would result in duplicated traffic).

Other changes are:
- Removed **netobserv_bandwidth_per_network_service_per_namespace**
- Added **netobserv_** prefix to all FLP internal metrics
- Changed some FLP internal metrics, cf [details here](https://github.com/netobserv/flowlogs-pipeline/pull/314#issuecomment-1256440714)
- Added **netobserv_** prefix to loki client metrics (instead of "promtail" previously), so that they are more easily identifiable


memodi (commented):
I tested this with PR image, looks like below metrics doesn't have "netobserv" prefix in them which are for protobuf/protobuf1.0, are these still relevant? if so, should they have "netobserv" prefix in them.

```
flow_decoder_count{name="protobuf",worker=""} 442
flow_process_nf_delay_summary_seconds{router="",version="protobuf1.0",quantile="0.5"} 1.741961812
flow_process_nf_delay_summary_seconds{router="",version="protobuf1.0",quantile="0.9"} 4.353887055
flow_process_nf_delay_summary_seconds{router="",version="protobuf1.0",quantile="0.99"} 4.967831966
flow_process_nf_delay_summary_seconds_sum{router="",version="protobuf1.0"} 379294.89681290183
flow_process_nf_delay_summary_seconds_count{router="",version="protobuf1.0"} 185834
flow_traffic_summary_size_bytes{local_ip="0.0.0.0",local_port="2055",remote_ip="10.0.0.6",type="protobuf",quantile="0.5"} 37739
flow_traffic_summary_size_bytes{local_ip="0.0.0.0",local_port="2055",remote_ip="10.0.0.6",type="protobuf",quantile="0.9"} 42017
flow_traffic_summary_size_bytes{local_ip="0.0.0.0",local_port="2055",remote_ip="10.0.0.6",type="protobuf",quantile="0.99"} 44748
flow_traffic_summary_size_bytes_sum{local_ip="0.0.0.0",local_port="2055",remote_ip="10.0.0.6",type="protobuf"} 1.6958135e+07
flow_traffic_summary_size_bytes_count{local_ip="0.0.0.0",local_port="2055",remote_ip="10.0.0.6",type="protobuf"} 442
```

jotak:
@memodi good point, these are metrics that we reuse from goflow2 (they are used natively by goflow2 in the context of IPFIX agent, and we mimic them in the context of the eBPF agent).
Since now we are focusing just on the ebpf agent, I think we can forget about keeping them identical to the goflow's ones, right? So we can be more consistent with the rest of our metrics. I'll update the PR

jotak:
actually, they reside in FLP, I need to open a new PR for FLP, but there's nothing more to do on the operator (so this PR can still be merged)

OlivierCazade (approved):

PR #175, opened 2022-10-03
NETOBSERV-489: Use authorize feature from console plugin API

OlivierCazade:
This PR enable the new option from the console from this PR:

https://github.com/netobserv/network-observability-console-plugin/pull/198

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -416,6 +416,15 @@ type FlowCollectorLoki struct {
 	TLS ClientTLS `json:"tls"`
 }
 
+const (
+	IgnoreUserToken  = "IGNORE"
+	ForwardUserToken = "FORWARD"
+)
+
+func (spec *FlowCollectorConsolePlugin) ForwardUserToken() bool {
+	return spec.UserToken == ForwardUserToken
+}
+
 // FlowCollectorConsolePlugin defines the desired ConsolePlugin state of FlowCollector
 type FlowCollectorConsolePlugin struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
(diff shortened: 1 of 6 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -446,6 +446,11 @@ type FlowCollectorConsolePlugin struct {
 	// imagePullPolicy is the Kubernetes pull policy for the image defined above
 	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
 
+	//+kubebuilder:default:=false
+	// sendAuthToken is a flag to enable or disable forwarind auth token
+	// if enabled, this overide loki.sendAuthToken
+	ForwardUserAuthToken bool `json:"forwardUserAuthToken,omitempty"`
```
mariomac, line 452: Following the API recommendations, discouraging the use of bools, maybe I'd move this to a string property. Something like: `userAuthToken: {BLOCK | FORWARD}`

jpinsonneau: maybe we should merge with `FlowCollectorLoki.SendAuthToken` to a single enum then ? :thinking:

OlivierCazade: I change the value to an enum.

OlivierCazade: > maybe we should merge with `FlowCollectorLoki.SendAuthToken` to a single enum then ? thinking

`FlowCollectorLoki.SendAuthToken` apply to both FLP and Console plugin while this new value only apply to the Console Plugin.

Also `FlowCollectorLoki.SendAuthToken` is still required for FLP when the user token forwarding is enabled.

jpinsonneau: Yes. My thinking was about having `FlowCollectorLoki.SendAuthToken` as:
```yaml
enum:
  - IGNORE
  - SERVICE_ACCOUNTS
  - MIXED
  type: string
```

With proper documentation to explain `MIXED` behavior

OlivierCazade: Done

mariomac (approved):
LGTM. I haven't tested it.

PR #177, opened 2022-10-06
NETOBSERV-578: Disable FLP prometheus if "metricsServer" property is not set

mariomac:
CPU is sensibly lower:

image

Memory is also sensibly lower:

image


api/v1alpha1/flowcollector_types.go
```diff
@@ -284,7 +284,7 @@ type FlowCollectorFLP struct {
 
 	// metricsServer endpoint configuration for Prometheus scraper
 	// +optional
-	MetricsServer MetricsServerConfig `json:"metricsServer,omitempty"`
+	MetricsServer *MetricsServerConfig `json:"metricsServer,omitempty"`
```
jotak, line 287: Remember we can't rely on nil pointers for decisions due to the OLM console form.
What about a list enum that would also include the "enableKubeProbes" bool?

Something like:

```yaml
  monitoring: ['kube-probes', 'metrics']  
```

(though there's probably a better name than "monitoring" ..  )

config/manager/kustomization.yaml
```diff
@@ -12,7 +12,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/netobserv/network-observability-operator
-  newTag: 0.1.4
+  newName: quay.io/mmaciasl/network-observability-operator
```
jotak, line 15: to revert :)

jotak:
I don't see such a gain on my side, when disabling metrics. I see something more of the magnitude of 5-10MB decrease on a total of 160MB, so much less significant, I don't know why ...
Note: I was first trying to see if just removing the metrics with the higher cardinality would help enough; we can do that already with the current CRD, just set:
```
processor:
  ignoreMetrics: ['rate']
```
It does have a small impact, like 5MB gain on 160MB total

mariomac:
@jotak thanks for the review. Actually this PR is WIP and I forgot to label it as is, sorry 🥲

I agree that improvement doesn't seem so high at the moment so I'll put this on-hold as a WIP research.

mariomac:
I see a higher gain probably because the load is higher. I usually deploy this file to generate high number of flows: https://github.com/mariomac/flplite/blob/main/doc/perftest/loadgen.yaml

I wonder if in a scenario requiring many CPUs, several GB per FLP, the difference is bigger.

PR #178, opened 2022-10-06
Fix auth token mode for FLP

OlivierCazade:
My changes after code review introduced a bug, FLP host token needs to be configured in both modes.


jotak (approved):
/lgtm

mariomac (approved):

PR #179, opened 2022-10-10
NETOBSERV-466: Add metric to count packet number in FLP

OlivierCazade:
This add a metric to count packet number.

Story was to add only total packet count but I aligned this with the bytes count, and we can still sum different tag to get the total number of packets.

### DIFF
--- a/controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
@@ -9,7 +9,6 @@ tags:
   - egress
   - bytes
   - workloads
-  - namespaces
 encode:
   type: prom
   prom:
--- a/controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
@@ -1,12 +1,12 @@
 #flp_confgen
 description:
-  This metric observes the incoming traffic
+  This metric observes the outgoing traffic
 details:
-  Sum bytes for incoming traffic per source and destination namespaces and owners
+  Sum bytes for outgoing traffic per source and destination namespaces and owners
 usage:
-  Evaluate network incoming usage breakdown per source and destination namespaces and owners
+  Evaluate network outgoing usage breakdown per source and destination namespaces and owners
 tags:
-  - ingress
+  - egress
   - bytes
   - workloads
   - namespaces
@@ -14,12 +14,12 @@ encode:
   type: prom
   prom:
     metrics:
-      - name: received_bytes_total
+      - name: workload_egress_bytes_total
         type: counter
         valuekey: Bytes
         filter:
           key: FlowDirection
-          value: "0"
+          value: "1"
         labels:
           - SrcK8S_Namespace
           - DstK8S_Namespace
--- a/controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
@@ -18,7 +18,7 @@ encode:
         valuekey: Packets
         filter:
           key: FlowDirection
-          value: "0"
+          value: "1"
         labels:
           - SrcK8S_Namespace
           - DstK8S_Namespace
--- a/controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
@@ -1,12 +1,12 @@
 #flp_confgen
 description:
-  This metric observes the incoming traffic
+  This metric observes the outgoing traffic
 details:
-  Sum packet number for incoming traffic per source and destination namespaces and owners
+  Sum packet number for outgoing traffic per source and destination namespaces and owners
 usage:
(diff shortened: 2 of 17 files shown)


memodi:
@OlivierCazade  Also, any reason we're not adding metric for "FlowDirection: 1" ? 

cc @jotak -   I believe this metric would also scale similarly as sent/received_bytes_total

jotak:
@memodi yes ... I'd suggest some improvements in our handling of these metrics. I will create a new JIRA for that.

Basically what I'd like to do is to have some default config to not expose all these metrics. The mechanism to ignore metrics already exist, it works with the `tags` defined [here](https://github.com/netobserv/network-observability-operator/pull/179/files#diff-8d1dc7c5c07a7d81c569256b50addac6916775e82020d97d96bd3c30ee17e22fR8-R12) and the `ignoreMetrics` config [there](https://github.com/netobserv/network-observability-operator/blob/main/config/samples/flows_v1alpha1_flowcollector.yaml#L47).

Basically I think we could only expose `received_bytes_total` by default (btw: I don't like its name ... "received" is confusing, maybe we should use the terms ingress/egress instead) ; other metrics would have to be enabled explicitely.
So we could have this set of tags to control what we want:
- ingress
- egress
- bytes
- packets
- workloads
- namespaces

Also I'd like to introduce new similar metrics that have only namespace labels, not workloads, also disabled by default. So that can be used as a mitigation solution if users see too high cardinality with metrics, putting too much pressure on prometheus & FLP, without entirely turning off telemetry: they could turn off per-workload metrics and turn on per-namespace metrics instead.

As said, I'll create a new JIRA

jotak:
> Also, any reason we're not adding metric for "FlowDirection: 1" ?

+1 I think we should provide that as well (it can catch some traffic not caught on ingress), however, as said above, we may want to disable it through config by default.

OlivierCazade:
Thanks for feedback!

>  Also, any reason we're not adding metric for "FlowDirection: 1" ?

The reason for not catching `FLowDirection: 1` was to be consistent with the `received_bytes_total` metric

> +1 I think we should provide that as well (it can catch some traffic not caught on ingress)

This is true, this mean this metrics are reporting wrong value, should we remove both until we find a way to clean duplicate? May be we have to wait for connection tracking

Would you be fine with splitting these metrics into four ?
Something like  `captured_ingress_bytes_total` `captured_egress_bytes_total` `captured_ingress_packets_total` and `captured_egress_packets_total` (and may be enabling by default only (captured_ingress_bytes_total`).

OlivierCazade:
I modified current metrics tag and default ignored metrics.

This PR also address part of this task :
https://github.com/netobserv/network-observability-operator/pull/179

The only remaining part is about having "light" metrics with less cardinality.

@memodi , the last part is renaiming the metrics. Is it fine or would it create a lot of work on your side?

memodi:
> @memodi , the last part is renaiming the metrics. Is it fine or would it create a lot of work on your side?
if it's absolutely desired, go ahead renaming them. thanks!

controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
```diff
@@ -6,16 +6,15 @@ details:
...
 tags:
-  - bandwidth
-  - rate
-  - namespace
-  - owner
-  - outgoing
+  - egress
+  - bytes
+  - workloads
+  - namespaces
```
jotak, line 12: I wouldn't set "namespaces" here, else if we later add namespace-only metrics later, we wouldn't be able to filter them out without impacting this one as well (or we'd need to find a trick to do it)

controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
```diff
@@ -0,0 +1,29 @@
...
+  This metric observes the outgoing traffic
+details:
+  Sum packet number for outgoing traffic per source and destination namespaces and owners
+usage:
+  Evaluate network incoming usage breakdown per source and destination namespaces and owners
+tags:
+  - egress
+  - packets
+  - workloads
+  - namespaces
```
jotak, line 12: same here and same for the other workload metrics

controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
jotak: well, depends how we do the namespace-only metrics ... you plan to add label-removal feature like we discussed previously, right? If so, then it's ok for me

controllers/flowlogspipeline/metrics_definitions/workload_egress_packets_total.yaml
```diff
@@ -0,0 +1,29 @@
...
+encode:
+  type: prom
+  prom:
+    metrics:
+      - name: workload_egress_packets_total
+        type: counter
+        valuekey: Packets
+        filter:
+          key: FlowDirection
+          value: "0"
```
jotak, line 22: should be "1" for egress

controllers/flowlogspipeline/metrics_definitions/workload_egress_bytes_total.yaml
OlivierCazade: I plan to do the label-removal, but I am not sure that this is the priority.

I removed the namespaces label and I will add it again once the label removal task is done.

jotak (approved):

PR #180, opened 2022-10-10
NETOBSERV-578: allow passing env vars to FLP

mariomac:
Forwarding environment variables to FLP will allow tuning some performance parameters from e.g. the Go runtime.

Here is an example of trading with CPU/Memory by setting `GOGC=400` vs `GOGC=200` vs `GOGC=100` (default value, unset):

image

image


jpinsonneau (approved):
:+1:

api/v1alpha1/flowcollector_types.go
```diff
@@ -350,6 +350,13 @@ type FlowCollectorFLP struct {
 	// +optional
 	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
 	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
+
+	// env allows passing custom environment variables to the Flowlogs-Pipeline pod.
+	// This field is useful for passing some concrete performance-tuning options
+	// (e.g. GOGC, GOMAXPROCS) that shouldn't be
+	// publicly exposed as part of the FlowCollector descriptor.
+	//+optional
+	Env map[string]string `json:"env,omitempty"`
```
jotak, line 359: As long as it's there, it's publicly exposed. I'm not 100% sure what we can do it ; e.g. in terms of support, if people start to put any env variable, how does it work? We should consider that first.

PR #181, opened 2022-10-14
NETOBSERV-660: Create metric section

OlivierCazade:
As an anticipation to more work regarding metrics in FLP and before going GA, create sub metric section in processor section.


mariomac (approved):

jotak:
/hold
let see with @memodi if we want it in 0.1.5

memodi:
> /hold let see with @memodi if we want it in 0.1.5

thanks @jotak, seems straightforward change, should be fine adding it in if its already reviewed and good to be approved.

jotak (approved):

memodi:
/qe-approved
tested pre-merge.

PR #182, opened 2022-10-17
NETOBSERV-447: Changed default FLP retry value to prevent FLP beeing oom-killed

OlivierCazade:


jpinsonneau (commented):
Does this means we would have only 3 queries in total during a 10s frame before dropping flows ?

If loki is down, should we get the `/ready` enpoint result before retrying ?

jotak (approved):
lgtm

PR #183, opened 2022-10-24
NETOBSERV-645 Kafka export (flagged as experimental feature)

jotak:
Allow defining a Kafka exporter (or several of them). FLP will then send flows there as JSON. It is an additional stage: all existing stages are unchanged (kafka for internal processing, writing to loki, etc.)

No required change in FLP, which already has everything needed.

Documentation PR: https://github.com/netobserv/documents/pull/33


jpinsonneau (approved):
LGTM on code side (not tested)

PR #184, opened 2022-10-24
NETOBSERV-382: Added status field in HPA configuration

OlivierCazade:
This add a status field in HPA section to deploy or not HPA.

At some point in the future we may want a third option: not deploying the HPA but also not reconciling pod count so an external HPA can be used.

### DIFF
--- a/api/v1alpha1/flowcollector_types.go
+++ b/api/v1alpha1/flowcollector_types.go
@@ -346,7 +346,7 @@ type FlowCollectorFLP struct {
 	// kafkaConsumerAutoscaler spec of a horizontal pod autoscaler to set up for flowlogs-pipeline-transformer, which consumes Kafka messages.
 	// This setting is ignored when Kafka is disabled.
 	// +optional
-	KafkaConsumerAutoscaler *FlowCollectorHPA `json:"kafkaConsumerAutoscaler,omitempty"`
+	KafkaConsumerAutoscaler FlowCollectorHPA `json:"kafkaConsumerAutoscaler,omitempty"`
 
 	//+kubebuilder:default:=1000
 	// +optional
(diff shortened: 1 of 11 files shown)


api/v1alpha1/flowcollector_types.go
```diff
@@ -366,7 +366,19 @@ type FlowCollectorFLP struct {
...
 
+const (
+	HPAStatusDisabled = "DISABLED"
+	HPAStatusEnabled  = "ENABLED"
+)
+
 type FlowCollectorHPA struct {
+	// +kubebuilder:validation:Enum:=DISABLED;ENABLED
+	// +kubebuilder:default:=DISABLED
+	// AuthToken describe the way to get a token to authenticate to Loki
```
jotak, line 377: wrong description, I bet you copy-pasted :D

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -94,7 +94,6 @@ spec:
     imagePullPolicy: IfNotPresent
     port: 9001
     logLevel: info
-    autoscaler: null
```
jotak, line 97: could you add a sample autoscaler config, but commented out? I find it quite convenient when trying configs that are not obvious

api/v1alpha1/flowcollector_types.go
OlivierCazade: Yes sorry, fixed.

config/samples/flows_v1alpha1_flowcollector.yaml
OlivierCazade: Done!

config/samples/flows_v1alpha1_flowcollector.yaml
```diff
@@ -94,7 +94,17 @@ spec:
...
+      status: DISABLED
+      minReplicas: 1
+      maxReplicas: 3
+      metrics:
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 50
```
jotak, line 107: sorry I meant by "commented out" => as comments :

```suggestion
#    autoscaler:
#      status: DISABLED
#      minReplicas: 1
#      maxReplicas: 3
#      metrics:
#      - type: Resource
#        resource:
#          name: cpu
#          target:
#            type: Utilization
#            averageUtilization: 50
```

config/samples/flows_v1alpha1_flowcollector.yaml
jotak: thanks! Actually better than commented out :)

PR #185, opened 2022-10-25
added metrics service to operator

KalmanMeth:
This PR adds the creation of the flowlogs-pipeline-metrics service to the operator.
The next step is to add the ServiceMonitor to expose these metrics to the Openshift console.


jotak:
I think there was a misunderstanding, the metrics service already exists (called prom service in code), or does this one do anything different?

PR #187, opened 2022-10-27
NETOBSERV-480 quick filters

jotak:
Requires console plugin PR: https://github.com/netobserv/network-observability-console-plugin/pull/216

Adds new quick filters configuration to CRD
Set default filters:
- Applications ( = everything not in openshift- or netobserv)
- OpenShift infra ( = openshift and netobserv)
- Exclude services ( kind != service )

Note, I find the latter super useful to easily deduplicate flows to pods vs flows to services.


jotak:
@jpinsonneau about showing not traffic or not, as per your comment https://github.com/netobserv/network-observability-console-plugin/pull/216#pullrequestreview-1156179607

It' tricky to have a one-fits-all approach because you may or may not be interested in seeing node-to-app traffic.

But what we could do is, instead of the "Exclude Services" filter that I added, we could have three default filters:
- Pods network (src/dst kind == pod)
- Services network (src/dst kind == service)
- Nodes network (src/dst kind == node)

That way you can compose via quick filters easily.
We could have by default "Applications" and "Pods network" ?

OlivierCazade (approved):

jpinsonneau (approved):
LGTM

PR #188, opened 2022-10-27
NETOBSERV-565 added service monitors for flp prometheus and console plugin

KalmanMeth:
To view the metrics in the console, apply the following:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: cluster-monitoring-config
  namespace: openshift-monitoring
data:
  config.yaml: |
    enableUserWorkload: true
```


controllers/flowlogspipeline/flp_service_monitor.go
```diff
@@ -0,0 +1,48 @@
...
+	serviceMonitorObject := buildPrometheusServiceMonitorObject(b)
+	// apply object to kubernetes
+	if err := cl.CreateOwned(ctx, serviceMonitorObject); err != nil {
+		return err
+	}
+	return nil
+}
+
+func buildPrometheusServiceMonitorObject(b *builder) *monitoringv1.ServiceMonitor {
+	flpServiceMonitorObject := monitoringv1.ServiceMonitor{
```
jpinsonneau, line 22: since this rely on `ServiceMonitor` custom resource, we should either ensure `servicemonitors.monitoring.coreos.com` CustomResourceDefinition exists or let the user configure this from our CRD.

Else it will crash as you can see in the [CI running on kind](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/netobserv_network-observability-operator/188/pull-ci-netobserv-network-observability-operator-main-ci/1586679116831133696#1:build-log.txt%3A500):
`"error": "no kind is registered for the type v1.ServiceMonitor in scheme \"k8s.io/client-go/kubernetes/scheme/register.go:74\"`

KalmanMeth: added check for existence of `servicemonitors.monitoring.coreos.com`.

jpinsonneau: thanks @KalmanMeth !

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -84,6 +85,39 @@ func (b *builder) consolePlugin() *osv1alpha1.ConsolePlugin {
...
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					Port:     "main",
+					Interval: "15s",
+					Scheme:   "https",
+					TLSConfig: &monitoringv1.TLSConfig{
+						SafeTLSConfig: monitoringv1.SafeTLSConfig{
+							InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
KalmanMeth: @jpinsonneau I'm not sure what I need to do here. Do I need to duplicate the configMap or just provide `CertFile: "/var/serving-cert"` ?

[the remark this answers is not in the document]
jpinsonneau: You can't use `"/var/serving-cert"` here since it's only mounted in `ConsolePlugin` `Pod`.

`SafeTLSConfig` seems to be able to load `Cert` from secret:
```go
SafeTLSConfig: monitoringv1.SafeTLSConfig{
	Cert: monitoringv1.SecretOrConfigMap{
		Secret: &corev1.SecretKeySelector{
			LocalObjectReference: corev1.LocalObjectReference{
				Name: console-serving-cert
			},
			Key: "tls.crt",
		},
	},
},
```

Have you tried that ?

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -84,6 +85,39 @@ func (b *builder) consolePlugin() *osv1alpha1.ConsolePlugin {
 	}
 }
 
+func (b *builder) consolePluginServiceMonitor() *monitoringv1.ServiceMonitor {
+	return &monitoringv1.ServiceMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.PluginServiceMonitorName,
+			Namespace: constants.DefaultOperatorNamespace,
```
jpinsonneau, line 92:
```suggestion
			Namespace: b.namespace,
```

Namespace should be taken from CRD

controllers/consoleplugin/consoleplugin_objects.go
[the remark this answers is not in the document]
KalmanMeth: done

jotak:
I'm going to test it
thanks for the PR @KalmanMeth

jotak:
@KalmanMeth it doesn't work for me, there is a permission issue when trying to read the servicemonitor CRD to determine if they can be installed.
In fact, we use a discovery mechanism to check when external APIs are available (we already do this to check whether ConsolePlugin API exists, for instance) : look at this file: https://github.com/netobserv/network-observability-operator/blob/main/pkg/discover/apis.go

So to check ServiceMonitor, you can edit this file to add `HasServiceMonitor`, similar to the existing `HasConsole` / `HasCNO`

jotak (changes requested):
There's something more to do: manage namespace changes. It is possible to update the netobserv installation to change its namespace, so what the operator does in that case is to remove all the created resources and recreate them in the other namespace.

You don't have to implement all of this, you just need to declare the servicemonitor resource when the reconcilers are created, see here: https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/consoleplugin_reconciler.go#L32-L56

- you need to add a reference in `ownedObjects`
- you need to call `nobjMngr.AddManagedObject`

once it's done, upon namespace change, the controller will take care of deleting these resources and re-creating them. 

Same thing to do for FLP monolith, FLP ingester and FLP transformer

KalmanMeth:
> @KalmanMeth it doesn't work for me, there is a permission issue when trying to read the servicemonitor CRD to determine if they can be installed. In fact, we use a discovery mechanism to check when external APIs are available (we already do this to check whether ConsolePlugin API exists, for instance) : look at this file: https://github.com/netobserv/network-observability-operator/blob/main/pkg/discover/apis.go
> 
> So to check ServiceMonitor, you can edit this file to add `HasServiceMonitor`, similar to the existing `HasConsole` / `HasCNO`. Doing this way, we don't have to add a new read permission on CRD kind itself.

@jotak I added code to check for `HasServiceMonitor`. I'm not sure what the environment was in which you got a failure, so I could not check it.

jotak:
@KalmanMeth sorry for the delay, I'll take another look today

main.go
```diff
@@ -79,9 +83,9 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.StringVar(&config.EBPFAgentImage, "ebpf-agent-image", "", "The image of the eBPF agent")
-	flag.StringVar(&config.FlowlogsPipelineImage, "flowlogs-pipeline-image", "", "The image of Flowlogs Pipeline")
-	flag.StringVar(&config.ConsolePluginImage, "console-plugin-image", "", "The image of the Console Plugin")
+	flag.StringVar(&config.EBPFAgentImage, "ebpf-agent-image", "quay.io/netobserv/netobserv-ebpf-agent:main", "The image of the eBPF agent")
```
eranra, line 86: @KalmanMeth this will not be needed once we merge https://github.com/netobserv/network-observability-operator/pull/203

controllers/constants/constants.go
```diff
@@ -5,8 +5,10 @@ const (
 	DefaultOperatorNamespace = "netobserv"
 	OperatorName             = "netobserv-operator"
 	FLPName                  = "flowlogs-pipeline"
+	FLPServiceMonitorName    = "flowlogs-pipeline-monitor"
 	FLPPortName              = "flp" // must be <15 chars
 	PluginName               = "netobserv-plugin"
+	PluginServiceMonitorName = "netobserv-console-plugin"
```
eranra, line 11: s/PluginServiceMonitorName/ConsoleServiceMonitorName

controllers/flowlogspipeline/flp_service_monitor.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	serviceMonitorObject := buildPrometheusServiceMonitorObject(b)
+	// apply object to kubernetes
+	if err = cl.CreateOwned(ctx, serviceMonitorObject); err != nil {
+		return err
+	}
+	return nil
+}
+
+func buildPrometheusServiceMonitorObject(b *builder) *monitoringv1.ServiceMonitor {
```
eranra, line 36: This function is very similar to `consolePluginServiceMonitor` --- Can you (1) make sure that at least the names will be similar ---starting with the word build and finishing with Object makes sense to me so s/consolePluginServiceMonitor/buildConsolePluginServiceMonitorObject 

Also, if you can re-use the same code in both functions by taking repeated code into a helper function it will be best.

controllers/constants/constants.go
KalmanMeth: done

KalmanMeth:
> So I still get the same error (plus also an error saying missing permissions to get/create servicemonitors)
> I get this error with the nominal test deployment, which is deploying the operator on a running cluster + the default FlowCollector resource. Basically I just build & push the image, then run `IMG=<that image> make deploy deploy-sample-cr`

@jotak I made updates according to your comments. I hope it is now satisfactory.

KalmanMeth:
performed rebase

jotak:
Hi Kalman,
Thanks for the update, it now deploys correctly without an error. I get the servicemonitors created and FLP metrics feeding prometheus 👍
But I see 2 issues:
1. I don't get the console plugin metrics. It seems to be a certificate issue. Not sure what exactly the problem is.. Setting `InsecureSkipVerify: true` would work, but that's not a good fix.
2. When I deploy with Kafka, we loose the FLP metrics. At first glance it seems because the FLP ServiceMonitor is not correctly adapted for this case (when using Kafka, the FLP deployment model changes from monolithic to publisher/subscriber, that's the reason why we have this distinction of flp-ingester / flp-transformer in code, and we need to adapt the service monitor to point to these workloads.

Let me know if you have an idea for point 1. For point 2, I can try doing a fix and adding it to your PR (I know this part of the code is tricky to apprehend) ; If we can't get that fixed quickly we can go ahead & merge, and address them in follow-ups

jotak (approved):
/lgtm
thanks!

Follow-up JIRA for the console plugin issue: https://issues.redhat.com/browse/NETOBSERV-765

PR #190, opened 2022-11-07
NETOBSERV-690 (spike): move eBPF image pullspec from CRD to CSV

mariomac:
From "[Best Practices](https://docs.engineering.redhat.com/display/CFC/Best_Practices)" document:
> Images to be used for operands MUST be in the relatedImages field and SHOULD be in environment variables to the operator.

As an initial test, I'm moving first only one image to evaluate how will it work with CPaaS. Once we succeed, we will move the rest of images.

### DIFF
--- a/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
+++ b/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
@@ -95,11 +95,6 @@ spec:
                         items:
                           type: string
                         type: array
-                      image:
-                        default: quay.io/netobserv/netobserv-ebpf-agent:v0.2.1
-                        description: image is the NetObserv Agent image (including
-                          domain and tag)
-                        type: string
                       imagePullPolicy:
                         default: IfNotPresent
                         description: imagePullPolicy is the Kubernetes pull policy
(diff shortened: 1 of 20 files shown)


bundle/manifests/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -280,6 +275,7 @@ spec:
                       set up for the plugin Deployment.
                     properties:
                       maxReplicas:
+                        default: 3
```
jpinsonneau, line 278: :thinking: are these changes related to your environment ?

I see FlowCollectorHPA.MaxReplicas defined using:
```go
MaxReplicas int32 `json:"maxReplicas" protobuf:"varint,3,opt,name=maxReplicas"`
```
instead of using annotations like:
```go
//+kubebuilder:validation:Minimum=0
//+kubebuilder:default:=3
```

We should create a followup to harmonize these if it's the case

jpinsonneau (approved):
LGTM in terms of code. Thanks @mariomac !

bundle/manifests/flows.netobserv.io_flowcollectors.yaml
mariomac: No, they aren't. But the generated flowcollectors manifest was outdated so when I regenerated mine, the old changes took effect.

Maybe we should add a test target that verifies that the manifests are always up to date.

PR #191, opened 2022-11-08
NETOBSERV-694: add missing annotations to netobserv-privileged namespace

mariomac:
Openshift 4.12 was missing some SCC-related annotations in the netobserv-privileged namespace and couldn't properly deploy the eBPF agent daemonset.

PR #193, opened 2022-11-09
NETOBSERV-638: avoid infinite update-retrigger loop

mariomac:
This fixes the idempotency of the agent and FLP generation when the `env` sections were defined. It also fixes the comparison of some other resources that were unnecessarily reconciliated continuously.

This PR also reverts  PR #191 ([NETOBSERV-694](https://issues.redhat.com//browse/NETOBSERV-694)), as we realized that the error messages were mostly caused by the bug fixed in this PR.


jpinsonneau (approved):
LGTM, good catch @mariomac !

PR #194, opened 2022-11-11
NETOBSERV-707: cut image version labels > 63 chars

mariomac:
Using downstream images prevents the flowcollector from being deployed as they use very long sha digests for the version.


mariomac:
Due to holidays in the rest of reviewer's country and that it's a minor bugfix that is blocking QE, I'll use my admin powers to merge it without prior review.

PR #198, opened 2022-11-22
Amend image patching instructions

mariomac:
`oc set env` or `kubectl set env` is a much more convenient tool for changing the related images.


OlivierCazade (approved):
Thanks!

PR #203, opened 2022-11-27
fix make run (ocp and local)

eranra:
- Fix the sample by pinning it to a specific version (4.1)
- Adding required parameters to run operator command (from the command line)


jotak:
/lgtm
thanks!

PR #204, opened 2022-11-28
CSV: replace placeholder by netobserv namespace name

mariomac:
I'm not sure if `placeholder` value was se there by a reason or it was just a mistake.


jotak:
isn't it something managed by olm when installing the operator?

mariomac:
@jotak yes I just verified it. It changes depending on the namespace you install it. Closing the PR

PR #200, opened 2022-11-22
NETOBSERV-719 Documentation update

jotak:
- Update configuration section (quick-filters, exporters)
- New section "performance fine-tuning"
- New section "securing data and communications"

PR #205, opened 2022-11-30
NETOBSERV-727 Update bundle process

jotak:
Requires https://github.com/netobserv/network-observability-operator/pull/200

The bundle process needs to work both for upstream and downstream
Every time someone changes something that would alter the bundle (e.g.
CSV description), they should run "make bundle" (with no options)

It will generate the CSV with injected upstream description.
The downstream description will be updated too, but not injected: this
step will be done later on, in midstream.

### DIFF
--- a/bundle/metadata/annotations.yaml
+++ b/bundle/metadata/annotations.yaml
@@ -4,9 +4,9 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: netobserv-operator
-  operators.operatorframework.io.bundle.channels.v1: v1.0.x
-  operators.operatorframework.io.bundle.channel.default.v1: v1.0.x
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.25.2
+  operators.operatorframework.io.bundle.channels.v1: v0.2.x
+  operators.operatorframework.io.bundle.channel.default.v1: v0.2.x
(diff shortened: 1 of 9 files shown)


bundle/metadata/annotations.yaml
```diff
@@ -4,9 +4,9 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: netobserv-operator
-  operators.operatorframework.io.bundle.channels.v1: v1.0.x
-  operators.operatorframework.io.bundle.channel.default.v1: v1.0.x
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.25.2
+  operators.operatorframework.io.bundle.channels.v1: v0.2.x
+  operators.operatorframework.io.bundle.channel.default.v1: v0.2.x
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.22.2
```
mariomac, line 9: Shall we use latest operator-sdk by default or stick to a given version?

mariomac (approved):

bundle/metadata/annotations.yaml
jotak: I think we don't need to enforce a specific version, as long as the produced bundle stays the same, minus that particular line ?
In the new github action to verify the bundle, this line is ignored (as well as the "createdAt" line in CSV)

jotak: given that, if someone uses a too old operator-sdk that produces different bundle, the github action will throw a failure, so that will force us to upgrade operator-sdk when necessary

PR #206, opened 2022-12-01
NETOBSERV-693: use ISO8601 for logging timestamps

mariomac:
Changed timestamps in epoch milliseconds by something like `2022-12-01T09:53:35.484Z`.


jpinsonneau (approved):
LGTM thanks @mariomac

PR #208, opened 2022-12-01
NETOBSERV-672: Document bug when flowcollector is deployed before kafka

OlivierCazade:
I tested and confirmed that the bug come from the kafka-go dependency.

I am trying to fix it and I opened a bug report in the library side.

Meanwhile for GA, I think it is better to document the workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,12 @@ kubectl get pods -n netobserv -l app=netobserv-plugin
 kubectl logs -n netobserv -l app=netobserv-plugin
 ```
 
+### I first deployed flowcollector, and then kafka. Flowlogs-pipeline is not consuming any flow from Kafka
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -237,6 +237,10 @@ kubectl get pods -n netobserv -l app=netobserv-plugin
 kubectl logs -n netobserv -l app=netobserv-plugin
 ```
 
+### I first deployed flowcollector, and then kafka. Flowlogs-pipeline is not consuming any flow from Kafka
+
+This is a known bug in one of flowlogs-pipeline dependencies. Please recreate the flowlogs-pipeline pods by killing them, they will be automatically recreated.
```
mariomac, line 242: Maybe add the link to the issue in the kafka lib, if it exists. I'd maybe suggest, as another alternative, to redeploy the flow collector, in addition to killing pods one by one.

mariomac (approved):

PR #210, opened 2022-12-05
Fix operator deploy to use main images by default

jotak:


jotak:
This is fixing the isue I had, with our deploying script not using the `main` images anymore.
There's a little issue: running `make deploy` leaves a bit more garbage than before, which should not be commited (if it's commited by mistake, the bundle-checker github action should complain)

jpinsonneau (approved):

PR #216, opened 2022-12-07
update upstream release doc

jotak:
fwiw, upgrade path worked correctly without the previous hack


jotak:
merging (it's just an update on the release instructions)

PR #219, opened 2022-12-08
Quick filters: add some examples in documentation

jotak:
[edit] I changed my approach, instead of adding new filters directly in CR, some being a bit hacky, I prefer just to document them as examples. Users can copy-paste them in their CR

--- 

This filter shows cluster traffic to outside.
Note there's a trick by excluding source port 443, this is to remove traffic from of kind "ingress to outside" ie. the response traffic of an ingress connection. It's a bit hacky.


jotak:
image

OlivierCazade (approved):

PR #227, opened 2022-12-14
Update golangci-lint

mariomac:
We can't update prow to 1.18 because golangci-lint is in an old version and can't recognize the newer format of some modules:

https://github.com/openshift/release/pull/34809
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/34809/rehearse-34809-pull-ci-netobserv-goflow2-kube-enricher-main-lint/1603049581493358592

We need first to update golangci-lint here and in release-4.12 to unlock the situation

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,5 @@
 linters:
   enable:
-    - deadcode
     - errcheck
     - errorlint
     - cyclop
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -1,27 +1,24 @@
...
-    - deadcode
     - errcheck
     - errorlint
     - cyclop
     - errname
     - exhaustive
     - exportloopref
     - gocritic
-    - goimports
+    - gofmt
```
mariomac, line 10: goimports complains with the generated sources. Will see if we can revert this change in a later version.

.golangci.yml
```diff
@@ -1,27 +1,24 @@
 linters:
   enable:
-    - deadcode
```
mariomac, line 3: deadcode, structcheck and varcheck linters are warned as removed by the linter, so I removed them from here.

OlivierCazade (approved):

mariomac:
Merged without passing tests because there is a deadlock: Prow needs to be updated to Go 1.18 but can't be updated if we don't fix linting here.

PR #234, opened 2022-12-16
NETOBSERV-704 Netflow traffic page doesn't appear sometimes

jpinsonneau:
Added mention on how to check and restart console pods to clear plugin cache

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -131,6 +131,18 @@ spec:
   - netobserv-plugin
 ```
 
+Ensure console pods are all in `Running` state using the following command:
+
+```bash
+oc get pods -n openshift-console -l app=console
+```
+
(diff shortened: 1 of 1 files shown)


mariomac (approved):

Amoghrd:
Small nit here @jpinsonneau all other commands in the doc use kubectl, for uniformity according to me its better to use kubectl instead of oc.

jpinsonneau:
> Small nit here @jpinsonneau all other commands in the doc use kubectl, for uniformity according to me its better to use kubectl instead of oc.

Usually we use `kubectl` for upstream and `oc` for downstream. Since this workaround is only openshift related, I have chosen `oc` here. Does it makes sense ?

Amoghrd:
ah okay, got it @jpinsonneau 👍

PR #237, opened 2022-12-21
Remove mention 'cannot be updated' on requests/limits

jotak:


jotak:
I'll backport it to 4.12 as it has an impact on the generated doc, which is wrong in that regard

mariomac (approved):
Shall we backport to `release-4.12`?

OlivierCazade (approved):

memodi:
> I'll backport it to 4.12 as it has an impact on the generated doc, which is wrong in that regard

to understand impact on downstream docs, are those copied/pasted from here upstream?

memodi (approved):

PR #243, opened 2023-01-04
NETOBSERV-772 check volume change in reconcile loop

jotak:
- Both for console plugin and FLP, need to check whether volumes changes in reconcile loop
- Some refactoring to share common comparison functions
- Remove namespace comparisons in several places because they are handled differently (namespace changes doesn't trigger the usual workload updates, it deletes all / recreate all)
- Add some tests

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -30,10 +30,6 @@ const configPath = "/opt/app-root/"
 const lokiCerts = "loki-certs"
 const tokensPath = "/var/run/secrets/tokens/"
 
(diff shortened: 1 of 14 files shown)


jpinsonneau (approved):
LGTM thanks @jotak !

Tested both CA configmap name change and namespace change.
I found a bug while changing namespace where old ebpf agent in privileged namespace are not deleted
image

I'm opening a bug since I can reproduce on main branch.
Related slack: https://coreos.slack.com/archives/C02939DP5L5/p1673001363041479

jotak:
Thanks Julien
I've opened another PR for the agent cleanup bug
/approve

PR #244, opened 2023-01-06
Fix typos in CSV desc

jotak:


jotak:
IMO we should backport it as the risk is very low (just typo fixes) ..

jpinsonneau (approved):
:+1:  thanks @jotak

PR #251, opened 2023-01-18
NETOBSERV-788: WIP, not working: v1 of FlowCollector + conversion webhook

mariomac:
This PR also updates the operator-sdk version as well as some library versions.

This is what I'd expect to have: to add a mutating webhook so every time I deploy a v1alpha1.FlowCollector, it is automatically converted to v1.FlowCollector.

However, at this given moment, if I try to deploy a v1alpha1 FlowCollector, I get this error:

```
Error from server: error when creating "flows_v1alpha1_flowcollector.yaml.out": conversion webhook for
flows.netobserv.io/v1alpha1, Kind=FlowCollector returned invalid object at index 0: invalid groupVersion (expected 
flows.netobserv.io/v1, received flows.netobserv.io/v1alpha1)
```

It's like the converter implemented in the `flowcollector_webhook.go` file did the opposite job. Maybe the webhook should have been defined for the v1 version?

If I deploy a v1 Flowcollector, it keeps showing this message:

```
2023-01-23T13:45:49.803Z	INFO	flowcollector-resource	converting v1.FlowCollector into v1alpha1.FlowCollector
```

Also, the generated manifests have a comment saying:

> Do NOT uncomment sections with prefix [CERTMANAGER], as OLM does not support cert-manager.

But  I haven't found a way to deploy the webhook without using the cert-manager operator, even with the provided patches in `config/manifests/kustomization.yml`. The generated sources say:

On the other side, the cert-manager seems to be supported at some time in the futer, as I see an actual downstream cert-manager in 4.12 OperatorHub:
image

But it seems to be currently in tech-preview:
image


jotak:
did you use `operator-sdk` scaffolding?

mariomac:
> did you use operator-sdk scaffolding?

yes, I did.

PR #252, opened 2023-01-24
NETOBSERV-567 Enable connection tracking in NOO

jpinsonneau:
To allow working / testing https://issues.redhat.com/browse/NETOBSERV-568 I started working on enabling connection tracking in netobserv operator.

The only configurable options are `connectionUpdateInterval` and `ConnectionEndTimeout` for consistency in console plugin.


ronensc (dismissed):

jpinsonneau:
Holding this PR as we need console plugin merged at the same time

/hold

api/v1alpha1/flowcollector_types.go
```diff
@@ -358,6 +358,16 @@ type FlowCollectorFLP struct {
 	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
 	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
 
+	//+kubebuilder:validation:Pattern:=^\d+(ns|ms|s|m)?$
+	//+kubebuilder:default:="30s"
+	// connection update interval is the duration of time to wait between update reports of a connection
+	ConnectionUpdateInterval string `json:"connectionUpdateInterval,omitempty"`
```
OlivierCazade, line 364: If I understand correctly, we should not modify anymore the v1alpha1 CRD

OlivierCazade (changes requested):

api/v1alpha1/flowcollector_types.go
jotak: I'm not 100% sure but I believe if the field is optional, and if we consider in code that it can be empty, then it could be still done as part of v1alpha1
(cc @msherif1234 )

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -300,6 +301,77 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
...
+
+	// Connection tracking stage
+	endTimeout, err := time.ParseDuration(b.desired.Processor.ConnectionEndTimeout)
+	if err != nil {
+		return err
+	}
+	updateInterval, err := time.ParseDuration(b.desired.Processor.ConnectionUpdateInterval)
+	if err != nil {
+		return err
+	}
```
jotak, line 313: I think this is where we need to be careful: even if there's a default duration provided in the CRD, we must assume, after an upgrade (and if we still use v1alpha1), those values could be empty strings. I'm not 100% sure of that though

jotak: cc @OlivierCazade @msherif1234

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -300,6 +301,77 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
...
+				Name:      "TimeFlowEndMs",
+				Operation: "max",
+			},
+		},
+		EndConnectionTimeout: api.Duration{
+			Duration: endTimeout,
+		},
+		UpdateConnectionInterval: api.Duration{
+			Duration: updateInterval,
+		},
```
ronensc, line 372: Once https://github.com/netobserv/flowlogs-pipeline/pull/357 is merged, this snippet should be change to something like:
```go
	Scheduling: []api.ConnTrackSchedulingGroup{
		{
			Selector:                 nil, // Default group. Match all flowlogs
			UpdateConnectionInterval: api.Duration{Duration: endTimeout},
			EndConnectionTimeout:     api.Duration{Duration: updateInterval},
		},
	},
```

jpinsonneau: Thanks for the info @ronensc !

api/v1alpha1/flowcollector_types.go
jpinsonneau: Let's follow this in https://github.com/netobserv/network-observability-operator/pull/258#discussion_r1100457798

controllers/flowlogspipeline/flp_common_objects.go
jpinsonneau: These should not at this level since it's supposed to be converted by `Convert_v1alpha1_FlowCollector_To_v1beta1_FlowCollector` function

https://github.com/netobserv/network-observability-operator/pull/258/files#diff-23e9a27bd609a9414ced41fdaf8063ade03a51b3154d85ad4edfdae228dd5531

api/v1alpha1/flowcollector_types.go
OlivierCazade: Following the discussion in #258 v1beta1 will not be merged for now

OlivierCazade (approved):
LGTM

controllers/flowlogspipeline/flp_common_objects.go
jpinsonneau: Is there a plan to create a new FLP release for this ? @jotak 

https://github.com/netobserv/network-observability-operator/pull/252/files#diff-c761638f65b982f9ecf3717616920a5a3c0bb1670fd8ee7e9ab51afc4f166581R378

https://github.com/netobserv/network-observability-operator/pull/252/files#diff-c761638f65b982f9ecf3717616920a5a3c0bb1670fd8ee7e9ab51afc4f166581R407

Makefile
```diff
@@ -133,11 +140,10 @@ doc: crdoc ## Generate
 	$(CRDOC) --resources config/crd/bases/flows.netobserv.io_flowcollectors.yaml --output docs/FlowCollector.md
 
 generate-go-conversions: $(CONVERSION_GEN) ## Run all generate-go-conversions
-		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1alpha1"
 		$(CONVERSION_GEN) \
 		--input-dirs=./api/v1alpha1 \
 		--build-tag=ignore_autogenerated_core \
-		--output-file-base=zz_generated.conversion \
+		--output-file-base=zz_generated.conversion $(CONVERSION_GEN_OUTPUT_BASE)\
```
jpinsonneau, line 146: We had to push this to fix `conversion_gen` missing file generating lint error.

For now, [I have disabled the conversion](https://github.com/netobserv/network-observability-operator/pull/252/files#diff-23e9a27bd609a9414ced41fdaf8063ade03a51b3154d85ad4edfdae228dd5531R45) and added the new fields in both v1alpha1 and v1beta1 apis

api/v1alpha1/flowcollector_types.go
```diff
@@ -358,6 +358,18 @@ type FlowCollectorFLP struct {
 	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
 	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
 
+	//+kubebuilder:validation:Pattern:=^\d+(ns|ms|s|m)?$
+	//+kubebuilder:default:="30s"
+	// +optional
+	// connection update interval is the duration of time to wait between update reports of a connection
+	ConnectionUpdateInterval *string `json:"connectionUpdateInterval,omitempty"`
```
msherif1234, line 365: pls use `metav1.Duration` here instead of `string` with that u won't need to have the regex validation same to the other new field

Makefile
msherif1234: I am looking at this and also created an issue for the conversion-gen tool 
https://github.com/kubernetes/code-generator/issues/151

msherif1234: @jpinsonneau pls include this change to fix the conversion issue
https://github.com/netobserv/network-observability-operator/pull/268/commits/579cd3d525736634a72c703e023332417bb62d48#diff-23e9a27bd609a9414ced41fdaf8063ade03a51b3154d85ad4edfdae228dd5531

jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/252/commits/4a689b594c52a36ddd292e0303d3fd5e02ad6130 here we go, thanks @msherif1234 !

api/v1alpha1/flowcollector_types.go
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/252/commits/eb2796d108e620ff160f28c01828425e95973e15 done !

controllers/flowlogspipeline/flp_common_objects.go
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/252/commits/decd18365a3da2a23ff09c8a0b8be65c0c208e23

jpinsonneau:
/unhold

- [X] conversion wehbook fixed
- [X] FLP updated to `v0.1.8-rc0`
  - renamed updateConnection to `heartbeat`
  - Updated `ConnTrackSchedulingGroup`
  - renammed `DirectionInfo` to `NetworkTransformDirectionInfo`

Tested on OCP 4.12 and working fine from `v1alpha1` CRD

.mk/development.mk
```diff
@@ -142,23 +142,23 @@ get-related-images:
...
 	kubectl wait -n $(NAMESPACE) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
 	@echo -e "\n==> Wait a moment before FLP is fully redeployed"
 
 .PHONY: set-plugin-image
 set-plugin-image:
-	kubectl set env deployment netobserv-controller-manager -c "manager" RELATED_IMAGE_CONSOLE_PLUGIN=quay.io/$(USER)/network-observability-console-plugin:$(VERSION)
+	kubectl set env -n $(NAMESPACE) deployment netobserv-controller-manager -c "manager" RELATED_IMAGE_CONSOLE_PLUGIN=quay.io/$(USER)/network-observability-console-plugin:$(VERSION)
 	@echo -e "\n==> Redeploying..."
 	kubectl rollout status -n $(NAMESPACE) --timeout=60s deployment netobserv-controller-manager
 	kubectl wait -n $(NAMESPACE) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
```
msherif1234, line 164: we probably need to fix `get-related-images:`as well ?

Makefile
```diff
@@ -225,7 +231,7 @@ run: generate fmt lint ## Run a controller from your host.
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.1)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.0)
```
msherif1234, line 234: we probably don't need to update controller-gen at least in this PR, WDYT ?

Makefile
```diff
@@ -133,11 +140,10 @@ doc: crdoc ## Generate
 	$(CRDOC) --resources config/crd/bases/flows.netobserv.io_flowcollectors.yaml --output docs/FlowCollector.md
 
 generate-go-conversions: $(CONVERSION_GEN) ## Run all generate-go-conversions
-		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1alpha1"
```
msherif1234, line 136: pls add the cleanup back so we don't endup with stale auto generated conversion code

config/rbac/role.yaml
```diff
@@ -1,4 +1,3 @@
-
 ---
```
msherif1234, line 1: no changes in this file pls revert it

api/v1beta1/flowcollector_types.go
```diff
@@ -358,6 +358,16 @@ type FlowCollectorFLP struct {
 	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
 	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
 
+	//+kubebuilder:default:="30s"
+	// +optional
+	// connection update interval is the duration of time to wait between update reports of a connection
```
msherif1234, line 363: is there a min/max value for setting this ? and the new field blow ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -302,6 +303,80 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 		}
 		// Else: nothing for eBPF at the moment
 	}
+
+	// Connection tracking stage
+	endTimeout := 10 * time.Second
```
msherif1234, line 308: nit: create a new constant for both times and use it here and in the test below

.mk/development.mk
jpinsonneau: yes it seems to ! Thanks

Makefile
jpinsonneau: yes I will revert that. Is there a followup somewhere ? I feel we should keep up to date on these.

Makefile
jpinsonneau: Sure, done

config/rbac/role.yaml
jpinsonneau: I'm not sure how this one was updated but yeah I can revert :+1:

api/v1beta1/flowcollector_types.go
jpinsonneau: There are no min / max values for now. I guess we need to test deeper and even get feedback before forcing such.

@ronensc what do you think ?

controllers/flowlogspipeline/flp_test.go
```diff
@@ -83,6 +84,12 @@ func getConfig() flowslatest.FlowCollectorSpec {
 					},
 				}},
 			},
+			ConnectionUpdateInterval: &metav1.Duration{
+				Duration: 30 * time.Second,
```
msherif1234, line 88: nit: use`conntrackUpdateInterval` and `conntrackEndTimeout`

OlivierCazade (approved):
LGTM

jpinsonneau:
[Rebased + removed makefile changes](https://github.com/netobserv/network-observability-operator/compare/babd0d5cad75e82424de83d8c57bffc1bf400b4b..a0f7edba679161b64870bee1a759f789337f1129)

jpinsonneau:
[As suggested](https://redhat-internal.slack.com/archives/C02939DP5L5/p1676971547474739) connection tracking is now optionnal.

api/v1beta1/flowcollector_types.go
```diff
@@ -358,6 +358,21 @@ type FlowCollectorFLP struct {
 	// kafkaConsumerBatchSize indicates to the broker the maximum batch size, in bytes, that the consumer will accept. Ignored when not using Kafka. Default: 10MB.
 	KafkaConsumerBatchSize int `json:"kafkaConsumerBatchSize"`
 
+	//+kubebuilder:default:={"flowLog"}
+	// outputRecordTypes is a list of record types to export from ["flowLog","newConnection","heartbeat","endConnection"]
+	// +optional
+	OutputRecordTypes *[]string `json:"outputRecordTypes,omitempty"`
```
msherif1234, line 364: is there min number of elements in this list in another words can it be empty or it should have at minimum one entry ?

jpinsonneau: That's a good question. I would like to gather some feedback here before taking a decision.

We could either keep this open using a string array or force specific cases using an enum.
The connection tracking stage is really flexible and will throw an error [only if a provided record type is unknown](https://github.com/netobserv/flowlogs-pipeline/blob/9b96359be572c2b9d0fcf5620d5be3386f8a955d/pkg/pipeline/extract/conntrack/conntrack.go#L240)
- Providing an empty array here could be an option to say ["only export prometheus metrics"](https://github.com/netobserv/network-observability-operator/pull/260). I probably need to update a bit the pipeline for that. 
 
Loki will become optionnal and the console plugin would be disabled.

- It would allow also to only export all / some connection events (`NewConnection` / `Heartbeat` / `EndConnection`) without flowlogs.
- Or only `flowLog` as you can see in the default value.

The console plugin UI adapts options accordingly in these cases.

What do you think ?

msherif1234: IMO I think using enum with default here will be more appropriate but we should also hear what others thoughts on this

api/v1beta1/flowcollector_types.go
ronensc: I agree. There are no min/max. Also the default values were set arbitrarily.

api/v1beta1/flowcollector_types.go
```diff
@@ -358,6 +358,21 @@ type FlowCollectorFLP struct {
...
+
+	//+kubebuilder:default:="30s"
+	// +optional
+	// connection update interval is the duration of time to wait between update reports of a connection
+	ConnectionUpdateInterval *metav1.Duration `json:"connectionUpdateInterval,omitempty"`
```
ronensc, line 369: Are we ok with keeping the different setting names `ConnectionUpdateInterval` in the operator and `HeartbeatInterval` in FLP?

api/v1beta1/flowcollector_types.go
ronensc: > Providing an empty array here could be an option to say "only export prometheus metrics". I probably need to update a bit the pipeline for that.

IIUC, the proposal is that an empty array will be equivilant to an array with a single value "flowLog" which indicates disabling connection tracking?

api/v1beta1/flowcollector_types.go
jpinsonneau: Makes sense to be more aligned. Would `ConnectionHeartbeatInterval` work ?

api/v1beta1/flowcollector_types.go
jpinsonneau: @ronensc no; I'm suggesting to be able to entirely skip Loki stage with empty array. Would it make sense ?

FYI I have played a bit with empty / default value and the yaml act as:
```
outputRecordTypes:
connectionUpdateInterval: 30s
connectionEndTimeout: 10s
```
-> is falling back to the default value `flowLog`

```
outputRecordTypes: []
connectionUpdateInterval: 30s
connectionEndTimeout: 10s
```
-> is explicitly using empty string but it feels a bit hacky

api/v1beta1/flowcollector_types.go
ronensc: Sure!

I wonder whether I should have renamed it to `ConnectionHeartbeatInterval` instead of `HeartbeatInterval` in FLP... 😅

api/v1beta1/flowcollector_types.go
ronensc: Now I see. Hm... I can't think of a reason for users to set an empty array.
I mean, if the purpose is to skip Loki, wouldn't it be more expressive to remove Loki from the pipeline?
In addition, if the conntrack stage has no output records at all, then the prometheus stage which follows it won't be able to compute metrics.

api/v1beta1/flowcollector_types.go
jpinsonneau: In `conntrack` stage it makes sense; it's less obvious in the CRD as we are in the `processor` config here

api/v1beta1/flowcollector_types.go
jpinsonneau: sure, let's move to an enum then :+1: 

thanks @msherif1234 & @ronensc

jpinsonneau:
Addressed feedback on ConnectionHeartbeatInterval renaming + OutputRecordTypes enum

jpinsonneau:
@msherif1234 splitted commits as requested

@Amoghrd also [fixed the sample yaml](https://github.com/netobserv/network-observability-operator/compare/b7aa6c50573a0a2b2f6fb8e4ff659dcce499ca34..abda7ed572cfebb94b9fdc50957c6fe773c5df27#diff-f56c4ef1268ddd69d4da318cd3ffc855de718f70c7de5de760fd18ef69f79aebR48)

I have updated to flp v0.1.8 and rebased as there was conflicts in go dependencies

/ok-to-test

PR #259, opened 2023-02-01
NETOBSERV-855 do not recommend using HOST mode

jotak:


api/v1alpha1/flowcollector_types.go
```diff
@@ -452,9 +452,9 @@ type FlowCollectorLoki struct {
 	//+kubebuilder:default:="DISABLED"
 	// AuthToken describe the way to get a token to authenticate to Loki.
 	// DISABLED will not send any token with the request.
-	// HOST will use the local pod service account to authenticate to Loki.
-	// FORWARD will forward user token, in this mode, pod that are not receiving user request like the processor will use the local pod service account. Similar to HOST mode.
-	// When using the Loki Operator, set it to `HOST` or `FORWARD`.
+	// HOST will use the local pod service account to authenticate to Loki, without any check on the RBAC permissions of the logged-in user (hence it is not secure).
+	// FORWARD will forward user token, in this mode, pod that are not receiving user request like the processor will use the local pod service account.
+	// When using the Loki Operator, we recommend using `FORWARD`.
```
stleerh, line 457: We should take a stronger stance.  Suggest:  
When using the Loki Operator, use \`FORWARD\` in production as HOST is insecure.

jotak: Ok but I would add ".. unless you secure the netobserv namespace access with a network policy" ; and btw we should provide an example of network policy that mitigates this problem.

jpinsonneau (approved):

jotak:
/hold
Putting on hold; I would like to see if we can have a decent protection with HOST mode when protected with network policies ... knowing that this is not only the netobserv namespace that has to be restricted, but also the console one - because of the proxy endpoints

api/v1alpha1/flowcollector_types.go
stleerh: Ok.  Maybe with the latest revelation on HOST, nothing needs to be changed.

PR #260, opened 2023-02-02
Grafana dashboards

KalmanMeth:
Take grafana dashboards created by flp confgenerator [https://github.com/netobserv/flowlogs-pipeline/pull/369] and insert them in the Openshift console.
It is required to first merge FLP PR369, then update go.mod, go mod vendor, etc, and then to merge this PR.

### DIFF
--- a/.mk/ocp.mk
+++ b/.mk/ocp.mk
@@ -45,7 +45,10 @@ undeploy-operator: ## stop the operator locally
 	kubectl delete servicemonitor flowlogs-pipeline-monitor || true
 	kubectl delete service netobserv-plugin || true
 	kubectl delete deployment netobserv-plugin || true
-	kubectl delete servicemonitor netobserv-console-plugin || true
+	kubectl delete servicemonitor netobserv-plugin || true
+	kubectl delete configmap -n openshift-config-managed flowlogs-pipeline-metrics-dashboard || true
+	kubectl delete configmap -n netobserv flowlogs-pipeline-config || true
+	kubectl delete configmap -n netobserv console-plugin-config || true
 
 .PHONY: ocp-refresh-ovs
 ocp-refresh-ovs:
--- a/controllers/flowlogspipeline/flp_common_objects.go
+++ b/controllers/flowlogspipeline/flp_common_objects.go
@@ -55,6 +55,12 @@ const (
 	ConfKafkaTransformer ConfKind = "kafkaTransformer"
 )
 
+const (
+	dashboardName  = "netobserv"
+	dashboardTitle = "Netobserv Metrics"
+	dashboardTags  = "['netobserv','grafana','dashboard','flp']"
+)
+
 var FlpConfSuffix = map[ConfKind]string{
 	ConfMonolith:         "",
 	ConfKafkaIngester:    "-ingester",
@@ -111,11 +117,13 @@ func RoleBindingName(ck ConfKind) string      { return name(ck) + "-role" }
 func RoleBindingMonoName(ck ConfKind) string  { return name(ck) + "-role-mono" }
 func promServiceName(ck ConfKind) string      { return name(ck) + "-prom" }
 func configMapName(ck ConfKind) string        { return name(ck) + "-config" }
+func dbConfigMapName(ck ConfKind) string      { return name(ck) + "-metrics-dashboard" }
 func serviceMonitorName(ck ConfKind) string   { return name(ck) + "-monitor" }
 func prometheusRuleName(ck ConfKind) string   { return name(ck) + "-alert" }
 func (b *builder) name() string               { return name(b.confKind) }
 func (b *builder) promServiceName() string    { return promServiceName(b.confKind) }
 func (b *builder) configMapName() string      { return configMapName(b.confKind) }
+func (b *builder) dbConfigMapName() string    { return dbConfigMapName(b.confKind) }
 func (b *builder) serviceMonitorName() string { return serviceMonitorName(b.confKind) }
 func (b *builder) prometheusRuleName() string { return prometheusRuleName(b.confKind) }
 
@@ -259,10 +267,12 @@ func (b *builder) podTemplate(hasHostPort, hasLokiInterface, hostNetwork bool, c
 //go:embed metrics_definitions
 var metricsConfigEmbed embed.FS
 
-func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, error) {
+// obtainMetricsConfiguration returns the configuration info for the prometheus stage needed to
+// supply the metrics and also the dashboards for those metrics
+func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, string, error) {
 	entries, err := metricsConfigEmbed.ReadDir(metricsConfigDir)
 	if err != nil {
-		return nil, fmt.Errorf("failed to access metrics_definitions directory: %w", err)
+		return nil, "", fmt.Errorf("failed to access metrics_definitions directory: %w", err)
 	}
 
 	cg := confgen.NewConfGen(&confgen.Options{
@@ -270,31 +280,52 @@ func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, error) {
 		SkipWithTags:   b.desired.Processor.Metrics.IgnoreTags,
 	})
 
+	config := confgen.Config{
+		Visualization: confgen.ConfigVisualization{
+			Grafana: confgen.ConfigVisualizationGrafana{
+				Dashboards: []confgen.ConfigVisualizationGrafanaDashboard{
+					{
+						Name:          dashboardName,
+						Title:         dashboardTitle,
+						TimeFrom:      "now",
+						Tags:          dashboardTags,
+						SchemaVersion: "16",
+					},
+				},
+			},
+		},
+	}
+	cg.SetConfig(&config)
+
 	for _, entry := range entries {
 		fileName := entry.Name()
+		if fileName == "config.yaml" {
+			continue
+		}
 		srcPath := filepath.Join(metricsConfigDir, fileName)
 
 		input, err := metricsConfigEmbed.ReadFile(srcPath)
 		if err != nil {
-			return nil, fmt.Errorf("error reading metrics file %s; %w", srcPath, err)
+			return nil, "", fmt.Errorf("error reading metrics file %s; %w", srcPath, err)
 		}
 		err = cg.ParseDefinition(fileName, input)
 		if err != nil {
-			return nil, fmt.Errorf("error parsing metrics file %s; %w", srcPath, err)
+			return nil, "", fmt.Errorf("error parsing metrics file %s; %w", srcPath, err)
 		}
 	}
 
 	stages := cg.GenerateTruncatedConfig()
 	if len(stages) != 1 {
-		return nil, fmt.Errorf("error generating truncated config, 1 stage expected in %v", stages)
+		return nil, "", fmt.Errorf("error generating truncated config, 1 stage expected in %v", stages)
 	}
 	if stages[0].Encode == nil || stages[0].Encode.Prom == nil {
-		return nil, fmt.Errorf("error generating truncated config, Encode expected in %v", stages)
+		return nil, "", fmt.Errorf("error generating truncated config, Encode expected in %v", stages)
 	}
-	return stages[0].Encode.Prom.Metrics, nil
+	jsonStr, _ := cg.GenerateGrafanaJson()
+	return stages[0].Encode.Prom.Metrics, jsonStr, nil
 }
 
-func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
+func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev1.ConfigMap, error) {
 	lastStage := *stage
 	indexFields := constants.LokiIndexFields
 
@@ -464,9 +495,9 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 	}
 
 	// obtain encode_prometheus stage from metrics_definitions
-	promMetrics, err := b.obtainMetricsConfiguration()
+	promMetrics, dashboard, err := b.obtainMetricsConfiguration()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	// prometheus stage (encode) configuration
@@ -486,9 +517,25 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 	enrichedStage.EncodePrometheus("prometheus", promEncode)
 	b.addCustomExportStages(&enrichedStage)
 
-	return nil
+	dashboardConfigMap := b.makeMetricsDashboardConfigMap(dashboard)
+	return dashboardConfigMap, nil
 }
 
+func (b *builder) makeMetricsDashboardConfigMap(dashboard string) *corev1.ConfigMap {
+	configMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      b.dbConfigMapName(),
+			Namespace: "openshift-config-managed",
+			Labels: map[string]string{
+				"console.openshift.io/dashboard": "true",
+			},
+		},
+		Data: map[string]string{
+			"netobserv-metrics.json": dashboard,
+		},
+	}
+	return &configMap
+}
 func (b *builder) addCustomExportStages(enrichedStage *config.PipelineBuilderStage) {
 	for i, exporter := range b.desired.Exporters {
 		if exporter.Type == flowslatest.KafkaExporter {
--- a/controllers/flowlogspipeline/flp_monolith_reconciler.go
+++ b/controllers/flowlogspipeline/flp_monolith_reconciler.go
@@ -94,7 +94,7 @@ func (r *flpMonolithReconciler) reconcile(ctx context.Context, desired *flowslat
 	}
 
 	builder := newMonolithBuilder(r.nobjMngr.Namespace, r.image, &desired.Spec, r.useOpenShiftSCC, r.CertWatcher)
-	newCM, configDigest, err := builder.configMap()
+	newCM, configDigest, dbConfigMap, err := builder.configMap()
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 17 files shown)


go.mod
```diff
@@ -22,4 +22,77 @@ require (
...
+	k8s.io/klog/v2 v2.60.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
+	sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
 replace github.com/prometheus/common v0.32.1 => github.com/netobserv/prometheus-common v0.31.2-0.20220720134304-43e74fd22881
+
+replace github.com/netobserv/flowlogs-pipeline v0.1.7-0.20221221173558-e6e77158b956 => github.com/KalmanMeth/flowlogs2metrics v0.0.0-20230129124539-a447dc625fd4
```
jpinsonneau, line 98:
```suggestion
```

jpinsonneau (commented):
Looks good, thanks @KalmanMeth ! Just small suggestions on my side

controllers/flowlogspipeline/flp_monolith_reconciler.go
```diff
@@ -91,18 +92,28 @@ func (r *flpMonolithReconciler) reconcile(ctx context.Context, desired *flowsv1a
...
-	newCM, configDigest, err := builder.configMap()
+	newCM, configDigest, dbConfigMap, err := builder.configMap()
 	if err != nil {
 		return err
 	}
 	if !r.nobjMngr.Exists(r.owned.configMap) {
 		if err := r.CreateOwned(ctx, newCM); err != nil {
 			return err
 		}
+		if r.reconcilersCommonInfo.availableAPIs.HasConsole() {
```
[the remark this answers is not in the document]
jpinsonneau: @jotak do you know a good way to check this ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -254,42 +256,65 @@ func (b *builder) podTemplate(hasHostPort, hasLokiInterface, hostNetwork bool, c
...
 		SkipWithTags:   b.desired.Processor.Metrics.IgnoreTags,
 	})
 
+	config := confgen.Config{
+		Visualization: confgen.ConfigVisualization{
+			Grafana: confgen.ConfigVisualizationGrafana{
+				Dashboards: []confgen.ConfigVisualizationGrafanaDashboard{
+					{
+						Name:          "netobserv",
+						Title:         "Netobserv Metrics",
```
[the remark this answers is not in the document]
KalmanMeth: done

controllers/flowlogspipeline/flp_monolith_reconciler.go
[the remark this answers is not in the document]
KalmanMeth: Not sure about this. I tried to run via `make ocp-deploy-operator` with checking for `consoles.config.openshift.io` and I got false. But if I load the dashboards anyway they show up in the console under `Observe->Dashboards->Netobserv Metrics`.

go.mod
KalmanMeth: No longer relevant. I included the latest release of flp confgenerator, which now includes the function to create the dashboards.

controllers/flowlogspipeline/flp_monolith_reconciler.go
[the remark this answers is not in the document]
jpinsonneau: That's what you should get from oc command:
```
$ oc get consoles.config.openshift.io
NAME      AGE
cluster   29d
```

The `Observe->Dashboards` belongs to OCP console and not the dynamic plugin. That's why I'm suggesting this small change.

controllers/flowlogspipeline/flp_common_objects.go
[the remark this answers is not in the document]
jpinsonneau: thanks @KalmanMeth !

jpinsonneau:
@KalmanMeth I'm trying that on cluster bot using OCP 4.12.1; The dashboard is correctly created but I don't get data in the graphs.

image

FLP is using `quay.io/netobserv/flowlogs-pipeline:main` image with following config:
```
{
  "health": {
    "port": 8080
  },
  "log-level": "info",
  "metrics-settings": {
    "prefix": "netobserv_",
    "noPanic": true
  },
  "parameters": [
    {
      "name": "grpc",
      "ingest": {
        "type": "grpc",
        "grpc": {
          "port": 2055
        }
      }
    },
    {
      "name": "enrich",
      "transform": {
        "type": "network",
        "network": {
          "rules": [
            {
              "input": "SrcAddr",
              "output": "SrcK8S",
              "type": "add_kubernetes"
            },
            {
              "input": "DstAddr",
              "output": "DstK8S",
              "type": "add_kubernetes"
            },
            {
              "type": "reinterpret_direction"
            }
          ],
          "directionInfo": {
            "reporterIPField": "AgentIP",
            "srcHostField": "SrcK8S_HostIP",
            "dstHostField": "DstK8S_HostIP",
            "flowDirectionField": "FlowDirection",
            "ifDirectionField": "IfDirection"
          }
        }
      }
    },
    {
      "name": "loki",
      "write": {
        "type": "loki",
        "loki": {
          "url": "http://loki.netobserv.svc:3100/",
          "tenantID": "netobserv",
          "batchWait": "1s",
          "batchSize": 10485760,
          "timeout": "10s",
          "minBackoff": "1s",
          "maxBackoff": "5s",
          "maxRetries": 2,
          "labels": [
            "SrcK8S_Namespace",
            "SrcK8S_OwnerName",
            "DstK8S_Namespace",
            "DstK8S_OwnerName",
            "FlowDirection"
          ],
          "staticLabels": {
            "app": "netobserv-flowcollector"
          },
          "clientConfig": {
            "proxy_url": null,
            "tls_config": {
              "insecure_skip_verify": false
            },
            "follow_redirects": false
          },
          "timestampLabel": "TimeFlowEndMs",
          "timestampScale": "1ms"
        }
      }
    },
    {
      "name": "prometheus",
      "encode": {
        "type": "prom",
        "prom": {
          "metrics": [
            {
              "name": "namespace_flows_total",
              "type": "counter",
              "filter": {
                "key": "",
                "value": ""
              },
              "valueKey": "",
              "labels": [
                "SrcK8S_Namespace",
                "DstK8S_Namespace"
              ],
              "buckets": null
            },
            {
              "name": "node_ingress_bytes_total",
              "type": "counter",
              "filter": {
                "key": "FlowDirection",
                "value": "0"
              },
              "valueKey": "Bytes",
              "labels": [
                "SrcK8S_HostName",
                "DstK8S_HostName"
              ],
              "buckets": null
            },
            {
              "name": "workload_ingress_bytes_total",
              "type": "counter",
              "filter": {
                "key": "FlowDirection",
                "value": "0"
              },
              "valueKey": "Bytes",
              "labels": [
                "SrcK8S_Namespace",
                "DstK8S_Namespace",
                "SrcK8S_OwnerName",
                "DstK8S_OwnerName",
                "SrcK8S_OwnerType",
                "DstK8S_OwnerType"
              ],
              "buckets": null
            }
          ],
          "port": 9102,
          "prefix": "netobserv_"
        }
      }
    }
  ],
  "pipeline": [
    {
      "name": "grpc"
    },
    {
      "name": "enrich",
      "follows": "grpc"
    },
    {
      "name": "loki",
      "follows": "enrich"
    },
    {
      "name": "prometheus",
      "follows": "enrich"
    }
  ],
  "profile": {
    "port": 6060
  }
}
```

Am I missing something ?

KalmanMeth:
@jpinsonneau Did you run the hack/enable_metrics.sh script? Do you have data being generated, such as from the sample workload?

jpinsonneau:
> @jpinsonneau Did you run the hack/enable_metrics.sh script? Do you have data being generated, such as from the sample workload?

Thanks ! the content is now loading.
I thought that part was also automated. Is there any reason to avoid creating this automatically ?

Also I have issues with labels appearing empty:
image

controllers/flowlogspipeline/flp_monolith_reconciler.go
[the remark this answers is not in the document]
KalmanMeth: done

KalmanMeth:
@jpinsonneau 
> Also I have issues with labels appearing empty:

Yes, I also do not see proper labels. I might need some help to understand what is needed here. I am continuing to investigate.

controllers/flowlogspipeline/flp_monolith_reconciler.go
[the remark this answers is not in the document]
jpinsonneau: awesome, thanks @KalmanMeth !

jpinsonneau:
> Yes, I also do not see proper labels. I might need some help to understand what is needed here. I am continuing to investigate.

You should add `legendFormat` in targets such as:
```
 "targets": [
    {
       "expr": "topk(5,rate(netobserv_namespace_flows_total[1m]))",
       ...
       "legendFormat": "{{pod}}:{{Value}}",
    }
 ],
```

controllers/flowlogspipeline/metrics_definitions/namespace_flows_total.yaml
```diff
@@ -21,7 +21,7 @@ visualization:
   type: grafana
   grafana:
     - expr: 'topk(5,rate(netobserv_namespace_flows_total[1m]))'
-      legendFormat: '{{SrcK8S_Namespace}} : {{DstK8S_Namespace}}'
+      legendFormat: '{{pod}} : {{DstK8S_Namespace}}'
```
jpinsonneau, line 24: For all legendFormat, what about using the same format as console plugin ?
When namespace is available`{{namespace}}.{{name}}` (resource / owner) else only `{{name}}` (namespace / host / ip) for every peers
https://github.com/netobserv/network-observability-console-plugin/blob/68239b88737f676152a9cf51e3b58bc6b781bd75/web/src/utils/metrics.ts#L77

Then we use `->` between source and destination as:
`{{source}} -> {{destination}}`
https://github.com/netobserv/network-observability-console-plugin/blob/fe28d186fcd9acc1078c54db7690e86d6c83a293/web/src/components/metrics/metrics-helper.tsx#L147

Here is the result:
image

KalmanMeth: I added `->` between source and destination. Instead of `.` separating `namespace` and `owner` I used `,` because sometimes the name of an owner or namespace already includes `.` .

I included the `pod` in the legend because we get separate statistics for each flowlogs-pipeline instance that is running. Since the different FLPs report separately and have different labels, we do not have a cumulative summed total, since the same metrics may be spread across multiple FLPs. We have to think about how we want to handle this.

We can further change the legends in a future PR by simply changing the metrics_definitions yaml files, with no additional changes in the code.

jpinsonneau: Yes it's good enough to merge as is. +1 for followup on totals

 Thanks @KalmanMeth !

config/manager/kustomization.yaml
```diff
@@ -14,7 +14,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/netobserv/network-observability-operator
-  newTag: 1.0.1
+  newName: quay.io/kalmanmeth/network-observability-operator
+  newTag: test
```
jpinsonneau, line 18: Could you please revert this before we merge @KalmanMeth ?

jpinsonneau (changes requested):

config/manager/kustomization.yaml
KalmanMeth: done

jpinsonneau (dismissed):
LGTM, tested on OCP `4.12.4` and works fine :+1: 
Thanks @KalmanMeth !

image

controllers/flowlogspipeline/flp_monolith_reconciler.go
```diff
@@ -40,6 +40,7 @@ func newMonolithReconciler(info *reconcilersCommonInfo) *flpMonolithReconciler {
 		promService:    &corev1.Service{},
 		serviceAccount: &corev1.ServiceAccount{},
 		configMap:      &corev1.ConfigMap{},
+		dbConfigMap:    &corev1.ConfigMap{},
```
jotak, line 43: this new config map should also be added with `AddManagedObject`, same as the existing configmap. This is necessary to manage namespace switching: if `netobserv` is moved to a different namespace, it ensures the previous configmap is correctly going to be deleted.

controllers/flowlogspipeline/flp_monolith_reconciler.go
```diff
@@ -94,18 +95,28 @@ func (r *flpMonolithReconciler) reconcile(ctx context.Context, desired *flowslat
...
+			if err := r.CreateOwned(ctx, dbConfigMap); err != nil {
+				return err
+			}
+		}
 	} else if !equality.Semantic.DeepDerivative(newCM.Data, r.owned.configMap.Data) {
 		if err := r.UpdateOwned(ctx, r.owned.configMap, newCM); err != nil {
 			return err
 		}
+		if r.reconcilersCommonInfo.availableAPIs.HasConsoleConfig() {
+			if err := r.UpdateOwned(ctx, r.owned.dbConfigMap, dbConfigMap); err != nil {
```
jotak, line 116: not sure about that ... basically it is saying : "if the main configMap is changed, then update the DB config map" ?
I would rather copy-paste the whole if/else block and adapt it for dbConfiogMap

jotak: (same applies to `flp_transfo_reconciler` file)

controllers/flowlogspipeline/flp_monolith_reconciler.go
jotak: (same applies to `flp_transfo_reconciler` file)

jotak:
Thanks for the PR
Some remarks on the reconcilers & objects lifecycle, but other than that looks good!

.mk/ocp.mk
```diff
@@ -45,7 +45,10 @@ undeploy-operator: ## stop the operator locally
 	kubectl delete servicemonitor flowlogs-pipeline-monitor || true
 	kubectl delete service netobserv-plugin || true
 	kubectl delete deployment netobserv-plugin || true
-	kubectl delete servicemonitor netobserv-console-plugin || true
+	kubectl delete servicemonitor netobserv-plugin || true
+	kubectl delete configmap -n openshift-config-managed flowlogs-pipeline-metrics-dashboard || true
+	kubectl delete configmap -n netobserv flowlogs-pipeline-config || true
+	kubectl delete configmap -n netobserv console-plugin-config || true
```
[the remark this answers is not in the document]
jotak: (you just need to make sure that the operator is still running while doing `make undeploy-sample-cr`, else it won't work)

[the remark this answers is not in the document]
KalmanMeth: `make undeploy-operator` is very convenient to clean up all the allocated items in the case we are debugging the operator and something goes wrong.

controllers/flowlogspipeline/flp_monolith_reconciler.go
KalmanMeth: Yes, that is what I thought at first. But then I ran into an issue that this new config map is not really in the `netobserv` namespace. It technically is in the `openshift-config-managed` namespace, but we have to set it up when we create all the objects in the `netobserv` namespace. When the `netobserv` namespace is moved to a different name, it should not affect the dashboard objects.

controllers/flowlogspipeline/flp_monolith_reconciler.go
KalmanMeth: If the main configMap is changed, it is possible that the labels in the dashboard are changed because of a new prefix to the prometheus variables. Otherwise, there is no reason to track the changes to the new dashboards configMap, and as noted above, the dashboard configMap does not technically belong to the netobserv` namespace.

controllers/flowlogspipeline/flp_monolith_reconciler.go
jotak: Oh I didn't notice that, thanks for the explaination

controllers/flowlogspipeline/flp_monolith_reconciler.go
jotak: yeah, but tying the `dbConfigMap` lifecycle to the presence of `configMap` seems risky to me

For instance, imagine that thr process runs through the top `if !r.nobjMngr.Exists(r.owned.configMap)` at first creation, and `configMap` is created, so far so good. Then imagine there's an issue (e.g. a temporary connectivity issue) and the subsequent call to `r.CreateOwned(ctx, dbConfigMap)` fails: in that case,  `configMap` exists and `dbConfigMap`. In such scenario, the operator will never be able to recover from that state, as the reconcile doesn't create `dbConfigMap` when `configMap` exists.

KalmanMeth: The worst that happens is that there is no dashbaord, but everything else runs properly.

To manage it independently from the original configMap, we have to track it. We cannot put it in the `netobserv` nobjMngr structure because it is not in the `netobserv` namespace and it would cause other confusion. So we would need another place to save the status of this particular dbConfigMap. Where would you suggest?

jotak: Yes, I agree. Something like this should work I guess:

```go
func (r *reconcilersCommonInfo) reconcileDbConfig(ctx context.Context, dbConfigMap *corev1.ConfigMap) error {
	curr := &corev1.ConfigMap{}
	if err := r.Get(ctx, types.NamespacedName{
		Name:      dbConfigMap.Name,
		Namespace: dbConfigMap.Namespace,
	}, curr); err != nil {
		if errors.IsNotFound(err) {
			return r.CreateOwned(ctx, dbConfigMap)
		}
		return err
	}
	if !equality.Semantic.DeepDerivative(dbConfigMap.Data, curr.Data) {
		return r.UpdateOwned(ctx, curr, dbConfigMap)
	}
	return nil
}
```

(Same function can be shared between monolith & transfo reconcilers)
It makes also the `dbConfigMap:    &corev1.ConfigMap{}` in `owned` struct useless

KalmanMeth: done

jotak (dismissed):
Thanks, lgtm!

KalmanMeth:
rebased

KalmanMeth:
reintegrated with latest code base

jotak:
I tested the PR, it looks good, there's still a few things to do but they can be addressed in a follow-up:

1. When all metrics are disabled, the dashboard is still present (and empty). It should be removed in that case.
2. We need to tune the defaults so that untested/undocumented dashboards (wrt QE/doc tasks) are removed for the moment
3. Name & flags look a bit weird compared to others. I think we must be missing some guidelines here (to check with the monitoring team)

Capture d’écran du 2023-03-09 17-00-13

jotak:
/approve
/lgtm
thanks @KalmanMeth !

PR #262, opened 2023-02-03
NETOBSERV-856 Fix infinite reconcile & add more debugging utility

jotak:
The infinite reconcile was due to a bad change comparison of liveness/readiness probes.

This commit is also adding better reconciliation change tracking utility for debugging.


jotak:
@OlivierCazade @jpinsonneau I'm adding here more logs that help a lot debugging - but there's a drawback, that it is doing more logging stuff at runtime.
Actually I don't find this logger (`sigs.k8s.io/controller-runtime/pkg/log`) very nice to use (or maybe that's me who don't understand it well). For instance I would like more fine grained log level but it only offers Info and Error. Also, it doesn't seem possible to do something like `log.IsInfoEnabled` to avoid unnecessary processing when log level is Error.

Maybe we should replace it with logrus, that we use in the other repos, wdyt? (in a next PR)

jpinsonneau (approved):
LGTM. I'm fine with the idea of changing the logger in a followup.

FYI in `loki-operator` they use https://github.com/ViaQ/logerr

jpinsonneau (approved):

PR #271, opened 2023-02-16
WIP: Use regular deploy make commands for OCP local runs

msherif1234:
by default webhook is enabled but when run locally we can't have tls.cert local mounts so only disable webhook when run the operator locally used only for development.

Error seen during local operator deployment
```shell
2023-02-16T09:12:42.227+0200    ERROR   setup   problem running manager {"error": "open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"}
```

### DIFF
--- a/.mk/ocp.mk
+++ b/.mk/ocp.mk
@@ -31,21 +31,17 @@ ocp-run: ocp-undeploy ocp-deploy ocp-deploy-operator  ## OCP-deploy + run the op
 ocp-deploy-operator: ## run flp from the operator
 	@echo "====> Enable netobserv-plugin in OCP console"
 	oc patch console.operator.openshift.io cluster --type='json' -p '[{"op": "add", "path": "/spec/plugins", "value": ["netobserv-plugin"]}]'
-	@echo "====> Running the operator locally"
-	go run ./main.go \
-		-ebpf-agent-image=quay.io/netobserv/netobserv-ebpf-agent:main \
-		-flowlogs-pipeline-image=quay.io/netobserv/flowlogs-pipeline:main \
-		-console-plugin-image=quay.io/netobserv/network-observability-console-plugin:main
+	$(MAKE) deploy
 
-.PHONY: undeploy-operator
-undeploy-operator: ## stop the operator locally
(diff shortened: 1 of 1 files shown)


jotak:
@KalmanMeth are these changes good to you?
I think you've been using the `ocp-deploy` make target differently than the common `deploy` target because you wanted to have the operator running locally on your machine, right? IMO with the suggested changes, the `ocp-deploy / undeploy` targets become more redundant with the usual `deploy/undeploy` and maybe can be removed?

Or do we need to keep a way to run the operator locally?

msherif1234:
> @KalmanMeth are these changes good to you? I think you've been using the `ocp-deploy` make target differently than the common `deploy` target because you wanted to have the operator running locally on your machine, right? IMO with the suggested changes, the `ocp-deploy / undeploy` targets become more redundant with the usual `deploy/undeploy` and maybe can be removed?
> 
> Or do we need to keep a way to run the operator locally?

I agree wuth @jotak since we already use cluster when run locally it doesn't save us much and add overhead as we add more functionality to the operator if in agreement we can remove ocp deploy options and just use regular deploy/undeploy

KalmanMeth:
I find it much easier to iteratively debug the operator if I run it locally, so that I don't have to keep making container images, pushing it, etc, on each debug iteration. 

I tried again to run:
```
make build
make install
make ocp-deploy
make ocp-deploy-operator
```
and flp is not run from the operator.

jotak:
I think the "ocp" make target names are misleading, perhaps part of the problem is that it's not clear what they do:

- `deploy-all` does actually not deploy the operator; but it deploys a test-ready environment with loki, grafana, a sample CR, a sample workload ... Maybe it should be "deploy-test-environment" ?

- `ocp-deploy-operator` could probably be removed as it overlaps with `run`. `run` must be fixed if necessary, as it may be broken today. Btw the `oc patch` thing that ` make ocp-deploy-operator` is doing should not be necessary (that's automated by the operator itself).

- The `ocp-*` targets really don't do anything very special afaict, except they expose services via routes. Maybe, instead of having all these variants, we could just have keep the `ocp-expose` and we would call it explicitly rather than implicitly. It would help to shrink a little bit these makefiles.

PR #272, opened 2023-02-17
Fix doc for operator-sdk command & make generate now must be explicit

jotak:
When using makefile commands, operator-sdk now is downloaded to bin; not all docs were updated when this change was made, fixing it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,15 +112,13 @@ help: ## Display this help.
 # Directories.
 
 ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-BIN_DIR := bin
-TOOLS_DIR := hack/tools
-TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
+BIN_DIR := $(abspath bin)
 GO_INSTALL := ./scripts/go_install.sh
 CONVERSION_GEN_VER := v0.25.0
 CONVERSION_GEN_BIN := conversion-gen
 # We are intentionally using the binary without version suffix, to avoid the version
 # in generated files.
-CONVERSION_GEN := $(abspath $(TOOLS_BIN_DIR)/$(CONVERSION_GEN_BIN))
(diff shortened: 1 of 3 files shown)


jotak:
/hold
To make it correctly I also need to change the Makefile so that it doesn't rely on operator-sdk in path

Makefile
```diff
@@ -111,15 +111,13 @@ help: ## Display this help.
 # Directories.
 
 ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-BIN_DIR := bin
-TOOLS_DIR := hack/tools
-TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
+BIN_DIR := $(abspath bin)
```
jotak, line 114: cc @msherif1234 we already had all our executable tools in `./bin` (e.g. operator-sdk ...) , so this is moving `conversion-gen` to the same place to avoid too much fragmentation

Makefile
```diff
@@ -111,15 +111,13 @@ help: ## Display this help.
...
 GO_INSTALL := ./scripts/go_install.sh
 CONVERSION_GEN_VER := v0.25.0
 CONVERSION_GEN_BIN := conversion-gen
 # We are intentionally using the binary without version suffix, to avoid the version
 # in generated files.
-CONVERSION_GEN := $(abspath $(TOOLS_BIN_DIR)/$(CONVERSION_GEN_BIN))
+CONVERSION_GEN := $(abspath $(BIN_DIR)/$(CONVERSION_GEN_BIN))
```
stleerh, line 120: BIN\_DIR is already made absolute on line 114 so no need to do it again here.  
`CONVERSION_GEN := $(BIN_DIR)/$(CONVERSION_GEN_BIN)`

Makefile
msherif1234: Sure make sense

Makefile
jotak: done

jotak:
thanks,
/approve

PR #273, opened 2023-02-17
Enable operator metrics collection

OlivierCazade:
Enable collection for operator metrics.

### DIFF
--- a/config/openshift-olm/monitor.yaml
+++ b/config/openshift-olm/monitor.yaml
@@ -1,4 +1,3 @@
-
 # Prometheus Monitor Service (Metrics)
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
@@ -9,11 +8,14 @@ metadata:
   namespace: system
 spec:
   endpoints:
-    - path: /metrics
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      path: /metrics
+      interval: 30s
       port: https
       scheme: https
       tlsConfig:
-        insecureSkipVerify: true
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        serverName: netobserv-metrics-service.openshift-operators.svc
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -12,7 +12,6 @@ spec:
     - path: /metrics
       port: https
       scheme: https
-      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
       tlsConfig:
         insecureSkipVerify: true
   selector:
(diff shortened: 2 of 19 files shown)


config/prometheus/monitor.yaml
```diff
@@ -12,7 +12,6 @@ spec:
     - path: /metrics
       port: https
       scheme: https
```
jotak, line 14: What is missing for having a TLS without skipping verify?
It's basically the same reason why we have this bug: https://issues.redhat.com/browse/NETOBSERV-765

OlivierCazade: Similar to what we did for FLP we need to add the labels to the metric services to generate certificate and mount them in the kube-rbac-proxy container so they can be used.

OlivierCazade:
Added SSL and auth configuration for downstream collection, this made operator metrics endpoint incompatible with monitoring user workload.

@memodi appart from review comments, the PR is ready to be tested.
After deploying this branch and setting DOWNSTREAM_DEPLOYMENT to true, you should see operator metrics without enabling user workload.

image

memodi:
> Added SSL and auth configuration for downstream collection, this made operator metrics endpoint incompatible with monitoring user workload.

@OlivierCazade - is the idea, they would be compatible with SSL and auth configuration once operator is deployed with DOWNSTREAM_DEPLOYMENT to true?

OlivierCazade:
Yes, SSL and authentication for the metrics endpoint. And the prometheus deployment should automatically use it from the ServiceMonitor configuration

memodi:
@OlivierCazade - I tried with image from this PR, I am not able to see those metrics:
```
$ oc get pod/netobserv-controller-manager-7794cf659b-vcnhn -o yaml | egrep -A 1 "DOWN"
    - --downstream-deployment=$(DOWNSTREAM_DEPLOYMENT)
    command:
--
    - name: DOWNSTREAM_DEPLOYMENT
      value: "true"

$ oc get servicemonitor
NAME                        AGE
netobserv-metrics-monitor   5m19s
```
image

am I missing something?

OlivierCazade:
Most of the changes in this PR are at the bundle level, did you also used the bundle from this branch?

monitor.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
jotak, line 1: I don't think we want this file, do we? Is it included in the bundles, or is it just there as a helper (e.g. for make targets) ? afaict it should not be included in bundles

config/openshift-olm/monitor.yaml
```diff
@@ -9,12 +8,14 @@ metadata:
...
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      path: /metrics
+      interval: 30s
       port: https
       scheme: https
-      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
       tlsConfig:
-        insecureSkipVerify: true
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        serverName: netobserv-metrics-service.netobserv.svc
```
jotak, line 18: so, my understanding is that we should hardcode here the installation namespace, `openshift-operators`, right?

monitor.yaml
OlivierCazade: No sorry, this was one of my test file.

jotak (approved):

memodi:
adding NETOBSERV-900 as JIRA reference

PR #277, opened 2023-02-21
Fix deployment on non-ovnk (e.g. kind)

jotak:
Exit silently when ovn-k workloads aren't found, when they are not required

### DIFF
--- a/controllers/ovs/flowsconfig_ovnk_reconciler.go
+++ b/controllers/ovs/flowsconfig_ovnk_reconciler.go
@@ -56,7 +56,7 @@ func (c *FlowsConfigOVNKController) updateEnv(ctx context.Context, target *flows
 		Name:      c.config.DaemonSetName,
 		Namespace: c.config.Namespace,
 	}, ds); err != nil {
-		if kerr.IsNotFound(err) && !target.Spec.UseIPFIX() {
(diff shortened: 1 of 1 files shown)


jotak:
This bug was seen while testing the bundle on KIND (for upstream release). As a result I did not release 1.0.2 on non-openshift operatorhub

controllers/ovs/flowsconfig_ovnk_reconciler.go
```diff
@@ -50,9 +51,16 @@ func (c *FlowsConfigOVNKController) Reconcile(
...
-	if err != nil {
-		return err
+	ds := &appsv1.DaemonSet{}
+	if err := c.client.Get(ctx, types.NamespacedName{
+		Name:      c.config.DaemonSetName,
+		Namespace: c.config.Namespace,
+	}, ds); err != nil {
+		if kerr.IsNotFound(err) && !target.Spec.UseIPFIX() {
+			// If we don't want IPFIX and ovn-k daemonset is not found, assume there no ovn-k, just succeed silently
+			return nil
```
msherif1234, line 61: can we add a log info message here just stating non ovn-kubernetes mode ?

jotak: sure

jotak:
Thanks @msherif1234 
/approve

PR #279, opened 2023-02-27
NETOBSERV-906 convert and post-process flows spec

jotak:
That's the second step after https://github.com/netobserv/network-observability-console-plugin/pull/296

Once the latter is merged, perhaps we can change it so it pulls the markdown with curl/wget

@skrthomas the asciidoc output is `docs/flows-format.adoc`


OlivierCazade (dismissed):
LGTM, +1 for automatic pull

docs/flows-format.adoc
```diff
@@ -0,0 +1,250 @@
+// Automatically generated by 'hack/asciidoc-flows-gen.sh'. Do not edit.
+:_content-type: REFERENCE
+[id="network-observability-flows-format-specifications_{context}"]
+= Network Flows format specifications
+
+Description::
++
+This is the specification of the network flows format, used both internally and in exporter feature such as Kafka export.
```
skrthomas, line 8:
```suggestion
This is the specification of the network flows format, used both internally and when exporting flows to Kafka.
```

docs/flows-format.adoc
```diff
@@ -0,0 +1,250 @@
...
++
+The document is organized in two main categories: _Labels_ and regular _Fields_. This distinction only matters when querying Loki, as _Labels_, unlike _Fields_, must be used in link:https://grafana.com/docs/loki/latest/logql/log_queries/#log-stream-selector[stream selectors].
```
skrthomas, line 10:
```suggestion
The document is organized in two main categories: _Labels_ and regular _Fields_. This distinction only matters when querying Loki, as _Labels_, unlike _Fields_, which must be used in link:https://grafana.com/docs/loki/latest/logql/log_queries/#log-stream-selector[stream selectors].
```

docs/flows-format.adoc
```diff
@@ -0,0 +1,250 @@
...
++
+Outside of Loki context, for instance when reading this specification as a documentation for the Kafka export feature, you must ignore this distinction about _Labels_ and _Fields_, and consider all of them at the same level in the JSON structure.
```
skrthomas, line 12:
```suggestion
When using this specification as a reference for exporting flows to Kafka, you must ignore this distinction about _Labels_ and _Fields_, and consider all of them at the same level in the JSON structure.
```

skrthomas (commented):
A few edits but lgtm other than those small things.

docs/flows-format.adoc
skrthomas: Also one question, when you refer to using internally, are you referring to using to query Loki?

docs/flows-format.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+
+== Labels
+
+
+=== SrcK8S_Namespace
```
skrthomas, line 18: I'm getting asciibinder errors with these H3's. I was wondering about making these list items, as you can see here: https://github.com/openshift/openshift-docs/pull/56335/files

wdyt?

jotak: Sure
Do you see anything else that should be changed? The format is dictated by `typedoc`, so I can't do absolutely anything I want without too much pain, but some little things can certainly be adapted. For instance, I was wondering if the line separators `'''` were really useful or if we should remove them.. wdyt?

docs/flows-format.adoc
jotak: I'm not sure of that one? In its second part, the sentence is already with subject=Labels and verb=must so I don't think "which" would be correct here? (It could be reformulated as: "This distinction only matters when querying Loki. This is because _Labels_, unlike _Fields_, must be used in etc.". )

skrthomas: Okay, I think your reworking, "This distinction only matters when querying Loki. This is because Labels, unlike Fields, must be used in etc." is good, lets go with that.

docs/flows-format.adoc
jotak: When looking in my editor renderer, the "+" seem important to make all the description appear at the same level of indentation

skrthomas: Ah, I see. Is the 'Description::' a necessary element for your build? I think we can also do without the 'Description::' and leave the text as introductory text that follows the H1 "Network Flows format specifications". 

Another thing to mention about the Description is that I want to relocate this sentence: "This is the specification of the network flows format, used both internally and in exporter feature such as Kafka export" so that it is introductory text for the Assembly file. Like this: 
image

jotak: As discussed, I have changed the generated file to make it an ASSEMBLY file

OlivierCazade (approved):

PR #281, opened 2023-02-28
Allow modifying namespace from web console

msherif1234:
image


msherif1234:
@jotak I think there is still a value allowing custom NS instead of just having the default `openshift-operators` NS

jotak:
iirc `openshift-*` namespaces are reserved for in-payload operators

msherif1234:
> iirc, new `openshift-*` namespaces are reserved for in-payload operators, but I'm not totally sure The OLM team might know, but perhaps you already asked them ?

I didn't see any such limitation here https://docs.openshift.com/container-platform/4.8/operators/operator_sdk/osdk-generating-csvs.html#osdk-suggested-namespace_osdk-generating-csvs

jotak (changes requested):
cf previous comment, I have a doubt that this change is actually desirable

PR #283, opened 2023-03-01
NETOBSERV-773 Copy certificates across namespaces

jotak:
Certificates (configmaps or secrets) can now be referred to with their source namespace; when that namespace doesn't match their target namespace (e.g. netobserv or netobserv-privileged), they are copied.

Had to refactor a little bit to break an import cycle between cert watchers and helper

Add tests


jotak:
@jpinsonneau this is what we talked about today on copying certificates. I need to rebase and retest, seems like it's full of conflicts 😬

jotak:
I'll close and reopen a new one

PR #284, opened 2023-03-02
KIND: Fix output path for api conversation generation + Fix webhook ca bundle injection for flow-collector CRD

astoycos:
This PR fixes two bugs:  

1. 
There are still some webhook issues when running on KIND... specifically I spin things up with 

1. `IMG="quay.io/bpfd/network-observability-operator:test" make image-build image-push`
2. `make local-deploy`
3. `IMG="quay.io/bpfd/network-observability-operator:test" make deploy-kind` 
4. `make deploy-sample-cr` 

I run into 
```
==> Deploy sample CR
kubectl apply -f ./config/samples/flows_v1beta1_flowcollector.yaml || true
Error from server: error when creating "./config/samples/flows_v1beta1_flowcollector.yaml": conversion webhook for flows.netobserv.io/v1beta1, Kind=FlowCollector failed: Post "https://netobserv-webhook-service.netobserv.svc:443/convert?timeout=30s": x509: certificate signed by unknown authority
```

operator Logs: 

```
2023/03/02 06:33:07 http: TLS handshake error from 10.244.0.1:3302: remote error: tls: bad certificate
```

This PR fixes the aforementioned bug by ensuring we inject the webhook ca-bundle into the flow-collector CRD definition. 

2. 
The makefile logic to detect weather or not the cloned code is in gopath is flawed....specifically 

```
# Set --output-base for conversion-gen if we are not within GOPATH
ifneq ($(abspath $(ROOT_DIR)),$(shell go env GOPATH)/src/network-observability-operator)
	CONVERSION_GEN_OUTPUT_BASE := --output-base=$(ROOT_DIR)
else
	export GOPATH := $(shell go env GOPATH)
endif
```
Assumes that you specifically cloned the code to `$GOPATH/src/network-observability-operator` which probably isn’t the case for everyone

for example on my setup the repo lives at `/home/astoycos/go/src/github.com/netobserv/network-observability-operator` so although it’s in my GOPATH we still set --output-base here when we don’t need to.  Ultimately I was seeing `zz_generated.conversion.go` get put in the wrong spot because of this

The fix just makes sure that the $GOPATH is in wherever the code lives (i.e $ROOT_DIR)


msherif1234:
@astoycos pls make sure u update to main latest all webhook issues should have been addressed as well as kind issues

Makefile
```diff
@@ -217,7 +217,8 @@ generate-go-conversions: $(CONVERSION_GEN) ## Run all generate-go-conversions
 		$(CONVERSION_GEN) \
 		--input-dirs=./api/v1alpha1 \
 		--build-tag=ignore_autogenerated_core \
-		--output-file-base=zz_generated.conversion $(CONVERSION_GEN_OUTPUT_BASE)\
```
msherif1234, line 220: nit why split the line seems unneeded change ?

astoycos:
/assign @msherif1234 

This should be good to go now!

config/kubernetes/manager_webhook_patch.yaml
```diff
@@ -21,3 +21,11 @@ spec:
         secret:
           defaultMode: 420
           secretName: webhook-server-cert
+---
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
msherif1234, line 25: isn't better/cleaner to have this in its own patch file and let `kustomization` patch it ?

msherif1234 (commented):
overall looks good thanks for fixing KinD env, just left a couple of nits

msherif1234:
also there is no longer version updates so pls adjust the title and PR description just indicating for kind we need to have cert-mage CA or something similar

Makefile
astoycos: Just code cleanliness tbh :)

config/kubernetes/manager_webhook_patch.yaml
astoycos: Sure, it's just a preference thing I can do that if you'd like, I wish I could just point to `../crd/patches/cainjection_in_flowcollectors.yaml` but it won't let me point to that file from the `kubernetes/` directory

msherif1234: just wanted to keep the original sdk yamls as close as the original auto generated ones so any future interactions with sdk-operator will be smooth

msherif1234: I can do that later if needed

astoycos: All good I can do it now

astoycos: Done

msherif1234: Thanks!
/lgtm

jotak:
tested and lgtm, thanks @astoycos !
/approve

PR #285, opened 2023-03-02
NETOBSERV-847 Operator multi-arch builds (upstream)

jpinsonneau:
Implements multi arch builds commands
Also linked push commands to their related build


Makefile
```diff
@@ -255,21 +253,50 @@ coverage-report-html:
...
 else
 	$(OCI_BIN) push ${IMG}
 endif
 
+single-multiarch-linux-build/%:
+#The --load option is ignored by podman but required for docker
+	DOCKER_BUILDKIT=1 $(OCI_BIN) buildx build --load --build-arg TARGETPLATFORM=linux/$* --build-arg TARGETARCH=$* --build-arg BUILDPLATFORM=linux/amd64 -t ${IMG}-$* -f Dockerfile .
+
+single-multiarch-linux-push/%: single-multiarch-linux-build/%
+	DOCKER_BUILDKIT=1 $(OCI_BIN) push ${IMG}-$*
```
msherif1234, line 273: I think we should keep build and push separate and if needed we can have target to do both

Makefile
```diff
@@ -255,21 +253,50 @@ coverage-report-html:
...
-	$(OCI_BIN) build --build-arg BASE_IMAGE=$(IMG) -t $(IMG_SHA) -f ./shortlived.Dockerfile .
-
 image-push: ## Push OCI image with the manager.
 ifneq (,$(findstring quay.io/netobserv/,$(IMG)))
 	$(error Do not push to quay.io/netobserv)
 else
 	$(OCI_BIN) push ${IMG}
 endif
 
+single-multiarch-linux-build/%:
```
msherif1234, line 268: pls add to all new target ## comment style so they show in `make help`

jpinsonneau:
Should we add labels as [described here](https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/#docs) ?

Supported Labels
The following label convention defines the target OSes and architectures supported by an operator:
```
labels:
    operatorframework.io/arch.<GOARCH>: supported
    operatorframework.io/os.<GOOS>: supported
Where <GOARCH> and <GOOS> are one of the values [listed here](https://github.com/golang/go/blob/master/src/go/build/syslist.go).
```
Multiple Architectures
Some operators may support multiple node architectures or OSes. In this case, multiple labels can be added. For example, an operator that support both windows and linux workloads will sport the following labels:
```
labels:
    operatorframework.io/os.windows: supported
    operatorframework.io/os.linux: supported
```

msherif1234:
why we didn't pass in the new build-arg `make image-build` and make the platform env var and user can pass in the arch it needs to build for instead of having the new make targets ?

Makefile
```diff
@@ -75,13 +78,8 @@ GOBIN=$(shell go env GOBIN)
...
 
 # Image building tool (docker / podman)
-ifndef OCI_BIN
-ifeq (,$(shell which podman 2>/dev/null))
-OCI_BIN=docker
-else
-OCI_BIN=podman
-endif
-endif
+OCI_BIN_PATH = $(shell which podman  || which docker)
```
stleerh, line 81: Same comment as:  
https://github.com/netobserv/netobserv-ebpf-agent/pull/100#discussion_r1123719720

Makefile
```diff
@@ -255,21 +253,50 @@ coverage-report-html:
...
 image-push: ## Push OCI image with the manager.
 ifneq (,$(findstring quay.io/netobserv/,$(IMG)))
 	$(error Do not push to quay.io/netobserv)
 else
 	$(OCI_BIN) push ${IMG}
 endif
 
+single-multiarch-linux-build/%:
+#The --load option is ignored by podman but required for docker
+	DOCKER_BUILDKIT=1 $(OCI_BIN) buildx build --load --build-arg TARGETPLATFORM=linux/$* --build-arg TARGETARCH=$* --build-arg BUILDPLATFORM=linux/amd64 -t ${IMG}-$* -f Dockerfile .
```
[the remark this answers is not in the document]
OlivierCazade: BUILDPLATFORM is the platform used to build, se here we build on amd64 while TARGETPLATFORM is the platform where we want the image to run on.

Here, we are cross building on BUILDPLATFORM an image that is intended to run on TARGETPLATFORM.

Makefile
jpinsonneau: Sure, following your comment, I have refactored all the Makefiles and updated these docs :smile:

jpinsonneau:
Rebased with https://github.com/netobserv/network-observability-operator/pull/300

@jotak @OlivierCazade At the end, [it forces me to separate build / push jobs steps ](https://github.com/netobserv/network-observability-operator/compare/596afabf3e13fc05c3b278900efbbf89e0867cc6..24cfe0bddf3f0d18c7bcbbbacca8248bee8f32a4#diff-da7639ee7485b08c306c58f5c406d1263fe8a0b496322fd6825923fd8ec6b76dR37-R40)and I feel it was better using all at once to ensure arguments are correct.

I wonder if there is a smart way to mock `docker push ...` comands only from `test-workflow.sh` without modifying the Makefile.

PR #287, opened 2023-03-02
NETOBSERV-897 Report coverage

jotak:


jotak:
Note: we should, the engineering team, take some time to review where tests are missing. I don't think we should aim for a precise percentage of coverage: this is known to be counter-productive (this is a very flawed metric to measure quality, however this is a good tool to help analyze coverage lacks. To know more about what test coverage provide and do not provide, refer to Martin Fowler's [TestCoverage](https://martinfowler.com/bliki/TestCoverage.html) or Brian Marick's [How To Misuse Code Coverage](http://www.exampler.com/testing-com/writings/coverage.pdf).

PR #288, opened 2023-03-08
WIP: Bpfd Integration

astoycos:
See https://github.com/redhat-et/bpfd/issues/322 for more information


config/samples/flows_v1beta1_flowcollector.yaml
```diff
@@ -8,23 +8,26 @@ spec:
...
       sampling: 50
       cacheActiveTimeout: 5s
       cacheMaxFlows: 100000
-      interfaces: [ ]
-      excludeInterfaces: [ "lo" ]
-      logLevel: info
+      interfaces: [ "eth0", "veth5871bad5" ]
+      excludeInterfaces: [ ]
+      logLevel: debug
+      privileged: true
```
jotak, line 18: @astoycos the privileged mode here is not something that will be required when running with bpfd, or is it? It would increase the agent privileges requirements (while I guess they should be lowered). Or is it just something you did for testing/debugging convenience but that will not be required ultimately?

config/samples/flows_v1beta1_flowcollector.yaml
```diff
@@ -8,23 +8,26 @@ spec:
...
     ebpf:
-      imagePullPolicy: IfNotPresent
+      imagePullPolicy: Always
       sampling: 50
       cacheActiveTimeout: 5s
       cacheMaxFlows: 100000
-      interfaces: [ ]
-      excludeInterfaces: [ "lo" ]
-      logLevel: info
+      interfaces: [ "eth0", "veth5871bad5" ]
```
jotak, line 15: are we still going to be able to attach to all interfaces? (ie no explicit listing)

astoycos: So that feature is in progress,  See https://github.com/redhat-et/bpfd/issues/322#issuecomment-1460539896 for the specific issue!

config/samples/flows_v1beta1_flowcollector.yaml
astoycos: > Or is it just something you did for testing/debugging convenience but that will not be required ultimately?

Yep :)

jotak:
@astoycos I'm closing this PR, I guess you'll be ok with that?

PR #291, opened 2023-03-09
Fix remaining Makefile issues + annoying bundle date

jotak:
After the recent reworking on the Makefile, there were still some pending issues. E.g. we introduced a "deploy-latest" target but it did not allow to override the operator image.

This is rollbacking part of the last changes:
- deploy-latest is removed, we can run "make deploy" as we used to do before
- make deploy now uses again "main" images by default, as before
- keeping the behaviour that Mohamed introduced, "make bundle" keeps using the provided IMG
- UNLIKE before: running "make bundle" alone will NOT update the bundle so it is ready for commit; For that, a new target "make update-bundle" is introduced
- "make bundle" does not generate the creation date as it used to do. This is now done via "make bundle-build". It means that release scripts have to be updated so they generate the creation date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ IMAGE_TAG_BASE ?= $(REPO)/network-observability-operator
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(BUNDLE_VERSION)
 
 # Image URL to use all building/pushing image targets
-IMG ?= $(IMAGE_TAG_BASE):$(OPERATOR_VERSION)
+IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
 IMG_SHA = $(IMAGE_TAG_BASE):$(BUILD_SHA)
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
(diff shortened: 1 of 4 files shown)


OlivierCazade (dismissed):
LGTM

Makefile
```diff
@@ -327,9 +322,16 @@ bundle: bundle-prepare ## Generate final bundle files.
...
+.PHONY: update-bundle
+update-bundle: IMG=$(IMAGE_TAG_BASE):$(OPERATOR_VERSION)
+update-bundle: bundle ## Prepare a clean bundle to be commited
+
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
+	cp ./bundle/manifests/netobserv-operator.clusterserviceversion.yaml tmp-bundle
+	$(SED) -i -r 's~:created-at:~$(DATE)~' ./bundle/manifests/netobserv-operator.clusterserviceversion.yaml
 	$(OCI_BIN) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	mv tmp-bundle ./bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```
stleerh, line 334: If the Docker build fails (line 333), then it won't move the file back to its original state.  An improvement is to put a leading - so it ignores any error and hence, the next statement will run.  
 `-$(OCI_BIN) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .`

The slight drawback is that it will say the error is ignored.  It's also not a complete solution since aborting with ctrl-c will still be a problem.  The best practice is to modify the copied file instead, but I understand it may not be easily done in this situation.

jotak: I did what you suggested first - I think that's ok, we can live with the little drawback

stleerh: /lgtm

PR #294, opened 2023-03-13
NETOBSERV-918 dashboards follow-up

jotak:
- Delete (or do not create) dashboards CM when it is empty
- By default, do not create FLP metrics & dashboards
- Rename dashboard
- Add some tests


OlivierCazade (dismissed):
LGTM

OlivierCazade (approved):

PR #295, opened 2023-03-14
NETOBSERV-881: Added field to disable alerts creation

OlivierCazade:
Added disable list to not create alerts. Each alert listed in this field would not be created.


api/v1alpha1/flowcollector_webhook.go
```diff
@@ -53,6 +53,10 @@ func (r *FlowCollector) ConvertTo(dstRaw conversion.Hub) error {
 		dst.Spec.Processor.ConnectionEndTimeout = restored.Spec.Processor.ConnectionEndTimeout
 	}
 
+	if restored.Spec.Processor.Metrics.DisableAlerts != nil {
```
jotak, line 56: out of curiosity, did you have "something" telling you that you had something to do here, or was it something you had to keep in mind yourself? (trying to understand how error prone it can be if we forget about the conversion)

api/v1beta1/flowcollector_types.go
```diff
@@ -280,6 +280,10 @@ type FLPMetrics struct {
 	// ignoreTags is a list of tags to specify which metrics to ignore
 	//+kubebuilder:default:={"egress","packets"}
 	IgnoreTags []string `json:"ignoreTags,omitempty"`
+
+	// disableAlerts is a list of alerts related to FLP that should not be created
```
jotak, line 284: I am not sure what is the best is terms of usage here, a deny-list or an allow-list. Thoughts?
In any case we should document here what are the acceptable values, ie `NetObservNoFlows` and `NetObservLokiError` at the moment

api/v1alpha1/flowcollector_webhook.go
OlivierCazade: After adding the new field in v1beta1 running make generate was triggering some warning because of the new fields.

But then I mostly looked at previous PR to understand how to do it.

api/v1beta1/flowcollector_types.go
OlivierCazade: I think as a default behavior we want to enable alerts. Using an allow list would make upgrade with new alert a bit more trickier since we would not know if it is still the default list and if we should add the new alert to the upgrade.

I will improve the documentation to mention the possibilities.

jotak: Also, I would remove the FLP mention, as it might not always be clear to the users when it's related to FLP or not (they don't necessarily know well our architecture)

```suggestion
	// disableAlerts is a list of alerts that should not be created. Possible values are: `NetObservNoFlows`, which is triggered when no flows are being observed for a certain period, and `NetObservLokiError`, which is triggered when flows are being dropped due to Loki errors.
```

jotak: cc @skrthomas could you please also review the CRD doc added here?
Thanks

OlivierCazade: I added your change but also moved to an enum so we can control the potential values.

api/v1beta1/flowcollector_types.go
```diff
@@ -271,6 +271,16 @@ type MetricsServerConfig struct {
 	TLS ServerTLS `json:"tls"`
 }
 
+const (
+	AlertNoFlows   = "NetObservNoFlows"
+	AlertLokiError = "NetObservLokiError"
+)
+
+// Name of a processor alert
+// Possible values are: `NetObservNoFlows`, which is triggered when no flows are being observed for a certain period, and `NetObservLokiError`, which is triggered when flows are being dropped due to Loki errors.disableAlerts is a list of alerts related to FLP that should not be created
```
skrthomas, line 280:
```suggestion
// Possible values are: `NetObservNoFlows`, which is triggered when no flows are being observed for a certain period, and `NetObservLokiError`, which is triggered when flows are being dropped due to Loki errors. disableAlerts is a list of alerts related to FLP that should not be created
```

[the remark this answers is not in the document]
skrthomas: I agree. The bit "related to FLP" could be unpacked a little. Could it be a list of alerts related to the health of flows in the FLP? 🤔 Also , a space is needed after the period and before disableAlerts.

api/v1beta1/flowcollector_types.go
```diff
@@ -271,6 +271,16 @@ type MetricsServerConfig struct {
 	TLS ServerTLS `json:"tls"`
 }
 
+const (
+	AlertNoFlows   = "NetObservNoFlows"
+	AlertLokiError = "NetObservLokiError"
+)
+
+// Name of a processor alert
```
skrthomas, line 279:
```suggestion
// Name of a processor alert.
```

skrthomas (commented):
I left a few comments. Mostly a couple of punctuation suggestions and also wondering about the same phrase Joel was asking about.

jotak (approved):
lgtm

jotak:
@skrthomas since there is both this PR and also [Julien's one](https://github.com/netobserv/network-observability-operator/pull/297) that require regenerating the doc, what I can do is once they are merged I'll open a new one with just the doc update to have all at once

skrthomas (approved):
Sounds good, @jotak. lgtm!

PR #302, opened 2023-03-16
NETOBSERV-927 Improve servicemonitor/prometheusrule checking changes …

jotak:
…perf

Only check for spec change, not meta
NB: namespace or name changes follow a different process; it is safe to not check for name/namespace changes there; you cannot anyway update a resource for a namespace or name change


jotak:
@OlivierCazade when you have time to review, this should be a small perf improvement on the operator side

OlivierCazade (commented):
How about labels in meta?

jotak:
@OlivierCazade done - also added labels to service monitors

OlivierCazade (approved):

PR #303, opened 2023-03-20
NETOBSERV-928 Connection tracking renaming

jpinsonneau:
`outputRecordTypes` -> `logTypes`
`connection` -> `conversation`

```yaml
spec:
...
  processor:
...
    logTypes: FLOWS
    conversationEndTimeout: 10s
```

Let's merge this first for consistency (at user level) and see if we do the renaming in FLP component in a followup since there are lots of changes

### DIFF
--- a/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
+++ b/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
@@ -3460,15 +3460,15 @@ spec:
                   receives the flows from the agent, enriches them, and forwards them
                   to the Loki persistence layer.
                 properties:
-                  connectionEndTimeout:
+                  conversationEndTimeout:
                     default: 10s
-                    description: connection end timeout is the duration of time to
-                      wait from the last flow log to end a connection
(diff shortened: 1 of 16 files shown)


bundle/manifests/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -4078,6 +4078,19 @@ spec:
                     - fatal
                     - panic
                     type: string
+                  logTypes:
+                    default: FLOWS
+                    description: logTypes defines the desired record types to generate.
+                      Possible values are "FLOWS" (default) to export flowLogs, "CONVERSATIONS"
+                      to generate newConnection, heartbeat, endConnection events,
```
Amoghrd, line 4085: should it be `Conversation` here too?

jpinsonneau: without doing the FLP renaming, we will have to keep these events

let's secure this first and see how long it will take to merge FLP renaming ?

PR #304, opened 2023-03-22
Update CRD documentation

jotak:
- switch asciidoc gen to referencing v1beta1
- improve doc & default for ignored metrics (those changes were previously done on v1alpha1 instead of v1beta1, mistakenly)
- add enum info on disableAlerts doc (unfortunately the enum doc, while being correctly picked up for openapi, is not picked up for doc generation; so I am duplicating it)
- regenerate bundle & doc

### DIFF
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -289,11 +289,15 @@ type FLPMetrics struct {
 	// +optional
 	Server MetricsServerConfig `json:"server,omitempty"`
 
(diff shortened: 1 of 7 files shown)


jotak:
@skrthomas here's the CRD doc update, with the re-generated asciidoc file

@OlivierCazade I noticed that your PR here did not show the `disableAlerts` enum documentation .. whereas is works correctly with openapi, it doesn't with our doc generators. So I am duplicating the enum listing & doc so it can appear in doc

OlivierCazade (approved):
LGTM

PR #310, opened 2023-03-28
NETOBSERV-765 Add CA/Secret config to ServiceMonitor

jotak:
Add missing CA/Secret to the console plugin service monitor

Cf https://github.com/netobserv/network-observability-console-plugin/pull/315

We might also want to add a bearer token config

### DIFF
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -540,7 +540,7 @@ type FlowCollectorConsolePlugin struct {
 	//+kubebuilder:validation:Minimum=1
 	//+kubebuilder:validation:Maximum=65535
 	//+kubebuilder:default:=9001
-	// port is the plugin service port
+	// port is the plugin service port. Do not use 9002, which is reserved for metrics.
 	Port int32 `json:"port,omitempty"`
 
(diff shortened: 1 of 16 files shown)


OlivierCazade (dismissed):
LGTM

api/v1beta1/flowcollector_types.go
```diff
@@ -532,7 +532,7 @@ type FlowCollectorConsolePlugin struct {
 	//+kubebuilder:validation:Minimum=1
 	//+kubebuilder:validation:Maximum=65535
 	//+kubebuilder:default:=9001
-	// port is the plugin service port
+	// port is the plugin service port. Do not use 9002, which is reserved for metrics.
```
jotak, line 535: Note: I didn't want to add a new port config to the CRD, I think it brings very little value if any, while bloating the CRD with new configs ...

OlivierCazade (dismissed):

jpinsonneau (approved):

OlivierCazade (approved):
LGTM

api/v1beta1/flowcollector_types.go
memodi: @jotak while we're at it, do you think it would be good idea to address this: https://issues.redhat.com/browse/NETOBSERV-511 ?

PR #312, opened 2023-03-30
NETOBSERV-963 revert most of cert watching

jotak:
Reverting most of certificate watching ([NETOBSERV-684](https://issues.redhat.com//browse/NETOBSERV-684)) as it generates a lot of pods restart.

We did not necessarily have to do this certificate watching as CM/secrets are updated within volumes.

We might however monitor carefully if new (or old) problems arise, potentially due to the kubelet sync delay for updating volumes

And also make sure certificates aren't cached in our different workloads

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -18,7 +18,6 @@ import (
 	flowslatest "github.com/netobserv/network-observability-operator/api/v1beta1"
 	"github.com/netobserv/network-observability-operator/controllers/constants"
 	"github.com/netobserv/network-observability-operator/pkg/helper"
-	"github.com/netobserv/network-observability-operator/pkg/watchers"
 )
 
 const secretName = "console-serving-cert"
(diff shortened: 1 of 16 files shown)


jotak:
> do u know of a way to force cert rotation ?

@msherif1234 no, apart from deleting / recreating the LokiStack, which would generate a new certificate

jotak:
> @msherif1234 no, apart from deleting / recreating the LokiStack, which would generate a new certificate

Actually this is wrong: while deleting + recreating LokiStack does delete and recreate the certificate CM, this is the same certificate that is inside. I didn't notice that the first time I tried.

jpinsonneau:
> @stleerh if there are any remaining cached data, that wouldn't be in the operator, but in flp/plugin/ebpf.

Yes I guess we need to revisit eBPF agent kafka certificate reading that seems to be done at startup only:
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/cmd/netobserv-ebpf-agent.go#L42
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/agent/agent.go#L211

OlivierCazade (approved):
LGTM

jpinsonneau (approved):

PR #313, opened 2023-03-30
NETOBSERV-915 add conversation direction

jpinsonneau:
This PR adds direction fields to conversation tracking

~~Also it disable TCP flags that generates duplicated events in Pod to Pod conversations. That may be related to out of order flows~~

Related PRs:
- https://github.com/netobserv/network-observability-console-plugin/pull/317 
- https://github.com/netobserv/flowlogs-pipeline/pull/413
- https://github.com/netobserv/flowlogs-pipeline/pull/430


controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -415,8 +427,9 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
 				},
 			},
 			TCPFlags: api.ConnTrackTCPFlags{
-				FieldName:           "Flags",
-				DetectEndConnection: true,
+				FieldName: "Flags",
+				//TODO: fix TCP detection that generates duplicated events
+				DetectEndConnection: false,
```
jpinsonneau, line 432: @ronensc as discussed, this fix the duplicates when using direction. That seems to confirm your out of order theory :thinking:

jpinsonneau:
updated operation names https://github.com/netobserv/network-observability-operator/compare/34061af2b98ecc5371c41526939e7637675f71e4..a659f27b573a14ce30f35d4f06b7edeffe8a88ff

jpinsonneau:
rebased https://github.com/netobserv/network-observability-operator/compare/5b14c6c542de1d6dcf4377d3cf54fe630709c7ba..3ce1b5bacf6a25927ba2fd1c39e95e563b14a651

api/v1beta1/flowcollector_types.go
```diff
@@ -417,6 +417,11 @@ type FlowCollectorFLP struct {
 	// conversation end timeout is the duration of time to wait from the last flow log to end a conversation
 	ConversationEndTimeout *metav1.Duration `json:"conversationEndTimeout,omitempty"`
 
+	//+kubebuilder:default:="5s"
+	// +optional
+	// conversation terminating timeout is the duration of time to wait from detected FIN flag to end a connection
+	ConversationTerminatingTimeout *metav1.Duration `json:"conversationTerminatingTimeout,omitempty"`
```
jotak, line 423: I have a mixed opinion on this setting, while I get that it might be useful to fix conn track issues that we can expect (e.g. if there's an overall high latency in flow collection), I also think it makes conn-tracking sound more complex from a lambda user PoV ... and also because the name of this setting is vey close from the other one `ConversationEndTimeout`.
Wouldn't it be a good candidate for moving into the Debug section?

jotak: Just to clarify my comment: I think ideally this setting should be an implementation detail more than something to be configured, because in a "normal situation" we shouldn't expect to have to wait a long time before terminating the conversation, ie. it's only due to situations where there's a high latency for collection that we make it tweakable.. is that correct? So that's why it looks more to me like a debug kind of setting.

jpinsonneau: Both `conversationEndTimeout` and `conversationTerminatingTimeout` are implementation details. 
If you don't want to see these here I'm fine moving both to debug but as said in PM we are not mature enough to define default values that work best in every scenario.

I would like to avoid asking customer to update `debug` section to use supported features.

jpinsonneau: see latest comment from @ronensc:
https://issues.redhat.com/browse/NETOBSERV-915?focusedId=22417163&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22417163

PR #316, opened 2023-03-30
NETOBSERV-933: make sure flows collector CRD has the right namespace for OCP

msherif1234:
flowcollectors namespace in the bundle need to point to the new namespace `openshift-netobserv-operator`

### DIFF
--- a/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
+++ b/bundle/manifests/flows.netobserv.io_flowcollectors.yaml
@@ -12,7 +12,7 @@ spec:
       clientConfig:
         service:
           name: netobserv-webhook-service
-          namespace: netobserv
+          namespace: openshift-netobserv-operator
           path: /convert
       conversionReviewVersions:
       - v1alpha1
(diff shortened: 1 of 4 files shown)


jotak:
so when running `make deploy`, it will now deploy the operator in `openshift-netobserv-operator` ?
If so we need to check if there are other impacts (in doc, makefile, etc.)

jotak:
tested with `make deploy` +1
/lgtm

msherif1234:
/hold
seeing the following in the operator 
```bash
2023/04/03 12:23:40 http: TLS handshake error from 10.128.0.19:39758: remote error: tls: bad certificate
```

and the following in apiserver operator
```bash
E0403 12:23:26.363192       1 degraded_webhook.go:128] x509: certificate is valid for netobserv-controller-manager-service.openshift-netobserv-operator, netobserv-controller-manager-service.openshift-netobserv-operator.svc, not netobserv-webhook-service.openshift-netobserv-operator.svc
```

msherif1234:
it seems we don't this change, OLM should have set the namespace info based on where the operator is deployed in

PR #319, opened 2023-04-03
NETOBSERV-971 Fix couple of CRD issues on zero-values

jotak:
CRD cleanup / sanitizing

Find & fix a bunch of CRD issues where go zero-value was conflicting
with CRD default. The typical workaround is to use pointers or, for
lists and maps, remove "omitempty".

Add a test to help catch these issues

Found issues:
- portNamings.enable (this was the original issue)
- portNamings.portNames (could not set empty list)
- ebpf.interfaces and excludeInterfaces (e.g. could not set empty exclusion
  list)
- processor.metrics.ignoreTags and disableAlerts (could not set empty
  ignoreTags)
- processor.enableKubeProbles, dropUnusedFields (could not set false)
- processor.kafkaConsumerReplicas (could not set 0)
- loki.maxRetries (could not set 0)
- loki.staticLabels (could not set empty list)
- plugin.register (could not set false)
- plugin.replicas (could not set 0)
- plugin.quickFilters (could not set empry list)

### DIFF
--- a/controllers/flowcollector_controller_console_test.go
+++ b/controllers/flowcollector_controller_console_test.go
@@ -23,9 +23,6 @@ import (
 // For simplicity, we'll use a different namespace
 const cpNamespace = "namespace-console-specs"
 
-var tRue = true
-var fAlse = false
-
 // nolint:cyclop
 func flowCollectorConsolePluginSpecs() {
 	cpKey := types.NamespacedName{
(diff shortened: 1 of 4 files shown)


OlivierCazade (commented):
Nice catch, but I would suggest a different fix.

We have many others bool in the API and removing all of them the omitempty flag would create a lot of unnecessary mandatory fields.

After some test it would also work if we let the omitempty flag but swithc to a bool pointer.

The downside is we would have to check for nil each time which can be done in a sugar function but the good side is that we would keep the field optional in the API.

jotak:
@OlivierCazade I reviewed the whole CRD ... which leads to a bigger pull request

OlivierCazade (approved):
LGTM, and sorry for my previous comment. After discussion, your previous solution would have worked.

msherif1234 (commented):
pls consider using `"k8s.io/utils/pointer` for convert from type to *type

OlivierCazade (dismissed):

jpinsonneau:
Rebased for merge
https://github.com/netobserv/network-observability-operator/compare/b30f2bfd2ecfb9ff6c16aa48f09bc46bfcd5c6d7..6948a3a20e1d128bfd9f7d1d003b1095795e786f

PR #324, opened 2023-04-07
Add ProfilePort to the API to allow enable pprof

msherif1234:
ebpf agent need profileport configuration to open http connection and allow pprof collection as described in https://github.com/netobserv/netobserv-ebpf-agent/blob/2dca243a8dc5aa449aa0ff18a6924252cfab6c79/docs/profiling.md

However the operator didn't have this configuration knob.


msherif1234:
It seems for pprof there is a generic way to configure ebpf profile port, so need to add additional fields to flowcontroller object
```yaml
---
agent:
  type: EBPF
  ebpf:
    imagePullPolicy: IfNotPresent
    sampling: 50
    cacheActiveTimeout: 5s
    cacheMaxFlows: 100000
    interfaces: []
    excludeInterfaces:
      - lo
    logLevel: info
    resources:
      requests:
        memory: 50Mi
        cpu: 100m
      limits:
        memory: 800Mi
    kafkaBatchSize: 10485760
    debug:
      env: PORT_PROFILE:"6060"
```

PR #328, opened 2023-04-19
NETOBSERV-976 Fix disabling alerts

jotak:


OlivierCazade (approved):
Nice catch, thanks.

LGTM

PR #329, opened 2023-04-21
NETOBSERV-764 - ADD v1beta2 api version with simplified Loki configuration options

acmenezes:
On this PR:
- Add v1beta2 api version
- Move the hub for conversion webhook to this new version
- Add the options for Loki configuration

Example usage (also available in `config/samples/flows_v1beta2_flowcollector_lokistack.yaml`):
```
apiVersion: flows.netobserv.io/v1beta2
kind: FlowCollector
spec:
  loki:
    lokiStack:
      name: loki
      namespace: netobserv
```

DOC: https://github.com/openshift/openshift-docs/pull/63276


acmenezes:
@jpinsonneau with my latest update to this code I have a working v1beta2 that behaves exactly as the v1beta1. Now it is the new hub. I compiled and tested on a cluster. Would be nice to have a second pair of eyes testing it as it is. I'll proceed to add changes to the loki types.

jpinsonneau:
> @jpinsonneau with my latest update to this code I have a working v1beta2 that behaves exactly as the v1beta1. Now it is the new hub. I compiled and tested on a cluster. Would be nice to have a second pair of eyes testing it as it is. I'll proceed to add changes to the loki types.

Tested and it works as expected :+1:  Code looks good also. I wonder if there is a way to simplify these changes in the future instead of changing all the imports :thinking:

jotak:
It's only the second time we introduce a new CRD version so I must say I'm not very familiar with the process, but in my understanding when doing so, we don't change immediately the "stored" version, it still has to be the previous one that is stored, for at least one release.
Which means, if I'm correct, that the conversions still needs to be `v1alpha1 <-> v1beta1`, plus the new one `v1beta1 <-> v1beta2`; but not (yet) `v1alpha1 <-> v1beta2` (conversions are from/to the stored version)

@msherif1234 do you confirm if this is correct?

Side question: should we stop serving and/or remove v1alpha1?

msherif1234:
> It's only the second time we introduce a new CRD version so I must say I'm not very familiar with the process, but in my understanding when doing so, we don't change immediately the "stored" version, it still has to be the previous one that is stored, for at least one release. Which means, if I'm correct, that the conversions still needs to be `v1alpha1 <-> v1beta1`, plus the new one `v1beta1 <-> v1beta2`; but not (yet) `v1alpha1 <-> v1beta2` (conversions are from/to the stored version)
> 
> @msherif1234 do you confirm if this is correct?
> 
> Side question: should we stop serving and/or remove v1alpha1?
so that is correct we can't make `v1beta2` as teh storgae in the same release we introduce it it needs to be done in the following release.
but however v1beta2 can be the new hub and both v1alpha1 and v1beta1 implements the API for CovertFrom() and ConvertTo() similar to what is done here and what was done originally when we introduced the conversion webhook once this PR out of `WIP` state I will take a closer look pls tag me when its ready

api/v1alpha1/flowcollector_webhook.go
```diff
@@ -58,7 +58,33 @@ func (r *FlowCollector) ConvertTo(dstRaw conversion.Hub) error {
 		dst.Spec.Processor.Metrics.DisableAlerts = restored.Spec.Processor.Metrics.DisableAlerts
 	}
 
-	dst.Spec.Loki.StatusTLS = restored.Spec.Loki.StatusTLS
+	dst.Spec.Loki.Manual.URL = restored.Spec.Loki.Manual.URL
```
jotak, line 61: just wondering, would `dst.Spec.Loki.Manual = restored.Spec.Loki.Manual` just work for all of that?

api/v1beta1/flowcollector_types.go
```diff
@@ -735,7 +736,6 @@ type FlowCollectorStatus struct {
 // +kubebuilder:printcolumn:name="Sampling (EBPF)",type="string",JSONPath=`.spec.agent.ebpf.sampling`
 // +kubebuilder:printcolumn:name="Deployment Model",type="string",JSONPath=`.spec.deploymentModel`
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[*].reason"
-// +kubebuilder:storageversion
```
jotak, line 738: so you must rollback and keep v1beta1 as the storage version, for at least one release. It shouldn't be an issue that the storage version differs from the hub version.

config/manager/kustomization.yaml
```diff
@@ -14,7 +14,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/netobserv/network-observability-operator
-  newTag: 1.0.2
+  newName: quay.io/acmenezes/network-observability-operator
+  newTag: dev
```
jotak, line 18: this will have to be rollbacked at some point (you can keep it as long as it's a draft if that's easier for you)

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,769 @@
+/*
```
jotak, line 1: This whole API has received a bunch of changes in the meantime, but since it's a new file it may get unnoticed when you'll rebase, so we need to be careful to not loose anything.
Maybe it will be safer to restart from a fresh copy of v1beta1 and reapply your changes

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,769 @@
...
+	LokiAuthUseHostToken     = "HOST"
+	LokiAuthForwardUserToken = "FORWARD"
+)
+
+type LokiManualParams struct {
+	//+kubebuilder:default:="http://loki:3100/"
+	// url is the address of an existing Loki service to push the flows to. When using the Loki Operator,
+	// set it to the Loki gateway service with the `network` tenant set in path, for example
+	// https://loki-gateway-http.netobserv.svc:8080/api/logs/v1/network.
+	URL string `json:"url,omitempty"`
```
jotak, line 468: With the Manual configuration we want to take this opportunity to refactor a bit how these URLs are handled, by making them more explicit, with each having a single responsibility. So we already have the querierUrl and the statusUrl, but now they will be required (in manual config) and not optional ; and instead of this general-purpose `URL`, we should have a `IngesterURL` . We'll need also to reformulate the doc text but that can be done in a later step.

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,769 @@
...
+
+	//+kubebuilder:validation:Minimum=0
+	//+kubebuilder:default:=2
+	// maxRetries is the maximum number of retries for client connections.
+	MaxRetries *int32 `json:"maxRetries,omitempty"`
+
+	//+kubebuilder:default:={"app":"netobserv-flowcollector"}
+	// +optional
+	// staticLabels is a map of common labels to set on each flow.
+	StaticLabels map[string]string `json:"staticLabels"`
```
jotak, line 530: All these settings, from "batchWait" to "staticLabels", are not dependent on the Loki `mode` (they are common for all 4 modes), so I think they should remain in `FlowCollectorLoki`.

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,769 @@
...
+
+	// tls client configuration for loki status URL.
+	// +optional
+	StatusTLS ClientTLS `json:"statusTls"`
+}
+
+// FlowCollectorLoki defines the desired state for FlowCollector's Loki client.
+type FlowCollectorLoki struct {
+
+	//+kubebuilder:validation:Enum=manual
```
jotak, line 544: In the `enum`, values must be capitalized (it's a best practice from the API guide), so should be MANUAL here.

jotak:
Thanks @acmenezes for starting this!
As you can see it needs a rebase and there's a bunch of conflicts 😨 
don't hesitate to ask if you need some help, I can take some time this week to work with you on it

api/v1alpha1/flowcollector_webhook.go
acmenezes: I think so, but not completely sure. I'm checking on that.

api/v1beta1/flowcollector_types.go
acmenezes: Ok. I'll try that. But remember having trouble keeping the hub on the new one and the storage version on v1beta1. Anyway I'll try it.

config/manager/kustomization.yaml
acmenezes: For sure. I won't let it there. I actually shouldn't have committed it.

config/manager/manager.yaml
```diff
@@ -30,11 +30,11 @@ spec:
...
         env:
           - name: RELATED_IMAGE_EBPF_AGENT
-            value: quay.io/netobserv/netobserv-ebpf-agent:v0.3.0
+            value: quay.io/netobserv/netobserv-ebpf-agent:main
           - name: RELATED_IMAGE_FLOWLOGS_PIPELINE
-            value: quay.io/netobserv/flowlogs-pipeline:v0.1.8
+            value: quay.io/netobserv/flowlogs-pipeline:main
           - name: RELATED_IMAGE_CONSOLE_PLUGIN
-            value: quay.io/netobserv/network-observability-console-plugin:v0.1.9
+            value: quay.io/netobserv/network-observability-console-plugin:main
```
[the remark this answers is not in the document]
acmenezes: Totally agree.

api/v1beta2/flowcollector_types.go
acmenezes: That's right. I'm tried rebasing today but I see a bunch of conflicts that I think it is not worth trying to solve for the sake to time. So I'm about to do exactly that and force update this branch.

api/v1beta2/flowcollector_types.go
acmenezes: Right... thanks for that. I didn't know about those fields. I'll take that into consideration as well.

api/v1alpha1/flowcollector_webhook.go
acmenezes: Finally getting this one. Yeah, it seems obvious this one and I honestly don't remember why I made all fields separately. Just changed. Thanks for catching that one.

api/v1beta1/flowcollector_webhook.go
```diff
@@ -16,17 +16,129 @@ limitations under the License.
...
+
+	if restored.Spec.Processor.ConversationEndTimeout != nil {
+		dst.Spec.Processor.ConversationEndTimeout = restored.Spec.Processor.ConversationEndTimeout
+	}
+
+	if restored.Spec.Processor.Metrics.DisableAlerts != nil {
+		dst.Spec.Processor.Metrics.DisableAlerts = restored.Spec.Processor.Metrics.DisableAlerts
+	}
+
+	dst.Spec.Loki.Manual.URL = restored.Spec.Loki.Manual.URL
```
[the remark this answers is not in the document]
acmenezes: Done!

config/manager/kustomization.yaml
acmenezes: It's cleaned up now.

config/manager/manager.yaml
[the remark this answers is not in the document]
acmenezes: Also cleaned up this one!

api/v1beta2/flowcollector_types.go
acmenezes: As you already know it didn't try to solve the conflicts.

So just to let my rebase process registered here (I actually didn't exactly rebased)
- I did a git reset --soft HEAD~3 taking off all and only my commits.
- I stashed those commits out of the working branch
- then rebased with the main branch updated with no errors
- then I opened my stash carefully file by file by using git show stash@{0}:<the file I wanted back> > a diferent name for that file like append a letter b
- then I opened compare selected in VSCode for each and every file
- so I only applied the changes that I actually made letting the new content intact

On the flowcollector_types I had to recopy the v1beta1 and apply my changes after that. If you want to double check, which is highly appreciated, used the compare selected in VSCode with v1beta1 and v1beta2.

jotak: just did a side by side comparison, looks good!

api/v1beta2/flowcollector_types.go
acmenezes: Done! Put them back on FlowCollectorLoki. Let me know if that's what you had in mind.

api/v1beta2/flowcollector_types.go
acmenezes: Looks like this one will require changing FLP. Is that correct? May I proceed with the change and open a second PR there?

jotak: Hmm I don't see changes necessary in FLP repo (only in `flp_common_objects.go`, from this repository, where you'd need to set

```go
URL:            b.desired.Loki.Manual.IngesterURL,
```

in the `api.WriteLoki` definition

Or do you see something else?

acmenezes: Ok. I saw the FLP breaking after the changes. I'll give it a second run and do some troubleshooting on this.

acmenezes: So there is actually one change that came in the makefile provoking this error while testing. amd64 is being appended silently to the tags.

acmenezes: This one is also done now.

Makefile
```diff
@@ -259,6 +259,13 @@ generate-go-conversions: $(CONVERSION_GEN) ## Run all generate-go-conversions
 		--output-file-base=zz_generated.conversion \
 		$(CONVERSION_GEN_OUTPUT_BASE) \
 		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
+		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1beta1"
+		$(CONVERSION_GEN) \
+		--input-dirs=./api/v1beta1 \
+		--build-tag=ignore_autogenerated_core \
+		--output-file-base=zz_generated.conversion \
+		$(CONVERSION_GEN_OUTPUT_BASE) \
+		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
```
jpinsonneau, line 268: any reason to separate these commands instead of running all at once ?

Something around this may work:
```suggestion
		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1alpha1 ./api/v1beta1"
		$(CONVERSION_GEN) \
		--input-dirs=./api/v1alpha1 \
		--input-dirs=./api/v1beta1 \
		--build-tag=ignore_autogenerated_core \
		--output-file-base=zz_generated.conversion \
		$(CONVERSION_GEN_OUTPUT_BASE) \
		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt			
```

jpinsonneau: done in https://github.com/netobserv/network-observability-operator/pull/329/commits/ae2fabd82a3b4011bf18925c3fbe6d66a9853989

jpinsonneau:
@jotak do you have an idea about the following error message while updating the bundle ?
```
msg="Warning: Value flows.netobserv.io/v1beta2, Kind=FlowCollector: provided API should have an example annotation" time="2023-09-15T13:01:13+02:00" level=warning msg="Warning: Value netobserv-operator.v1.0.4: owned CRD \"flowcollectors.flows.netobserv.io\" has an empty description"
please correct warnings and errors first
make: *** [Makefile:371: bundle] Error 255
```

The generated CSV lack doesn't append `description`  for `v1beta2` whereas it actually have one:
```
// `FlowCollector` is the schema for the network flows collection API, which pilots and configures the underlying deployments.
type FlowCollector struct {
```

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,905 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:printcolumn:name="Agent",type="string",JSONPath=`.spec.agent.type`
+// +kubebuilder:printcolumn:name="Sampling (EBPF)",type="string",JSONPath=`.spec.agent.ebpf.sampling`
+// +kubebuilder:printcolumn:name="Deployment Model",type="string",JSONPath=`.spec.deploymentModel`
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[*].reason"
+// +kubebuilder:storageversion
```
OlivierCazade, line 883: I missed some conversation, but are we finally switching to this new version as storage directly?

jpinsonneau: As far as I understood this was the case if we wanted this in 1.4 release to have `v1alpha1` <-> `v1beta1` <-> `v1beta2` path. It's not the case anymore.

@jotak can you please confirm ?

api/v1beta1/flowcollector_webhook.go
```diff
@@ -16,17 +16,100 @@ limitations under the License.
...
-// Hub marks this version as a conversion hub.
-// All the other version need to provide converters from/to this version.
-// https://book.kubebuilder.io/multiversion-tutorial/conversion-concepts.html
-func (*FlowCollector) Hub()     {}
-func (*FlowCollectorList) Hub() {}
+// This function need to be manually created because conversion-gen not able to create it intentionally because
+// we have new defined fields in v1beta2 not in v1beta1
+// nolint:golint,stylecheck,revive
+func Convert_v1beta2_FlowCollectorLoki_To_v1beta1_FlowCollectorLoki(in *v1beta2.FlowCollectorLoki, out *FlowCollectorLoki, s apiconversion.Scope) error {
+	return autoConvert_v1beta2_FlowCollectorLoki_To_v1beta1_FlowCollectorLoki(in, out, s)
```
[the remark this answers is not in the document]
jpinsonneau: That's right, currently the autoconvert doesn't cover these:

```
func autoConvert_v1beta2_FlowCollectorLoki_To_v1beta1_FlowCollectorLoki(in *v1beta2.FlowCollectorLoki, out *FlowCollectorLoki, s conversion.Scope) error {
	// WARNING: in.Mode requires manual conversion: does not exist in peer-type
	// WARNING: in.Manual requires manual conversion: does not exist in peer-type
	// WARNING: in.LokiStack requires manual conversion: does not exist in peer-type
	out.Enable = (*bool)(unsafe.Pointer(in.Enable))
	out.BatchWait = (*v1.Duration)(unsafe.Pointer(in.BatchWait))
	out.BatchSize = in.BatchSize
	out.Timeout = (*v1.Duration)(unsafe.Pointer(in.Timeout))
	out.MinBackoff = (*v1.Duration)(unsafe.Pointer(in.MinBackoff))
	out.MaxBackoff = (*v1.Duration)(unsafe.Pointer(in.MaxBackoff))
	out.MaxRetries = (*int32)(unsafe.Pointer(in.MaxRetries))
	out.StaticLabels = *(*map[string]string)(unsafe.Pointer(&in.StaticLabels))
	return nil
}
```

I can force `manual` mode and `bind each field`. I don't think parsing the config to check if `LokiStack` mode can be enabled is worth. WDYT ?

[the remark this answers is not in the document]
jpinsonneau: that one should do the trick https://github.com/netobserv/network-observability-operator/pull/329/commits/3b2bd0781a555f1de5997654e653801c620bef91

api/v1beta2/flowcollector_types.go
msherif1234: not sure I follow, but new api version can't be set as storage version in the same release this is what I did when transition from v1alpha1 to v1beta1

api/v1alpha1/flowcollector_types.go
```diff
@@ -55,6 +55,7 @@ type FlowCollectorSpec struct {
 	Processor FlowCollectorFLP `json:"processor,omitempty"`
 
 	// loki, the flow store, client settings.
+	// +k8s:conversion-gen=false
```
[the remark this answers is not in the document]
jotak: I don't think it was the intent, but that said, we should think about removing alpha1 ... Since it's alpha we don't even need to go through a deprecation notice

pkg/volumes/builder.go
```diff
@@ -129,3 +129,16 @@ func buildVolumeAndMount(refType flowslatest.MountableType, refName string, volu
 		MountPath: "/var/" + volumeName,
 	}
 }
+
+func (b *Builder) insertOrReplace(vi *VolumeInfo) {
```
jotak, line 133: why do we need that? Are we adding several times the same volume?

pkg/helper/flowcollector.go
```diff
@@ -99,6 +83,139 @@ func UseLoki(spec *flowslatest.FlowCollectorSpec) bool {
 	return spec.Loki.Enable == nil || *spec.Loki.Enable
 }
```
jotak, line 85: In this whole file, I think instead of having getter functions for every single field, in that case, it would be simpler to create a new struct (that would be very similar to the Manual struct.. maybe derived from it) and have conversion functions to this struct for every mode.
I can do a suggestion around that if this isn't clear...

Makefile
```diff
@@ -252,13 +252,15 @@ doc: crdoc ## Generate markdown documentation
 	$(CRDOC) --resources config/crd/bases/flows.netobserv.io_flowcollectors.yaml --output docs/FlowCollector.md
 
 generate-go-conversions: $(CONVERSION_GEN) ## Run all generate-go-conversions
-		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1alpha1"
+		$(MAKE) clean-generated-conversions SRC_DIRS="./api/v1alpha1 ./api/v1beta1 ./api/v1beta2"
```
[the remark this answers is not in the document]
jpinsonneau: correct I can remove it from here

pkg/volumes/builder.go
jpinsonneau: While adding the tests on console plugin I faced issues with volumeInfos duplicated.
This solve the issues and ensure each volume is appened once.

https://github.com/netobserv/network-observability-operator/pull/329/commits/9f52e81b6f5ddc0ae528e58a57a69c345648eadc

pkg/helper/flowcollector.go
jpinsonneau: Why not. That's a good candidate for the followup with `monolith` and `distributed` modes.

api/v1beta2/flowcollector_types.go
jotak: No I think @msherif1234 is right we can't change the storage version now, we should rollback this change

jpinsonneau: ok sure, I'm on it :+1:

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
jpinsonneau: Removed in https://github.com/netobserv/network-observability-operator/pull/329/commits/656cb0fb4fd5982bad0d6f0110a2eb1bb4f2e898

What's the best practice here ? Removing `v1alpha1` while introducing `v1beta2` or wait for next release ?

api/v1beta1/flowcollector_types.go
```diff
@@ -57,7 +57,8 @@ type FlowCollectorSpec struct {
 	// enriches them, generates metrics, and forwards them to the Loki persistence layer and/or any available exporter.
 	Processor FlowCollectorFLP `json:"processor,omitempty"`
 
-	// Loki, the flow store, client settings.
+	// loki, the flow store, client settings.
+	// +k8s:conversion-gen=false
```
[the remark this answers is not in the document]
jpinsonneau: removed in https://github.com/netobserv/network-observability-operator/pull/329/commits/656cb0fb4fd5982bad0d6f0110a2eb1bb4f2e898

api/v1beta2/flowcollector_types.go
jpinsonneau: Moved the storage version to `v1beta1` in https://github.com/netobserv/network-observability-operator/pull/329/commits/656cb0fb4fd5982bad0d6f0110a2eb1bb4f2e898

jotak:
@jpinsonneau thanks for the changes on storage version, it's ok to address the other feedback in a follow-up (and the remaining stuff for other modes)
I've one more thing though: shouldn't the `Hub` be defined on the stored version? cc @msherif1234 
Currently, hub is moved to v1beta2 ; I'm not sure if this is a problem or not

Makefile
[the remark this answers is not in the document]
jpinsonneau: done in https://github.com/netobserv/network-observability-operator/pull/329/commits/c2d87fbe653d1534272ca3b250bd35f34780025d

api/v1alpha1/flowcollector_webhook.go
```diff
@@ -20,63 +20,75 @@ import (
...
+		// fallback on current loki config as Manual mode if metadata are not available
+		dst.Spec.Loki.Mode = v1beta2.LokiModeManual
+		dst.Spec.Loki.Manual.IngesterURL = r.Spec.Loki.URL
+		dst.Spec.Loki.Manual.QuerierURL = r.Spec.Loki.QuerierURL
+		dst.Spec.Loki.Manual.StatusURL = r.Spec.Loki.StatusURL
+		dst.Spec.Loki.Manual.TenantID = r.Spec.Loki.TenantID
+		dst.Spec.Loki.Manual.AuthToken = r.Spec.Loki.AuthToken
+		if err := Convert_v1alpha1_ClientTLS_To_v1beta2_ClientTLS(&r.Spec.Loki.TLS, &dst.Spec.Loki.Manual.TLS, nil); err != nil {
+			return fmt.Errorf("copying v1alplha1.Loki.TLS into v1beta2.Loki.Manual.TLS: %w", err)
+		}
```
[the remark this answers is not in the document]
jpinsonneau: I had issues when the storage was on `v1beta2` :thinking: 
Let's see if I can remove these now that the storage is on `v1beta1`

jpinsonneau:
> @jpinsonneau thanks for the changes on storage version, it's ok to address the other feedback in a follow-up (and the remaining stuff for other modes) I've one more thing though: shouldn't the `Hub` be defined on the stored version? cc @msherif1234 Currently, hub is moved to v1beta2 ; I'm not sure if this is a problem or not

Yeah I had the same question yesterday and Mohamed confirmed `v1beta2` is now the hub [as same as v1beta1 was when we introduced it](https://github.com/netobserv/network-observability-operator/pull/258/files#diff-757ddd7331ae0bbbd68d3fc993140dcb71d0613782afd0d54c68f32752a3a86f)

jpinsonneau:
I still have issues with the test bench volumes not mounted since I move the storage to `v1beta1`. It's working on a real environment :thinking:  
If you have any idea in mind I would appreaciate the help

controllers/flowlogspipeline/flp_test.go
```diff
@@ -98,29 +99,57 @@ func getConfig() flowslatest.FlowCollectorSpec {
...
-		},
+		Loki: getLoki(lokiMode...),
 		Kafka: flowslatest.FlowCollectorKafka{
 			Address: "kafka",
 			Topic:   "flp",
 		},
 	}
 }
 
+func getLoki(lokiMode ...string) flowslatest.FlowCollectorLoki {
```
jotak, line 110: this looks a bit weird to me to allow variable args, while only the 0 index is used, for switching between modes . Is there a reason to do that instead of an explicit switch / single arg?
(anyway this can be addressed in a follow-up)

api/v1alpha1/flowcollector_webhook.go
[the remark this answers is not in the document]
jpinsonneau: I confirm we still need these pieces for the first deployment.
You can reproduce by deploying the operator and applying `flows_v1alpha1_flowcollector.yaml` with a custom loki url from scratch (no existing CR instance).

Without this part of code we lost the custom url and it fallback to its default since the field was moved from `spec.loki.url` to `spec.loki.manual.url`.

api/v1alpha1/flowcollector_types.go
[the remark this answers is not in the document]
jpinsonneau: deprecated in https://github.com/netobserv/network-observability-operator/pull/329/commits/282b28ff4a20fdc7242a1903978b0cb098100565

jpinsonneau:
I fixed the tests according to @msherif1234 solution in https://github.com/netobserv/network-observability-operator/pull/329/commits/282b28ff4a20fdc7242a1903978b0cb098100565 Thanks for the help :partying_face: 

It's not perfect but it forces `v1beta2` to be storage in the unit tests.
Warning: we need to keep the hack CRD up to date to keep it work. That can be automated if we need it long term (but it's not supposed to)

controllers/flowlogspipeline/flp_test.go
jpinsonneau: the goal was to have the ability to call `getConfig` with some optionnal param(s) like `getConfig("LOKISTACK")` and keep the default empty `getConfig()`

Can you please add any thoughts on testing in [NETOBSERV-1302](https://issues.redhat.com/browse/NETOBSERV-1302) so we can improve all at once ?

jpinsonneau:
FYI I'm working on a followup [introducing missing modes](https://github.com/netobserv/network-observability-operator/pull/454/files#diff-5e052f7497247563d56f113532f1ee1408f78ec69321db4a34b6c45bc54731e4R566-R591) and [automating the hack CRD](https://github.com/netobserv/network-observability-operator/pull/454/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L383)

api/v1beta1/flowcollector_webhook.go
```diff
@@ -16,17 +16,134 @@ limitations under the License.
...
+	dst.Spec.Processor.LogTypes = restored.Spec.Processor.LogTypes
+	if restored.Spec.Processor.ConversationHeartbeatInterval != nil {
+		dst.Spec.Processor.ConversationHeartbeatInterval = restored.Spec.Processor.ConversationHeartbeatInterval
+	}
+	if restored.Spec.Processor.ConversationEndTimeout != nil {
+		dst.Spec.Processor.ConversationEndTimeout = restored.Spec.Processor.ConversationEndTimeout
+	}
+	if restored.Spec.Processor.Metrics.DisableAlerts != nil {
+		dst.Spec.Processor.Metrics.DisableAlerts = restored.Spec.Processor.Metrics.DisableAlerts
+	}
```
jotak, line 70: these fields were already present in v1beta1, I don't think this conversion is really needed?

api/v1beta2/flowcollector_types.go
```diff
@@ -0,0 +1,901 @@
...
+	// +kubebuilder:validation:Required
+	Type ExporterType `json:"type"`
+
+	// Kafka configuration, such as the address and topic, to send enriched flows to.
+	// +optional
+	Kafka FlowCollectorKafka `json:"kafka,omitempty"`
+
+	// IPFIX configuration, such as the IP address and port to send enriched IPFIX flows to. [Unsupported (*)].
+	// +optional
+	IPFIX FlowCollectorIPFIXReceiver `json:"ipfix,omitempty"`
```
jotak, line 860: A `diff` with v1beta1 shows some unintended changes here, here's a fix:

```suggestion
	// `type` selects the type of exporters. The available options are `KAFKA` and `IPFIX`.
	// +unionDiscriminator
	// +kubebuilder:validation:Enum:="KAFKA";"IPFIX"
	// +kubebuilder:validation:Required
	Type ExporterType `json:"type"`

	// Kafka configuration, such as the address and topic, to send enriched flows to.
	// +optional
	Kafka FlowCollectorKafka `json:"kafka,omitempty"`

	// IPFIX configuration, such as the IP address and port to send enriched IPFIX flows to.
	// +optional
	IPFIX FlowCollectorIPFIXReceiver `json:"ipfix,omitempty"`
```

jpinsonneau: good catch !

api/v1beta1/flowcollector_webhook.go
jpinsonneau: that's right, I missed to remove that part. Thanks

jotak (approved):
thanks, LGTM, I'll report my unresolved comments on the follow-up PR

jpinsonneau:
@nathan-weinberg since the API changed, I guess this will impact your test bench, at least if you want to create the latest API version and use `LOKISTACK` mode. Else you can keep creating `v1beta1` and it will fallback to `MANUAL` mode.

Please let us know if you have any issue, and feel free to merge when you are ready.
Thanks !

nathan-weinberg:
> @nathan-weinberg since the API changed, I guess this will impact your test bench, at least if you want to create the latest API version and use `LOKISTACK` mode. Else you can keep creating `v1beta1` and it will fallback to `MANUAL` mode.
> 
> Please let us know if you have any issue, and feel free to merge when you are ready. Thanks !

Thanks for the heads up @jpinsonneau - two quick questions, first does this PR require pre-merge QE testing, second will old flowcollector configs still be compatable? If the latter and not the former I can create a QE automation task to update it sometime this release and we can go ahead and merge this, if the former I can do both at once

jpinsonneau:
> does this PR require pre-merge QE testing ?

No, it can be done after to unlock other PRs (any touching API is impacted)
@jotak @msherif1234 is it good for you guys to merge as is ?

> will old flowcollector configs still be compatable?

Yes it is ! you can create `v1beta1` and it will be automatically converted to `v1beta2`

jotak:
@acmenezes big thanks for the initial work and for your patience!

PR #331, opened 2023-04-28
NETOBSERV-1062: Add TCP drop and DNS tracking to netobserv operator

msherif1234:
TCPdrop tracepoint need special volume mount and added a flag to control TCPdrop feature
by default its turned off


controllers/ebpf/agent_controller.go
```diff
@@ -160,13 +162,48 @@ func (c *AgentController) current(ctx context.Context) (*v1.DaemonSet, error) {
...
 		return nil
 	}
 	version := helper.ExtractVersion(c.config.EBPFAgentImage)
 	volumeMounts := []corev1.VolumeMount{}
 	volumes := []corev1.Volume{}
+
+	if coll.Spec.Agent.EBPF.EnableTCPDrop != nil && *coll.Spec.Agent.EBPF.EnableTCPDrop {
+		//NOTE: To be able to mount kernel trace events needed by TCPDrop feature,
+		// POD has to run as privileged.
+		coll.Spec.Agent.EBPF.Privileged = true
```
jpinsonneau, line 188: Could we please mention that in the CRD description ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -403,6 +403,24 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
+				},
+				{
+					Name:      "TCPDropPackets",
+					Operation: "sum",
+				},
+				{
+					Name:      "TCPDropPackets",
+					Operation: "sum",
+					SplitAB:   true,
+				},
```
jpinsonneau, line 423: We should conditionnaly append these when `EnableTCPDrop` is true

jpinsonneau:
After testing this on clusterbot `4.13 aws,large` with `enableTCPDrop: true`, I get the following error:
```
Error creating: pods "netobserv-ebpf-agent-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used, spec.containers[0].securityContext.runAsUser: Invalid value: 0: must be in the ranges: [1000710000, 1000719999], spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed, spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used, provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```

```
  relatedImages:
    - image: 'quay.io/netobserv/netobserv-ebpf-agent:adfa724'
      name: ebpf-agent
    - image: 'quay.io/netobserv/flowlogs-pipeline:06ebf5e'
      name: flowlogs-pipeline
    - image: 'quay.io/netobserv/network-observability-console-plugin:main'
      name: console-plugin
```

docs/FlowCollector.md
```diff
@@ -4007,6 +4007,15 @@ ebpf describes the settings related to the eBPF-based flow reporter when the "ag
           Debug allows setting some aspects of the internal configuration of the eBPF agent. This section is aimed exclusively for debugging and fine-grained performance optimizations (for example GOGC, GOMAXPROCS env vars). Users setting its values do it at their own risk.<br/>
         </td>
         <td>false</td>
+      </tr><tr>
+        <td><b>enableTCPDrop</b></td>
+        <td>boolean</td>
+        <td>
+          enable TCP drop flows logging feature<br/>
+          <br/>
+            <i>Default</i>: false<br/>
```
eranra, line 4016: Any reason why this is not enabled by default? The overhead should not be huge, and this is a very useful feature for customers. If somehow doesn't want that he can just disable.

msherif1234: we can for sure make the default on just felt it might be less overhead/risk to start with it as off by default and based on feedback we can decide

jotak: I didn't realize it requires privileged / mounting kernel debug FS. That could be a reason to not turn it on by default :-(

msherif1234: yes

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -319,6 +319,89 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
+		outDNSTrackingFields := []api.OutputField{
+			{
+				Name:      "DnsRequestTimeMs",
+				Operation: "min",
+			},
+			{
+				Name:      "DnsResponseTimeMs",
+				Operation: "max",
+			},
+		}
```
jpinsonneau, line 401: I can see request and response times using this method however we don't have both in the same conversation since we [reintroduced FlowDirection](https://issues.redhat.com/browse/NETOBSERV-915) to avoid duplicates.

image

Adding DNS Id on conversations may also be tricky since it rely on `min` / `max` aggregators on `DnsRequestTimeMs` / `DnsResponseTimeMs` fields.

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -319,6 +319,89 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
+			{
+				Name:      "TCPDropPackets",
+				Operation: "sum",
+			},
+			{
+				Name:      "TCPDropPackets",
+				Operation: "sum",
+				SplitAB:   true,
+			},
+		}
```
jpinsonneau, line 387: Should we manage `TcpDropLatestState` & `TcpDropLatestDropCause` here ?

msherif1234: Not sure can explain how that would help from ui I can add them

jpinsonneau: On `FlowCollector` spec side, you can define to log flow / conversations or both.

On frontend side, the query option dropdown show what's available.

In case you are using only conversations, it's usefull to be able to query metrics based on conversations.
These metrics are used for both `Overview` and `Topology` tabs.

config/openshift-olm/rbac.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jotak, line 1: wouldn't that install the privileged mode unconditionally? I'd say it should be installed only if `privileged` is true in FlowCollector, no?

msherif1234: I don't think so @jotak this just creating service account the error we get when enable privileged mode was
```sh
Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```
I don't know how if that config knob ever worked before seems like day1 bug ?

jotak: I just did the test to switch on privileged mode (not with your PR, but from `main` ) : it works as expected, ie. the agent pods don't have the CAP_* stuff anymore but just have the `privileged` flag. So that doesn't look like a day1 bug, maybe that's something that is additionally needed with your changes?

Anyway, my point wasn't about that, but it's that this role binding should only be installed when `privileged` is `true`, and removed when set back to `false`, so something to reconcile probably somewhere here: https://github.com/netobserv/network-observability-operator/blob/main/controllers/ebpf/internal/permissions/permissions.go

api/v1beta1/flowcollector_types.go
```diff
@@ -220,6 +220,20 @@ type FlowCollectorEBPF struct {
 	// (for example GOGC, GOMAXPROCS env vars). Users setting its values do it at their own risk.
 	// +optional
 	Debug DebugConfig `json:"debug,omitempty"`
+
+	// enable TCP drop flows logging feature, this feature requires mounting
+	// kernel debug filesystem hence the ebpf pod has to run as privileged
+	// if privileged isn't set an error will be reported.
```
jotak, line 226: Some doc format:

```suggestion
	// Enable TCP drop flows logging feature. This feature requires mounting
	// kernel debug filesystem hence the eBPF pod has to run as privileged.
	// If privileged is not set, an error will be reported.
```

cc @skrthomas : any other suggestions?

api/v1beta1/flowcollector_types.go
```diff
@@ -220,6 +220,20 @@ type FlowCollectorEBPF struct {
...
+	//+kubebuilder:default:=false
+	//+optional
+	EnableTCPDrop *bool `json:"enableTCPDrop,omitempty"`
+
+	// enable DNS tracking feature, this feature requires mounting
+	// kernel debug filesystem hence the ebpf pod has to run as privileged
+	// if privileged isn't set an error will be reported.
```
jotak, line 233:
```suggestion
	// Enable DNS tracking feature. This feature requires mounting
	// kernel debug filesystem hence the eBPF pod has to run as privileged.
	// If privileged is not set, an error will be reported.
```

cc @skrthomas

controllers/ebpf/agent_controller.go
```diff
@@ -160,6 +175,30 @@ func (c *AgentController) desired(ctx context.Context, coll *flowslatest.FlowCol
 	if err != nil {
 		return nil, err
 	}
+	volumeMounts := c.volumes.GetMounts()
+	volumes := c.volumes.GetVolumes()
+
+	if helper.IsTCPDropEnabled(&coll.Spec) || helper.IsDNSTrackingEnabled(&coll.Spec) {
+		if !coll.Spec.Agent.EBPF.Privileged {
+			return nil, fmt.Errorf("to use TCPDrop feature privileged mode need to be enabled")
```
jotak, line 183: Seems a bit harsh to return an error: it will make reconciliation failing in loop until the user fixes it, so nothing would be usable, as the agent pods won't be deployed. It could be a strategy to force users fix the issue (they'll have to look at the operator logs to figure it out) - but still I find it harsh.

I see two alternatives:
- Just log the error instead and proceed as if the two features were disabled (ideally we should also propagate some warning into the CR status/conditions - could be a follow-up, I'm not sure if we implement that already)
- Or make these features also turn on the privileged mode, automatically. This is probably more user-friendly, but we should be crystal-clear about that in the documentation.

any thoughts?

msherif1234: I originally was enabling privileged under the hood but changed that as I don't think this is good practice its better to be explicit with the configs rather that implicit and we also call that clearly in the doc if u enable those features u have to enable privileged otherwise it won't work so we can log and err/warning and proceed ?

api/v1beta1/flowcollector_types.go
```diff
@@ -221,6 +221,20 @@ type FlowCollectorEBPF struct {
 	// such as GOGC and GOMAXPROCS env vars. Users setting its values do it at their own risk.
 	// +optional
 	Debug DebugConfig `json:"debug,omitempty"`
+
+	// Enable TCP drop flows logging feature, This feature requires mounting
+	// kernel debug filesystem hence the eBPF pod has to run as privileged.
+	// If privileged is not set, an error will be reported.
```
skrthomas, line 227:
```suggestion
	// Enable the TCP drop flows logging feature. This feature requires mounting
	// the kernel debug filesystem, so the eBPF pod has to run as privileged.
	// If privileged is not set, an error is reported.
```

skrthomas: Suggestions from me + some of what Joel recommended as well.

api/v1beta1/flowcollector_types.go
skrthomas: I made another suggestion. One question `privileged` here is the `spec.agent.eBPF.privileged` parameter? If so, I suggest:
> If the `spec.agent.eBPF.privileged` parameter is not set, an error is reported.

jpinsonneau (approved):
Looks good to me, tested and working fine :+1: 

thanks @msherif1234 !

controllers/constants/constants.go
```diff
@@ -16,7 +16,11 @@ const (
 	EBPFServiceAccount     = EBPFAgentName
 	EBPFSecurityContext    = EBPFAgentName
 
-	OpenShiftCertificateAnnotation = "service.beta.openshift.io/serving-cert-secret-name"
+	OpenShiftCertificateAnnotation  = "service.beta.openshift.io/serving-cert-secret-name"
+	OpenShiftClusterRoleName        = OperatorName + "-privileged-scc"
+	OpenShiftClusterRoleBindingName = OperatorName + "-privileged-scc"
```
jotak, line 21: those are now unused, aren't they?

jotak:
just some unused vars to remove, apart from that lgtm
(we'll have also to clarify the need for privilege mode for DNS)

api/v1beta1/flowcollector_types.go
```diff
@@ -221,6 +221,20 @@ type FlowCollectorEBPF struct {
...
+	// Enable the TCP drop flows logging feature. This feature requires mounting
+	// the kernel debug filesystem, so the eBPF pod has to run as privileged.
+	// If the spec.agent.eBPF.privileged parameter is not set, an error is reported.
+	//+kubebuilder:default:=false
+	//+optional
+	EnableTCPDrop *bool `json:"enableTCPDrop,omitempty"`
+
+	// Enable the DNS tracking feature. This feature requires mounting
+	// the kernel debug filesystem hence the eBPF pod has to run as privileged.
+	// If the spec.agent.eBPF.privileged parameter is not set, an error is reported.
```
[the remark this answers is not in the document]
msherif1234: the tracepoint hooks needs access to host FS

controllers/constants/constants.go
msherif1234: yeah will remove those unused

jotak:
Looks good to me
@memodi @nathan-weinberg @Amoghrd : I guess you'll want to test NETOBSERV-1062 all at once: it also needs the console PR https://github.com/netobserv/network-observability-console-plugin/pull/324 - should we hold that one for pre-merge testing?

jpinsonneau (dismissed):
/lgtm

memodi:
@msherif1234 @jpinsonneau  curious, what is "breaking change" here?

msherif1234:
> @msherif1234 @jpinsonneau curious, what is "breaking change" here?

@memodi that tag was added when we had an issue with older releases but issue was addressed by https://github.com/netobserv/netobserv-ebpf-agent/pull/144

memodi:
> could we list these 3 options in flowcollector yaml config in order:
never mind, I think it follows alphabetical order.

PR #333, opened 2023-05-04
NETOBSERV-1003: Added monitoring role to bundle

OlivierCazade:
Add monitoring role to the bundle.

### DIFF
--- a/config/openshift-olm/monitor.yaml
+++ b/config/openshift-olm/monitor.yaml
@@ -19,3 +19,34 @@ spec:
   selector:
     matchLabels:
       control-plane: controller-manager
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: openshift-netobserv-operator-prometheus
+  namespace: openshift-netobserv-operator
+rules:
+  - apiGroups:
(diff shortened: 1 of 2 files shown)


OlivierCazade:
/hold

Testing

memodi (dismissed):
LGTM
I tried testing  using `make-deploy` but it did not deploy operator in `openshift-netobserv-operator` NS and also had to update `DOWNSTREAM_DEPLOYMENT` variable. If your testing so far  looks good. I think we should go ahead merge it and can do further testing on Downstream builds.

/label qe-approved

msherif1234:
@OlivierCazade do u think it’s better to group role and role_binding for Prometheus with the rest under Config/rbac folder ?

OlivierCazade:
@msherif1234 config/rbac is general and used in every kind of deployment. The monitoring part is proper to Openshift this is why I added the rules there to avoid deploying unnecessary rule.

jotak:
shouldn't it be in the `config/openshift-olm` directory? I'm a bit confused.. I think `openshift-olm` is the one used for downstream builds.. Maybe add it in both places ?

PR #335, opened 2023-05-15
NETOBSERV-1033: Lock down Network Observability

stleerh:
Network policy to lock down ingress traffic to 'netobserv' namespace

### DIFF
--- a/config/policy/lockdown-netobserv.yaml
+++ b/config/policy/lockdown-netobserv.yaml
@@ -1,0 +1,21 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: lockdown-netobserv
+  namespace: netobserv
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - podSelector: {}  # without namespaceSelecor, it defaults to 'same namespace'
+        - podSelector: {}
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: openshift-console
+        - podSelector: {}
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: openshift-monitoring
+  policyTypes:
+    - Ingress
+status: {}


msherif1234:
@stleerh `netobserv` namespace is configurable in the flowcollector object, the default value is `netobserv` but user can change it, so we probably need to doc when this happen to change the netpol ns as well so they stay in sync correct ?

stleerh:
Alternatively, we can leave out `namespace: netobserv`, but then you are expected to be in the right namespace when you apply this.  I think it's better to leave it in and document it.

config/policy/lockdown-netobserv.yaml
```diff
@@ -0,0 +1,21 @@
...
+  namespace: netobserv
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - podSelector: {}  # without namespaceSelecor, it defaults to 'same namespace'
+        - podSelector: {}
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: openshift-console
```
jotak, line 14: just wondering how safe is relying on `kubernetes.io/metadata.name` ie. can't a user put any label on a namespace they own, hence usurping another namespace name? Or is it somehow controlled and validated by the control plane ?
PS: I know netpols don't just simply work with namespace names directly, sadly

config/policy/lockdown-netobserv.yaml
memodi, line 1: @stleerh - any reason why it's in this repo, instead of [docs](https://github.com/netobserv/documents) repo?  I don't believe it's managed by NOO.

config/policy/lockdown-netobserv.yaml
stleerh: In Kubernetes, this is a [recommended label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#:~:text=Type-,app.kubernetes.io/name,-The%20name%20of).  However, it's pretty much standard.  If you create a project with `oc new-project {project}`, `kubectl create namespace {ns}`, or do this in Web Console, this label will be created.

stleerh:
Moving PR to `documents` repository.

config/policy/lockdown-netobserv.yaml
jotak: yeah my question was more if that's possible to hack & set that same label on another namespace that you have control over ... but I just did the test, editing `kubernetes.io/metadata.name` is immediately rollbacked to the actual namespace name

PR #341, opened 2023-05-23
fix CI manifest

jpinsonneau:
Manifest image needs to be pushed to created the shortlived image

### DIFF
--- a/.github/workflows/push_image_pr.yml
+++ b/.github/workflows/push_image_pr.yml
@@ -41,8 +41,10 @@ jobs:
         run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make image-build
       - name: push images
         run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make image-push
-      - name: build and push manifest
-        run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make ci-manifest
+      - name: build temp manifest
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -316,7 +316,7 @@ else
 endif
 
 .PHONY: ci-manifest-build
-ci-manifest-build: manifest-build ## Build CI manifest
+ci-manifest-build: manifest ## Build CI manifest
```
jotak, line 319: if "ci-manifest-build" runs "manifest" which itself does a push, then "ci-manifest-push" is now useless (or overlapping)?

jpinsonneau: but it works like this in other components:
- `make images` build and push manifest
- `make ci-manifest` overrides the manifest shortlived

Makefile
```diff
@@ -316,7 +316,7 @@ else
 endif
 
 .PHONY: ci-manifest-build
-ci-manifest-build: manifest-build ## Build CI manifest
+ci-manifest-build: ## Build CI manifest
```
msherif1234, line 319: do we need to do the same for all other repos too ?

jpinsonneau: :heavy_check_mark: yes !

.github/workflows/push_image_pr.yml
```diff
@@ -41,8 +41,10 @@ jobs:
         run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make image-build
       - name: push images
         run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make image-push
-      - name: build and push manifest
-        run: IMAGE_ORG=${{ env.ORG }} IMAGE=${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.short_sha }} make ci-manifest
+      - name: build temp manifest
```
jotak, line 44: I'm trying these steps locally, it fails because it looks for image tags like `temp-amd64`

jpinsonneau:
Replaced by https://github.com/netobserv/network-observability-operator/pull/342

PR #342, opened 2023-05-23
Attempt to fix multiarch workflow

jotak:
Shortlived images are now built from scratch, not from a base. (Same can be done later for bundle & catalog)

It simplifies the workflows especially with multiarch. We don't need anymore the "ci*" make targets.

Restore the SHA-builds on merged commits

Also make multiarch configurable, and only for merged commits, not per PR

### DIFF
--- a/.github/workflows/push_image.yml
+++ b/.github/workflows/push_image.yml
@@ -8,6 +8,7 @@ env:
   REGISTRY: quay.io/netobserv
   IMAGE: network-observability-operator
   ORG: netobserv
+  MULTIARCH_TARGETS: amd64 arm64 ppc64le
   VERSION: main
 
 jobs:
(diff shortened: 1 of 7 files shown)


jotak:
per-PR workflow tested here: https://github.com/netobserv/network-observability-operator/actions/runs/5059748527/jobs/9081725300?pr=343
This one seems good.
We'll need also to check post-merge workflow

jotak:
hold on ... there's something fishy ; the generated bundle seems wrong, its containerImage is odd

jotak:
next try: https://github.com/netobserv/network-observability-operator/pull/344

jotak:
Hey @msherif1234 - added two more commits:
- one to allow testing on a specific branch, `workflow-test`, that will also trigger the post-merge workflow
- a final one to document all of that

So I think I'm done with this PR, although (as said on slack) I've found new issues related to multi-arch - but that's not related to this PR, I think it can still be merged

jpinsonneau (approved):

PR #356, opened 2023-06-01
NETOBSERV-1005 dashboard: use sum...by

jotak:
- Fix metrics aggregations by using sum...by
- Do not split by FLP pod reporter as it's an implementation detail

1.3 backport: https://github.com/netobserv/network-observability-operator/pull/358

### DIFF
--- a/.mk/development.mk
+++ b/.mk/development.mk
@@ -151,7 +151,7 @@ set-agent-image:
 ifeq ("", "$(CSV)")
 	kubectl set env -n $(NAMESPACE) deployment netobserv-controller-manager -c "manager" RELATED_IMAGE_EBPF_AGENT=quay.io/$(USER)/netobserv-ebpf-agent:$(VERSION)
 else
-	./hack/swap-image-csv.sh $(CSV) $(OPERATOR_NS) ebpf_agent RELATED_IMAGE_EBPF_AGENT quay.io/$(USER)/netobserv-ebpf-agent:$(VERSION)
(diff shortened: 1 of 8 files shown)


OlivierCazade (approved):
LGTM

jpinsonneau (approved):

memodi:
@jotak this overall looks good to me. It would have been nice to have it limited to Top 5 with `topK` query but based on some reading it looks like when plotting w/ graphs that's not possible. Grafana has some features but probably OCP console Dashboards doesn't have it. 

Also, could we do something about those empty parentheses? quick search online I didn't find any straightforward solution to handle this in prometheus query. For our netflow-traffic page, we do that by replacing it with `non-pods` I guess that's done in console-plugin.

image

So if no changes are being further treat this as `/label qe-approved` and feel to free to merge it.

jotak:
@memodi there's probably something we can do with [label_replace](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace) , I'll make a try

jotak:
@memodi so my last commit should address your second comment. On the first one (the top5 thing), yeah I didn't find a way to force restricting 5 outputs, and anyway I thought having a bit more than 5 could be interesting, hence I just removed the "Top 5" mention from the title

jotak:
let me try one last thing: I'd like to remove the node-to-node or external-to-node traffic from the workloads-kind of metrics

jotak:
Ok, I've modified the "workloads" graph to exclude node-to-node traffic, so it's node overlapping with the other graph dedicated to nodes

Capture d’écran du 2023-06-02 12-06-07

wdyt?

memodi:
> wdyt?

I am yet to try, but this looks great :D , thank you!

jotak:
thanks!
/approve
There's the backporting PR then to /lgtm: https://github.com/netobserv/network-observability-operator/pull/358

PR #359, opened 2023-06-01
NETOBSERV-1076: fix flowcollector creation status

msherif1234:


pkg/helper/client_helper.go
```diff
@@ -29,7 +29,7 @@ func UnmanagedClient(cl client.Client) Client {
 // CreateOwned is an helper function that creates an object, sets owner reference and writes info & errors logs
 func (c *Client) CreateOwned(ctx context.Context, obj client.Object) error {
 	log := log.FromContext(ctx)
-	c.SetChanged(true)
+	c.SetChanged(false)
```
jotak, line 32: hmm .. that sounds wrong? When an object is being created, it means some reconciliation / change is happening

PR #326, opened 2023-04-17
NETOBSERV-773 Copy certificates across namespaces

jotak:
- Any certificate (secret/cm) can now be referenced from any namespace, which makes the operator watch the original and copy it to a target namespace. It allows not only to deploy Loki (or Kafka) in any namespace, but also fixes the issue of ebpf pods not having access to kafka CA/key without manual intervention

And quite a big refactoring:
- New "watchers" and "volumes" packages
- Creation of volumes is now using a builder-style approach allowing to incrementally add volumes and get at the same time their path. It avoid having discrepancies between mounted volumes and their related path reference.
- Watching certificates (or any CM/secret) now uses digest hash of content instead of metadata, to avoid triggering pods restart when a cm/secret was changed despite its content remaining the same
- Some things are moved around to make internal APIs easier to use, less parameters in functions, etc.
- New extensive integration tests on certificates management

PR #360, opened 2023-06-02
NETOBSERV-1076 Fix service change detection

jotak:
use DeepDerivative on labels & annotations only, rather than full ObjectMeta

### DIFF
--- a/pkg/helper/comparators.go
+++ b/pkg/helper/comparators.go
@@ -75,7 +75,8 @@ func probeChanged(old, new *corev1.Probe) bool {
 }
 
 func ServiceChanged(old, new *corev1.Service, report *ChangeReport) bool {
-	return report.Check("Service meta changed", !equality.Semantic.DeepDerivative(new.ObjectMeta, old.ObjectMeta)) ||
+	return report.Check("Service annotations changed", !equality.Semantic.DeepDerivative(new.Annotations, old.Annotations)) ||
+		report.Check("Service labels changed", !equality.Semantic.DeepDerivative(new.Labels, old.Labels)) ||
 		report.Check("Service spec changed", !equality.Semantic.DeepDerivative(new.Spec, old.Spec))
 }
 


pkg/helper/comparators.go
```diff
@@ -75,7 +75,8 @@ func probeChanged(old, new *corev1.Probe) bool {
 }
 
 func ServiceChanged(old, new *corev1.Service, report *ChangeReport) bool {
-	return report.Check("Service meta changed", !equality.Semantic.DeepDerivative(new.ObjectMeta, old.ObjectMeta)) ||
+	return report.Check("Service annotations changed", !equality.Semantic.DeepDerivative(new.Annotations, old.Annotations)) ||
+		report.Check("Service labels changed", !equality.Semantic.DeepDerivative(new.Labels, old.Labels)) ||
```
stleerh, line 79: What part of corev1.Service changed that triggered this to return true, or does it not matter because it's clear that only Annotations and Labels are relevant?  Why was this not happening before?

jotak: Many parts of the `ObjectMeta` will differ when comparing an existing object with a newly built object; e.g. are the uuid, creation date, etc. It's safe to rely only on labels & annotations as we aren't because that's at the moment the only things in `ObjectMeta` that we partly derive from configuration*. You can see an example here of a service that we create: https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/consoleplugin_objects.go#L300-L322

*: with the exception of the namespace, but a change of namespace triggers a whole different reconciliation process.

Why it happens only now? I wasn't sure... but now I remember something: in https://github.com/netobserv/network-observability-operator/pull/326, during the refactoring I fixed something related to the `DidChange` / `IsInProgress` stuff: previously it wasn't fully working in every situation because these were [fields of a structure that happened to be sometimes copied as argument](https://github.com/netobserv/network-observability-operator/pull/326/files#diff-7340f1cb3173df7c142997d58f7230bfd7aa1564f65be4ca3f88abe87e7d6a0dL19-L23) - which means that the "didChange" field of the parent was never updated when the one from the copy was. I fixed it by [making didChange/isInProgress variables now only created and owned from the parent, and updated via lambdas](https://github.com/netobserv/network-observability-operator/pull/326/files#diff-b1b7336af11d201e740a99a59ac20aa4845637c2b80f3eda38d1b03861ed74c4R102-R104).

So, tl;dr:  it *was* happening before, but without the same impact on the "Updating" CR status.

stleerh: Thanks for the detailed explanation.

PR #370, opened 2023-06-09
Doc updates:

jotak:
- Generalize using backticks for cross-references of spec fields, enum values, etc.
- Format enum options as list when their description isn't too short
- A few rephrasing (e.g. in conversation tracking)
- should -> must

### DIFF
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -37,62 +37,63 @@ const (
 // This is a workaround for the related issue:
 // https://github.com/kubernetes-sigs/controller-tools/issues/622
 
-// FlowCollectorSpec defines the desired state of FlowCollector.
+// Defines the desired state of the FlowCollector resource.
 // <br><br>
 // *: the mention of <i>"unsupported"</i>, or <i>"deprecated"</i> for a feature throughout this document means that this feature
-// is not officially supported by Red Hat. It may have been, for instance, contributed by the community
-// and accepted without a formal agreement for maintenance. The product maintainers may provide some support
+// is not officially supported by Red Hat. It might have been, for instance, contributed by the community
+// and accepted without a formal agreement for maintenance. The product maintainers might provide some support
 // for these features as a best effort only.
 type FlowCollectorSpec struct {
 	// Important: Run "make generate" to regenerate code after modifying this file
 
-	// namespace where NetObserv pods are deployed.
+	// Namespace where NetObserv pods are deployed.
 	// If empty, the namespace of the operator is going to be used.
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
 
-	// agent for flows extraction.
+	// Agent configuration for flows extraction.
 	// +kubebuilder:default:={type:"EBPF"}
 	Agent FlowCollectorAgent `json:"agent"`
 
-	// processor defines the settings of the component that receives the flows from the agent,
-	// enriches them, and forwards them to the Loki persistence layer.
+	// `processor` defines the settings of the component that receives the flows from the agent,
+	// enriches them, generates metrics, and forwards them to the Loki persistence layer and/or any available exporter.
 	Processor FlowCollectorFLP `json:"processor,omitempty"`
 
-	// loki, the flow store, client settings.
+	// Loki, the flow store, client settings.
 	Loki FlowCollectorLoki `json:"loki,omitempty"`
 
-	// consolePlugin defines the settings related to the OpenShift Console plugin, when available.
+	// `consolePlugin` defines the settings related to the OpenShift Console plugin, when available.
 	ConsolePlugin FlowCollectorConsolePlugin `json:"consolePlugin,omitempty"`
 
-	// deploymentModel defines the desired type of deployment for flow processing. Possible values are "DIRECT" (default) to make
-	// the flow processor listening directly from the agents, or "KAFKA" to make flows sent to a Kafka pipeline before consumption
-	// by the processor.
-	// Kafka can provide better scalability, resiliency and high availability (for more details, see https://www.redhat.com/en/topics/integration/what-is-apache-kafka).
+	// `deploymentModel` defines the desired type of deployment for flow processing. Possible values are:<br>
+	// - `DIRECT` (default) to make the flow processor listening directly from the agents.<br>
+	// - `KAFKA` to make flows sent to a Kafka pipeline before consumption by the processor.<br>
+	// Kafka can provide better scalability, resiliency, and high availability (for more details, see https://www.redhat.com/en/topics/integration/what-is-apache-kafka).
 	// +unionDiscriminator
 	// +kubebuilder:validation:Enum:="DIRECT";"KAFKA"
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default:=DIRECT
 	DeploymentModel string `json:"deploymentModel"`
 
-	// kafka configuration, allowing to use Kafka as a broker as part of the flow collection pipeline. Available when the "spec.deploymentModel" is "KAFKA".
+	// Kafka configuration, allowing to use Kafka as a broker as part of the flow collection pipeline. Available when the `spec.deploymentModel` is `KAFKA`.
 	// +optional
 	Kafka FlowCollectorKafka `json:"kafka,omitempty"`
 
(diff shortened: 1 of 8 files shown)


api/v1beta1/flowcollector_types.go
```diff
@@ -404,27 +406,30 @@ type FlowCollectorFLP struct {
...
 	// +optional
-	// conversation end timeout is the duration of time to wait from the last flow log to end a conversation
+	// conversation end timeout is the time to wait after a network flow is received, to consider the conversation ended.
+	// This delay is ignored when a FIN packet is collected for TCP flows (see `conversationTerminatingTimeout` instead).
 	ConversationEndTimeout *metav1.Duration `json:"conversationEndTimeout,omitempty"`
 
 	//+kubebuilder:default:="5s"
 	// +optional
-	// conversation terminating timeout is the duration of time to wait from detected FIN flag to end a connection
+	// conversation terminating timeout is the time to wait from detected FIN flag to end a conversation. Only relevant for TCP flows.
```
jotak, line 432: @jpinsonneau can you review in particular this part? Is this all correct and makes sense?

jotak:
@skrthomas I'll try also to fix something else in a next commit: our asciidoc generator doesn't handle very well the arrays kind of fields, e.g. "exporters" appears twice, first as "exporters | array" and then again as "exporters[] | object" - and only the description of the former is meant to be rendered. I'll see if I can fix that as a post-processing.

skrthomas (commented):
I meant to publish these comments on Friday @jotak !! I finished reviewing and have a few things I noticed throughout. I've incorporated these new updates into the product doc.

jotak:
thanks @skrthomas , all comments addressed

skrthomas:
@jotak I just got my peer review back and it seems like we have some more edits to make. Nothing big mostly just replacing may/might, some `` needed around parameters, commas, things like that.

api/v1beta1/flowcollector_types.go
jpinsonneau: Yes the frontend [translate the events as](https://github.com/netobserv/network-observability-console-plugin/blob/32dd09a3a03f3fc55c2e103ff051c84ade79cab7/web/src/utils/columns.ts#L813):
```
        switch (f.labels._RecordType) {
          case 'newConnection':
            return t('Conversation start');
          case 'heartbeat':
            return t('Conversation tick');
          case 'endConnection':
            return t('Conversation end');
          case 'flowLog':
          default:
            return t('Flow');
        }
```

api/v1beta1/flowcollector_types.go
```diff
@@ -485,26 +492,26 @@ type FlowCollectorLoki struct {
...
 	//+kubebuilder:default:="DISABLED"
-	// AuthToken describe the way to get a token to authenticate to Loki.
-	// DISABLED will not send any token with the request.
-	// HOST - <i>deprecated (*)</i> - will use the local pod service account to authenticate to Loki.
-	// FORWARD will forward the user token for authorization.
-	// When using the Loki Operator, this should be set to `FORWARD`.
+	// AuthToken describe the way to get a token to authenticate to Loki.<br>
+	// - `DISABLED` will not send any token with the request.<br>
+	// - `FORWARD` will forward the user token for authorization.<br>
+	// - `HOST` - <i>deprecated (*)</i> - will use the local pod service account to authenticate to Loki.<br>
```
jpinsonneau, line 513: same here: deprecated or unsupported for 1.3+ ?

jpinsonneau (approved):

api/v1beta1/flowcollector_types.go
```diff
@@ -89,24 +89,25 @@ type FlowCollectorSpec struct {
 // allow defining both fields.
 // +union
 type FlowCollectorAgent struct {
-	// type selects the flows tracing agent. Possible values are "EBPF" (default) to use NetObserv eBPF agent,
-	// "IPFIX" - <i>deprecated (*)</i> - to use the legacy IPFIX collector. "EBPF" is recommended in most cases as it offers better
-	// performances and should work regardless of the CNI installed on the cluster.
-	// "IPFIX" works with OVN-Kubernetes CNI (other CNIs could work if they support exporting IPFIX,
+	// type selects the flows tracing agent. Possible values are:<br>
+	// - `EBPF` (default) to use NetObserv eBPF agent.<br>
+	// - `IPFIX` - <i>deprecated (*)</i> - to use the legacy IPFIX collector.<br>
```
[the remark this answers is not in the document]
jotak: this one was ambiguous as it was deprecated before even our 1st GA ... but considering the history of the project upstream I would say it's deprecated, since it was once the only available option

api/v1beta1/flowcollector_types.go
jotak: Here I think it's clearly deprecated: it used to be a supported feature, and is not anymore

jotak:
@skrthomas I applied the feedback from downstream, with the exception of the backticks on the field name being described (this one: https://github.com/openshift/openshift-docs/pull/61006#discussion_r1229892462 ) : because it's a pattern we use almost everywhere in the file and not just at the place where it was suggested, if we want to do this change we should do it everywhere for consistency. So at the moment I haven't done the change ... but yes we could do it everywhere.

skrthomas:
@jotak OK, I recopied your file. If it isn't too much trouble to regenerate with the backticks for the field name descriptions, lets go ahead and do that.

jotak:
@skrthomas I'm not entirely sure yet how to proceed. We have 2 kinds of patterns:
1. With the field name being explicitly referenced and used as the subject of a verb like `is` or `defines`, e.g:
```
processor defines the settings of the component that receives the flows from the agent
```
In this case it seems legit to use the backticks: `processor` defines blah blah

2. More implicitly referencing the field name, e.g:
```
namespace where NetObserv pods are deployed
```
In this case it seems more "accidental" that the word is the same as the API field. This could be a problem for instance with this one:
```
Kafka configuration, allowing to use Kafka as a broker as part of the flow collection pipeline.
``` 
"Kafka" should take a capital K, unless it's referencing the field in which case perhaps we should reformulate using the 1. pattern:
```
`kafka` defines the configuration for Kafka, allowing to use it as a broker as part of the flow collection pipeline.
``` 
But it creates more repetitions so it's not pretty.

jotak:
@skrthomas my last two commits include generalization of backticks, and reintroducing some lost rephrasing that were done only downstream (as shown in the current diff https://github.com/openshift/openshift-docs/pull/61006/files )

skrthomas:
> @skrthomas I'm not entirely sure yet how to proceed. We have 2 kinds of patterns:
> 
>     1. With the field name being explicitly referenced and used as the subject of a verb like `is` or `defines`, e.g:
> 
> 
> ```
> processor defines the settings of the component that receives the flows from the agent
> ```
> 
> In this case it seems legit to use the backticks: `processor` defines blah blah
> 
>     2. More implicitly referencing the field name, e.g:
> 
> 
> ```
> namespace where NetObserv pods are deployed
> ```
> 
> In this case it seems more "accidental" that the word is the same as the API field. This could be a problem for instance with this one:
> 
> ```
> Kafka configuration, allowing to use Kafka as a broker as part of the flow collection pipeline.
> ```
> 
> "Kafka" should take a capital K, unless it's referencing the field in which case perhaps we should reformulate using the 1. pattern:
> 
> ```
> `kafka` defines the configuration for Kafka, allowing to use it as a broker as part of the flow collection pipeline.
> ```
> 
> But it creates more repetitions so it's not pretty.

@jotak I think that the implicit references, like scenario 2, could be reformulated to be more like scenario 1. For example, it could be `The `namespace` field specifies the namespace where NetObserv pods are deployed`.

For Kafka, Ah, ok thanks for clarifying the capitalization situational differences. And I agree, it can be reformulated also more like the 1 pattern.

skrthomas (approved):

docs/flowcollector-flows-netobserv-io-v1beta1.adoc
```diff
@@ -37,7 +37,7 @@ Type::
 | `object`
 | FlowCollectorSpec defines the desired state of FlowCollector.  +
```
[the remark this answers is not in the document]
jotak: I've removed the mention of FlowCollectorSpec, but there must be other places with the same issue (that's not new btw, it was already like this previously)

PR #372, opened 2023-06-12
Implements s3 exporter

jpinsonneau:
This PR implements S3 exporter using the already existing [s3 encode](https://github.com/netobserv/flowlogs-pipeline/pull/336) stage.

Here is a sample of generated files in the bucket:
[sample.txt](https://github.com/netobserv/network-observability-operator/files/11723616/sample.txt)

Rely on https://github.com/netobserv/flowlogs-pipeline/pull/442


api/v1beta1/flowcollector_types.go
```diff
@@ -255,6 +257,26 @@ type FlowCollectorIPFIXReceiver struct {
...
+	// address of the s3 server
+	Endpoint string `json:"endpoint"`
+
+	//+kubebuilder:default:=""
+	// username to connect to server
+	AccessKeyID string `json:"accessKeyId"`
+
+	//+kubebuilder:default:=""
+	// password to connect to server
+	SecretAccessKey string `json:"secretAccessKey"`
```
jpinsonneau, line 273: It would be better to load a secret here but it's useless for now as FLP doesn't.
In result, the credentials would be in clear text in the FLP configmap.

jotak: Well, I think we should read it from a secret/CM anyway: not because it makes it more hidden (you're right that it will be visible in FLP anyway - which is ok as long as FLP is only accessible for cluster admins) , but in a gitops perspective users would like to decouple a config such as FlowCollector CR, and secrets management, for instance by using stuff like Vault: so that the FlowCollector CR can be safely stored in config repos (e.g. in git) with compromising secrets, and secrets are generated/injected at runtime.

I have a commit for Kafka / SASL that I haven't pushed as a PR yet (I should do that) where there's a similar problem to solve: https://github.com/jotak/network-observability-operator/commit/17a2ae96074b05cfbab57c3a9b66624e086a864a - maybe I should push it and we can align our 2 PRs & reuse some structs / functions

api/v1beta1/flowcollector_types.go
```diff
@@ -716,13 +747,14 @@ type ExporterType string
 const (
 	KafkaExporter ExporterType = "KAFKA"
 	IpfixExporter ExporterType = "IPFIX"
+	S3Exporter    ExporterType = "S3"
 )
 
 // FlowCollectorExporter defines an additional exporter to send enriched flows to.
 type FlowCollectorExporter struct {
-	// type selects the type of exporters. The available options are "KAFKA" and "IPFIX". "IPFIX" is <i>unsupported (*)</i>.
+	// type selects the type of exporters. The available options are "KAFKA", "IPFIX" and "S3". "IPFIX" is <i>unsupported (*)</i>.
```
jotak, line 755: I guess S3 would be unsupported too, or is there a request to support it?

api/v1beta1/flowcollector_types.go
jotak: (edit: rebased my sasl branch, new one is https://github.com/jotak/network-observability-operator/commit/3a4f105afb19ce39ac43410ff22ddbacda2f232f )

jpinsonneau: I was considering replacing these creds by a secret and mounting the secret inside FLP. 
Do you feel it's not necessary here ?

jotak: yes it can be done that way but I guess then you need to update FLP to read from file?
That's also what I did for sasl: https://github.com/jotak/network-observability-operator/commit/3a4f105afb19ce39ac43410ff22ddbacda2f232f#diff-adf3c77f719dd339bbc5c11eeb3e846bff5024d8f79a4f4461fa914743cb9c50R294-R312

jpinsonneau: yeah I'm fine updating FLP to add this capability if we are aligned

jpinsonneau:
Closing this as it's quite outdated

PR #374, opened 2023-06-15
NETOBSERV-1076 Fix CR stuck in Updating state

jotak:
The problem was that copying certificates across namespaces was done on every reconciliation call, with or without digest change. This is not a critical issue as it would only trigger a no-op update, but it was screwing up our "updating" state detection.

There's actually two changes in this commit:
1. Fixing the "Updating" flag, so that it's not set when an update is actually performed (we need to fetch back the object (from cache) and check if the resource version was actually modified)
2. Fixing the certificate watching process to avoid unnecessary tries to update

Also:
- Add / fix tests
- Modify some logs to make important information more visible
- Makefile change for building manifest without error (unrelated)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,13 +303,10 @@ image-push: ## Push MULTIARCH_TARGETS images
 .PHONY: manifest-build
 manifest-build: ## Build MULTIARCH_TARGETS manifest
 	@echo 'building manifest $(IMAGE)'
-ifeq (${OCI_BIN}, docker)
-	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE} $(foreach target,$(MULTIARCH_TARGETS), --amend ${IMAGE}-$(target));
-else
+	DOCKER_BUILDKIT=1 $(OCI_BIN) rmi ${IMAGE}; \
+	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE}
 	trap 'exit' INT; \
-	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE} ||:
 	$(foreach target,$(MULTIARCH_TARGETS),$(call manifest_add_target,$(target)))
-endif
 
 .PHONY: manifest-push
 manifest-push: ## Push MULTIARCH_TARGETS manifest
--- a/controllers/reconcilers/namespaced_objects_manager.go
+++ b/controllers/reconcilers/namespaced_objects_manager.go
@@ -68,7 +68,12 @@ func (m *NamespacedObjectManager) FetchAll(ctx context.Context) error {
 			// On success, placeholder is filled with resource. Caller should keep a pointer to it.
 		}
 	}
-	log.Info("Fetched: " + strings.Join(fetched, ",") + ". Not found: " + strings.Join(notFound, ","))
+	if len(fetched) > 0 {
+		log.Info("FETCHED: " + strings.Join(fetched, ","))
+	}
+	if len(notFound) > 0 {
+		log.Info("(Items not deployed: " + strings.Join(notFound, ",") + ")")
+	}
 	return nil
 }
 
@@ -83,7 +88,7 @@ func (m *NamespacedObjectManager) cleanup(ctx context.Context, namespace string)
 		ref := obj.placeholder.DeepCopyObject().(client.Object)
 		ref.SetName(obj.name)
 		ref.SetNamespace(namespace)
-		log.Info("Deleting old "+obj.kind, "Namespace", namespace, "Name", obj.name)
+		log.Info("DELETING "+obj.kind, "Namespace", namespace, "Name", obj.name)
 		err := m.client.Delete(ctx, ref)
 		if client.IgnoreNotFound(err) != nil {
 			log.Error(err, "Failed to delete old "+obj.kind, "Namespace", namespace, "Name", obj.name)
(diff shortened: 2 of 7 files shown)


jotak:
Backport: https://github.com/netobserv/network-observability-operator/pull/375

Makefile
```diff
@@ -303,13 +303,10 @@ image-push: ## Push MULTIARCH_TARGETS images
 .PHONY: manifest-build
 manifest-build: ## Build MULTIARCH_TARGETS manifest
 	@echo 'building manifest $(IMAGE)'
-ifeq (${OCI_BIN}, docker)
-	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE} $(foreach target,$(MULTIARCH_TARGETS), --amend ${IMAGE}-$(target));
-else
+	DOCKER_BUILDKIT=1 $(OCI_BIN) rmi ${IMAGE}; \
```
msherif1234, line 306: maybe use `-f` in case the image doesn't exits so it won't fail ?

Makefile
```diff
@@ -303,13 +303,10 @@ image-push: ## Push MULTIARCH_TARGETS images
...
+	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE}
```
msherif1234, line 307: hmm why changed this logic since docker has `--amend` support its better to use it than the work around we have for the older podman version that we are stuck with ?

controllers/reconcilers/namespaced_objects_manager.go
```diff
@@ -103,7 +108,7 @@ func (m *NamespacedObjectManager) TryDelete(ctx context.Context, obj client.Obje
 	if m.Exists(obj) {
 		log := log.FromContext(ctx)
 		kind := reflect.TypeOf(obj).String()
-		log.Info("Deleting old "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
+		log.Info("DELETING "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
```
msherif1234, line 111: nit: I don't know if there is a need to have all caps here I don't think this align with go coding style ?

pkg/helper/client_helper.go
```diff
@@ -58,12 +57,22 @@ func (c *Client) UpdateOwned(ctx context.Context, old, obj client.Object) error
...
 	if err != nil {
 		log.Error(err, "Failed to update "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
 		return err
 	}
+	err = c.Get(ctx, client.ObjectKeyFromObject(obj), obj)
+	if err != nil {
+		log.Error(err, "Failed to get updated resource "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
+		return err
+	}
+	if obj.GetResourceVersion() != old.GetResourceVersion() {
```
msherif1234, line 71: what about using deep.Equal(old, obj) to know if it was updated or not ?

Makefile
jotak: looks like -i is a better fit for that?
```
Options:
  -a, --all        Remove all images
  -f, --force      Force Removal of the image
  -i, --ignore     Ignore errors if a specified image does not exist
      --no-prune   Do not remove dangling images
```

@msherif1234  can you check if it's the same with docker ?

msherif1234:
```sh
 docker rmi --help

Usage:  docker rmi [OPTIONS] IMAGE [IMAGE...]

Remove one or more images

Options:
  -f, --force      Force removal of the image
      --no-prune   Do not delete untagged parents
```

Makefile
jotak: it's to avoid having multiple ways of doing things as much as we can ? I think when we can run the same script regardless of the environment, we should do that, rather than having different procedures for different environments?
Moreover the intent of this make target is really to create a fresh new manifest, not updating an existing one, so it seems to me that even with docker it makes sense to create a fresh one? For instance, what happens if you build an image for arm64 only and then, later, build an image for amd64 only? I think with the current script you would get the two of them in the manifest, which isn't what you asked for

controllers/reconcilers/namespaced_objects_manager.go
jotak: The intent is to make some important information more visible - I found it very useful when debugging, whereas with the normal sentence case it's sometimes hard to find relevant logs at a glance. That's really for pragmatism.

Makefile
msherif1234: we need to check on this I think we should endup with one manifest for the example u have at the end, my concern here what I did for podman was more of a workaround because it lack --amend support and now you proposing using the workaround for both so I see this as step backward while if we have the latest podman we would just do --amend for both

Makefile
jotak: seems like it's not for docker https://docs.docker.com/engine/reference/commandline/rmi/
so let's go with -f

Makefile
```diff
@@ -303,13 +303,8 @@ image-push: ## Push MULTIARCH_TARGETS images
 .PHONY: manifest-build
 manifest-build: ## Build MULTIARCH_TARGETS manifest
 	@echo 'building manifest $(IMAGE)'
-ifeq (${OCI_BIN}, docker)
+	DOCKER_BUILDKIT=1 $(OCI_BIN) rmi ${IMAGE} -f
 	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest create ${IMAGE} $(foreach target,$(MULTIARCH_TARGETS), --amend ${IMAGE}-$(target));
```
msherif1234, line 307: this will break on podman since it doesn't have `--amend` but since we default to docker at least out CI should be fine

msherif1234 (commented):
/LGTM

Makefile
jotak: podman does have `--amend` : https://docs.podman.io/en/latest/markdown/podman-manifest-create.1.html#options
It was only the outdated version shipped with the github CI image that didn't have it, so it was only a CI issue, but on our dev machines we are fine with a recent podman. FI I tested on my machine with podman.

msherif1234: https://github.com/containers/podman/releases/tag/v4.3.0 or later

OlivierCazade (approved):

PR #375, opened 2023-06-15
NETOBSERV-1076 Fix CR stuck in Updating state [1.3 backport]

jotak:
Backport of https://github.com/netobserv/network-observability-operator/pull/374

The problem was that copying certificates across namespaces was done on every reconciliation call, with or without digest change. This is not a critical issue as it would only trigger a no-op update, but it was screwing up our "updating" state detection.

There's actually two changes in this commit:
1. Fixing the "Updating" flag, so that it's not set when an update is actually performed (we need to fetch back the object (from cache) and check if the resource version was actually modified)
2. Fixing the certificate watching process to avoid unnecessary tries to update

Also:
- Add / fix tests
- Modify some logs to make important information more visible
- Makefile change for building manifest without error (unrelated)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,13 +303,10 @@ image-push: ## Push MULTIARCH_TARGETS images
 .PHONY: manifest-build
 manifest-build: ## Build MULTIARCH_TARGETS manifest
 	@echo 'building manifest $(IMAGE)'
-ifeq (${OCI_BIN}, docker)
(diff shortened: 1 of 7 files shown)


OlivierCazade (approved):

jotak:
@Amoghrd I think we didn't backport to 1.3 branch the various fixes that we did regarding the ok-to-test workflow ... so maybe it can be verified post-merge, on the downstream build directly? (given you verified already the PR on `main`)

PR #377, opened 2023-06-20
NETOBSERV-974: Add SASL support for KAFKA (unsupported feature)

jotak:


api/v1alpha1/flowcollector_types.go
```diff
@@ -228,10 +228,13 @@ type FlowCollectorKafka struct {
 	// kafka topic to use. It must exist, NetObserv will not create it.
 	Topic string `json:"topic"`
 
-	// tls client configuration. When using TLS, verify that the address matches the Kafka port used for TLS, generally 9093.
-	// Note that, when eBPF agents are used, Kafka certificate needs to be copied in the agent namespace (by default it's netobserv-privileged).
+	// TLS client configuration. When using TLS, verify that the address matches the Kafka port used for TLS, generally 9093.
 	// +optional
 	TLS ClientTLS `json:"tls"`
+
+	// SASL authentication configuration.
```
jotak, line 235: We might want to say "unsupported" here.

jotak: (note that we support strimzi / amq streams, and strimzi / amq streams supports sasl, so idk if supporting sasl should transitively apply to us)

jotak:
For info, I found a bug in the agent reconciler fixed in the last commit, not specific to this PR - if for some reason we delay merging this PR, I would need to open another one (but given it's marked as "unsupported" I don't think there's reasons to not merge it?)

jpinsonneau (approved):
code LGTM, thanks @jotak

PR #383, opened 2023-06-22
(trivial) makefile: avoid kubectl calls in global scope

jotak:
This kubectl call was ran everytime any make target was invoked, even something not requiring a cluster, resulting in errors or sometimes making any "make" call hanging (e.g. when kubectl points to an unreachable server / behind vpn)


jpinsonneau (approved):

jotak:
/approve
thanks @jpinsonneau

PR #385, opened 2023-06-29
Add nil-check for loki.enable

jotak:


controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -439,7 +439,7 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
 	})
 
 	// loki stage (write) configuration
-	if *b.desired.Loki.Enable {
+	if helper.PtrBool(b.desired.Loki.Enable) {
```
jotak, line 442: I'm not exactly sure when Enable can be nil (upgrade scenario?) - but we should actually consider nil as true, to keep consistency with the default behaviour. So I'll write a new helper method instead

jotak: (done)

jpinsonneau (approved):
thanks @jotak !

PR #386, opened 2023-06-29
NETOBSERV-1182 added cluster name to flp configuration

KalmanMeth:
Added transform_filter to flp pipeline to add a clusterName field to flow logs.
clusterName is taken from config file. If blank, take clusterID from clusterVersion struct.

### DIFF
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -448,6 +448,11 @@ type FlowCollectorFLP struct {
 	// `conversationTerminatingTimeout` is the time to wait from detected FIN flag to end a conversation. Only relevant for TCP flows.
 	ConversationTerminatingTimeout *metav1.Duration `json:"conversationTerminatingTimeout,omitempty"`
 
+	//+kubebuilder:default:=""
+	// +optional
+	// `clusterName` is the name of the cluster to appear in the flows data. This is useful in a multi-cluster context. When using OpenShift, leave empty to make it automatically determined.
+	ClusterName string `json:"clusterName,omitempty"`
+
 	// `debug` allows setting some aspects of the internal configuration of the flow processor.
 	// This section is aimed exclusively for debugging and fine-grained performance optimizations,
 	// such as GOGC and GOMAXPROCS env vars. Users setting its values do it at their own risk.
--- a/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
@@ -517,6 +517,14 @@ spec:
           - update
           - watch
         - apiGroups:
+          - config.openshift.io
+          resources:
+          - clusterversions
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
           - console.openshift.io
           resources:
           - consoleplugins
@@ -614,24 +622,15 @@ spec:
           - rbac.authorization.k8s.io
           resources:
           - clusterrolebindings
-          - rolebindings
-          verbs:
-          - create
-          - delete
-          - get
-          - list
-          - update
-          - watch
-        - apiGroups:
-          - rbac.authorization.k8s.io
-          resources:
           - clusterroles
+          - rolebindings
           - roles
           verbs:
           - create
           - delete
           - get
           - list
+          - update
           - watch
         - apiGroups:
           - security.openshift.io
--- a/config/samples/flows_v1beta1_flowcollector.yaml
+++ b/config/samples/flows_v1beta1_flowcollector.yaml
@@ -58,6 +58,8 @@ spec:
     conversationTerminatingTimeout: 5s
     conversationHeartbeatInterval: 30s
     conversationEndTimeout: 10s
+    # Append a unique cluster name to each record
+    # clusterName: <CLUSTER NAME>
   kafka:
     address: "kafka-cluster-kafka-bootstrap.netobserv"
     topic: network-flows
--- a/controllers/flowcollector_controller.go
+++ b/controllers/flowcollector_controller.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/netobserv/network-observability-operator/controllers/globals"
+	configv1 "github.com/openshift/api/config/v1"
 	osv1alpha1 "github.com/openshift/api/console/v1alpha1"
 	securityv1 "github.com/openshift/api/security/v1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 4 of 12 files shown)


KalmanMeth:
https://issues.redhat.com/browse/NETOBSERV-245

api/v1beta1/flowcollector_types.go
```diff
@@ -436,6 +436,11 @@ type FlowCollectorFLP struct {
 	// `conversationTerminatingTimeout` is the time to wait from detected FIN flag to end a conversation. Only relevant for TCP flows.
 	ConversationTerminatingTimeout *metav1.Duration `json:"conversationTerminatingTimeout,omitempty"`
 
+	//+kubebuilder:default:="defaultCluster"
+	// +optional
+	// `clusterName` is the name of the cluster.
+	ClusterName string `json:"clusterName,omitempty"`
+
```
jpinsonneau, line 443: For Openshift we can automatically retreive `ClusterID` from api
https://pkg.go.dev/github.com/openshift/api@v0.0.0-20230707160225-81d582da354b/config/v1#ClusterID

https://access.redhat.com/solutions/5280291

jpinsonneau:
```suggestion
	//+kubebuilder:default:=""
	// +optional
	// `clusterName` is the name of the cluster.
	ClusterName string `json:"clusterName,omitempty"`

```
I would suggest to set default as empty to avoid FLP extra stage for single cluster usage. WDYT ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -45,6 +45,7 @@ const (
 	conntrackTerminatingTimeout = 5 * time.Second
 	conntrackEndTimeout         = 10 * time.Second
 	conntrackHeartbeatInterval  = 30 * time.Second
+	clusterNameDefault          = "defaultCluster"
```
jpinsonneau, line 48:
```suggestion
```

api/v1beta1/flowcollector_types.go
KalmanMeth: If we automatically retrieve clusterID from openshift when clusterName is blank, we really cannot distinguish between single cluster vs multi-cluster. To distinguish between them, we would have to add an explicit parameter to the flowcollector config file.

controllers/flowlogspipeline/flp_common_objects.go
KalmanMeth: Changed default to blank, but in that case we take the clusterName from Openshift.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -351,7 +351,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring
     console.openshift.io/plugins: '["netobserv-plugin"]'
-    containerImage: quay.io/netobserv/network-observability-operator:1.0.3
+    containerImage: quay.io/meth/network-observability-operator:main
```
jpinsonneau, line 354: you can run `make update-bundle` to avoid this :smile_cat: 
https://github.com/netobserv/network-observability-operator/blob/84c7fe8057a58d9d5d23c2ea3ab92d83499acc6c/Makefile#L325-L327

config/samples/flows_v1beta1_flowcollector.yaml
```diff
@@ -49,6 +49,7 @@ spec:
     conversationTerminatingTimeout: 5s
     conversationHeartbeatInterval: 30s
     conversationEndTimeout: 10s
+    clusterName:
```
jpinsonneau, line 52:
```suggestion
```

```suggestion
    # Append a unique cluster name to each record
    # clusterName: <CLUSTER NAME>
```

If the default is empty, then we can remove it from the sample or add a comment as example

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
KalmanMeth: fixed

api/v1beta1/flowcollector_types.go
KalmanMeth: Default set to empty.

api/v1beta1/flowcollector_types.go
```diff
@@ -436,6 +436,11 @@ type FlowCollectorFLP struct {
 	// `conversationTerminatingTimeout` is the time to wait from detected FIN flag to end a conversation. Only relevant for TCP flows.
 	ConversationTerminatingTimeout *metav1.Duration `json:"conversationTerminatingTimeout,omitempty"`
 
+	//+kubebuilder:default:=""
+	// +optional
+	// `clusterName` is the name of the cluster.
```
jotak, line 441: Since this is a user-facing doc, it deserves some explanation, example:
 
```suggestion
	// `clusterName` is the name of the cluster to appear in the flows data. This is useful in a multi-cluster context. When using OpenShift, leave empty to make it automatically determined.
```

controllers/flowcollector_controller.go
```diff
@@ -103,6 +104,15 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	reconcilersInfo := r.newCommonInfo(ctx, desired, ns, previousNamespace, func(b bool) { didChange = b }, func(b bool) { isInProgress = b })
 
+	// obtain default cluster ID
+	cversion := &configv1.ClusterVersion{}
+	key := client.ObjectKey{Name: "version"}
+	if err := r.Client.Get(ctx, key, cversion); err != nil {
+		log.Error(err, "unable to obtain cluster ID")
+	} else {
+		constants.DefaultClusterID = cversion.Spec.ClusterID
+	}
```
jotak, line 114: This block should be run only once, and only for openshift. To do so you can wrap it in a block  `if r.permissions.Vendor(ctx) == discover.VendorOpenShift && constants.DefaultClusterID == "" {`

jotak: Also it is unexpected to update a "constant". I would move `DefaultClusterID` somewhere else. For instance, it could be part of the `FlowCollectorReconciler` struct, and passed down to reconcilers via `newCommonInfo`; or if you prefer to keep a global access, maybe create a new file `globals.go` in package `reconcilers` ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -310,15 +310,32 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
+	var clusterName string
+	if b.desired.Processor.ClusterName != "" {
+		clusterName = b.desired.Processor.ClusterName
+	} else {
+		//take clustername from openshift
+		clusterName = string(constants.DefaultClusterID)
+	}
+	transformFilterRules := []api.TransformFilterRule{
+		{
+			Input: "clusterName",
```
jotak, line 322: We should use the same pattern as the other fields, ie:

```suggestion
			Input: "K8S_ClusterName",
```

controllers/flowcollector_controller.go
KalmanMeth: In order to pass down the clusterName through the series of function calls it would have required many changes and would have made the code ugly. I therefore made it into a global, as suggested.

api/v1beta1/flowcollector_types.go
KalmanMeth: added

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -310,15 +310,32 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
 	indexFields := constants.LokiIndexFields
 
+	var clusterName string
+	if b.desired.Processor.ClusterName != "" {
+		clusterName = b.desired.Processor.ClusterName
+	} else {
+		//take clustername from openshift
+		clusterName = string(constants.DefaultClusterID)
+	}
+	transformFilterRules := []api.TransformFilterRule{
```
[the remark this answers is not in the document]
KalmanMeth: done

controllers/flowlogspipeline/flp_common_objects.go
KalmanMeth: done

controllers/flowcollector_controller.go
jotak: That's ok
The "CommonInfo" is already done in that purpose, to be passed to the reconcilers, so you wouldn't have to change all the function signatures, just adding it to a struct that is already spread across reconcilers.
But that's ok with the global approach - as long as not everything ends up like that, which would denote a lack of modularity and makes sometimes testing more difficult

jotak:
/lgtm
thanks @KalmanMeth !
There's one pending question, that we can answer later: should we define K8S_ClusterName as a Loki index? I think we should, since it will become a common aggregating key for queries in a multi-cluster scenario. We can do that later.

PS : looks like the linter is unhappy due to high cyclomatic ...

KalmanMeth:
> PS : looks like the linter is unhappy due to high cyclomatic ...

I refactored a bit to make the linter happy.

jotak:
Thanks !
/lgtm

mffiedler:
@KalmanMeth @jotak Can we re-add ok-to-test?

KalmanMeth:
performed rebase

mffiedler:
/label qe-approved

Basic testing looks good, can continue QE post-merge

PR #387, opened 2023-06-29
NETOBSERV-1131: Filter for Duplicate=false in metrics

jotak:
Requires FLP PR: https://github.com/netobserv/flowlogs-pipeline/pull/448

Some observations when comparing byte rates from the plugin versus from prom metrics, e.g. between two nodes, e.g. with a promQL like that: 
```
sum(rate(netobserv_node_ingress_bytes_total{DstK8S_HostName="ip-10-0-137-27.eu-west-3.compute.internal"}[90s])) by (SrcK8S_HostName, DstK8S_HostName)
```
And on plugin's side, similarly, filtering on destination node name="ip-10-0-137-27.eu-west-3.compute.internal" , using Scope=Node and observing the top 5 flow rates in Overview:

Before this PR I see increased values, on prometheus side, in a range going from +20% to +80%
After this PR I still see increased values on prom side but seems much more reasonable, from +10% to +25%
([data](https://docs.google.com/spreadsheets/d/1LM7VCmiljdkpDSXDGpnlYLi_ajVumidD9v7eKb8wKLI/edit#gid=0))

I don't have a definitive explanation as to why I continue to see increased values, but I can make the hypothesis that they come from differences in rate calculation between loki and prometheus - also, the metrics view in the console provide a better resolution than our overview charts, which might play a role in the discrepancies. Anyway, I think this PR brings back values in a much more acceptable range.

### DIFF
--- a/controllers/flowlogspipeline/metrics_definitions/node_egress_bytes_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/node_egress_bytes_total.yaml
@@ -16,9 +16,11 @@ encode:
       - name: node_egress_bytes_total
         type: counter
         valuekey: Bytes
-        filter:
-          key: FlowDirection
+        filters:
+        - key: FlowDirection
           value: "1"
+        - key: Duplicate
+          value: "false"
         labels:
           - SrcK8S_HostName
           - DstK8S_HostName
--- a/controllers/flowlogspipeline/metrics_definitions/node_ingress_bytes_total.yaml
+++ b/controllers/flowlogspipeline/metrics_definitions/node_ingress_bytes_total.yaml
@@ -16,9 +16,11 @@ encode:
       - name: node_ingress_bytes_total
         type: counter
         valuekey: Bytes
-        filter:
-          key: FlowDirection
+        filters:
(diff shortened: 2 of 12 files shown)


jpinsonneau (approved):
LGTM, thanks @jotak 

Is it worth having an extra metric from `namespace_flows_total` with the `duplicate: false` filter ?

jotak:
> Is it worth having an extra metric from `namespace_flows_total` with the `duplicate: false` filter ?

That raises an interesting point, personally I see this metric more as a Health metric than a Network insight metric: because it tells about flows and not about bytes or packets, which is an arbitrary measure for netobserv. If it's there for health reasons, filtering out duplicate doesn't make sense IMO (it's an indicator of the "load" that netobserv components deal with, answering the question: which namespaces are the most responsible of what netobserv is enduring? - removing duplicates would hide some of that load).

But that said, it's part of our NetObserv dashboard, not of NetObserv Health - which perhaps is a mistake ?

KalmanMeth (dismissed):

KalmanMeth (approved):

PR #388, opened 2023-07-03
NETOBSERV-1075 Dashboard feedback: split infra / applications

jpinsonneau:
- moved to embedded json approach instead of FLP for more flexibility
- splitted application / infrastructures namespace & workload graphs
- grouped by sections

image
image
image


controllers/flowcollector_controller.go
```diff
@@ -309,6 +309,12 @@ func (r *FlowCollectorReconciler) reconcileOperator(ctx context.Context, cmn *re
 			return err
 		}
 	}
+
+	desiredFlowDashboardCM := buildFlowMetricsDashboard()
+	if err := cmn.ReconcileConfigMap(ctx, desiredFlowDashboardCM); err != nil {
+		return err
+	}
```
jotak, line 316: I think it makes sense to move that under the same condition as the health dashboard, in `if r.availableAPIs.HasSvcMonitor()` below? Is there a reason to not do that?

controllers/flowcollector_controller_test.go
```diff
@@ -326,14 +323,6 @@ func flowCollectorControllerSpecs() {
 				Expect(ds.Spec.Template.Spec.Tolerations).
 					To(ContainElement(v1.Toleration{Operator: v1.TolerationOpExists}))
 			})
-
-			By("Expecting the monitoring dashboards configmap to be deleted")
```
jotak, line 330: I didn't realize that, but I guess that's the drawback of not using FLP: we loose the ability to tie dashboards with metrics presence. Hence this test not passing, I guess. Can we think about a simple alternative for that? Maybe it requires some discussion?

jpinsonneau: done in https://github.com/netobserv/network-observability-operator/pull/388/commits/71da325c421290cf515e8ef6aed34b6b60db0577 with other enhancements described below

jpinsonneau:
Updated this PRs to address all feedback gathered:
- tie metrics to dashboards using tags. Check [FilterDashboardRows](https://github.com/netobserv/network-observability-operator/pull/388/files#diff-72c617dc46a4e4b487ccbed0fd1a1c5a21ec6b9bf0c7cda484076fb35fe13d5e) helper
- added missing metrics definitions
- added / updated missing graphs in dashboards
  -  moved namespace flows to health dashboard
- added a new `health` flag (default) to ignore extra flows graphs per node / namespace / workload

Here are examples of dashboards with no filters:

image
image

api/v1beta1/flowcollector_types.go
```diff
@@ -318,8 +318,8 @@ type FLPMetrics struct {
 	Server MetricsServerConfig `json:"server,omitempty"`
 
 	// `ignoreTags` is a list of tags to specify which metrics to ignore. Each metric is associated with a list of tags. More details in https://github.com/netobserv/network-observability-operator/tree/main/controllers/flowlogspipeline/metrics_definitions .
-	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`.
-	//+kubebuilder:default:={"egress","packets"}
+	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `health`.
+	//+kubebuilder:default:={"egress","packets", "health"}
```
jotak, line 322: this would ignore health by default, this isn't what we want?

jpinsonneau: Maybe a better naming can be found ?

The metrics flagged with `health` are only flows `metrics definitions` per nodes / namespaces / workloads

The health dashboard will look like as before by default

jotak: (or health is not what I thought it is )

pkg/helper/dashboard.go
```diff
@@ -0,0 +1,47 @@
+package helper
+
+import "encoding/json"
+
+func FilterDashboardRows(dashboard string, ignoreFlags []string) (string, error) {
```
jotak, line 5: So, if I'm correct, tags are duplicated: in every dashboard row (for dashboards filtering) and in metrics definitions (for FLP confgen). I guess it's ok for now but I think it could be error prone, especially as the dashboard jsons are very noisy and hard to review. So maybe we can think about how would could "secure" this, either for right now or for a follow-up. I can imagine:
- modifying FLP confgen so that it gives us the list of excluded and included metrics; then based on that, we would do the dashboards filtering based on a grep (Contains) on the promql expression
- or do the same thing without confgen, but it means the operator parsing the metrics definition files
- or, since those dashboards are statically defined, we could also just write a unit test that validates that the dashboard tags correctly match the metric defs tags

wdyt?

jotak: btw speaking of unit test I think it would be good to have one here that ensures the dashboard files are correctly parsed without errors

controllers/infra_health_dashboard.json
```diff
@@ -1,641 +1,1083 @@
...
+            "repeatIteration": null,
+            "repeatRowId": null,
+            "showTitle": true,
+            "tags": [
+                "flows"
+            ],
+            "title": "Flows",
+            "titleSize": "h6"
+        },
+        {
```
jotak, line 130: If you don't mind me hijacking your PR :-)
I would very well see this kind of addition:
(this is a query that shows the "flows overhead" generated by netobserv itself compared with the total cluster flows. In other words, a way to see if netobserv is triggering a "runaway reaction" as we've already seen in the past, e.g. when we set too small batch sizes.


```suggestion
        },
        {
            "collapse": false,
            "height": "250px",
            "panels": [
                {
                    "aliasColors": {},
                    "bars": false,
                    "dashLength": 10,
                    "dashes": false,
                    "datasource": "prometheus",
                    "fill": 1,
                    "fillGradient": 0,
                    "gridPos": {
                        "h": 20,
                        "w": 25,
                        "x": 0,
                        "y": 0
                    },
                    "id": 3,
                    "legend": {
                        "alignAsTable": false,
                        "avg": false,
                        "current": false,
                        "max": false,
                        "min": false,
                        "rightSide": false,
                        "show": true,
                        "sideWidth": null,
                        "total": false,
                        "values": false
                    },
                    "lines": true,
                    "linewidth": 1,
                    "links": [],
                    "nullPointMode": "null",
                    "percentage": false,
                    "pointradius": 5,
                    "points": false,
                    "renderer": "flot",
                    "repeat": null,
                    "seriesOverrides": [],
                    "spaceLength": 10,
                    "stack": false,
                    "steppedLine": false,
                    "targets": [
                        {
                            "expr": "100 * sum(rate(netobserv_namespace_flows_total{SrcK8S_Namespace='netobserv'}[1m]) or rate(netobserv_namespace_flows_total{SrcK8S_Namespace!='netobserv',DstK8S_Namespace=~'netobserv'}[1m])) / sum(rate(netobserv_namespace_flows_total[1m]))",
                            "format": "time_series",
                            "intervalFactor": 2,
                            "legendFormat": "% overhead",
                            "refId": "A"
                        }
                    ],
                    "thresholds": [],
                    "timeFrom": null,
                    "timeShift": null,
                    "title": "Percentage of flows generated by NetObserv own traffic",
                    "tooltip": {
                        "shared": true,
                        "sort": 0,
                        "value_type": "individual"
                    },
                    "type": "graph",
                    "xaxis": {
                        "buckets": null,
                        "mode": "time",
                        "name": null,
                        "show": true,
                        "values": []
                    },
                    "yaxes": [
                        {
                            "format": "short",
                            "label": null,
                            "logBase": 1,
                            "max": null,
                            "min": null,
                            "show": true
                        },
                        {
                            "format": "short",
                            "label": null,
                            "logBase": 1,
                            "max": null,
                            "min": null,
                            "show": true
                        }
                    ]
                }
            ],
            "repeat": null,
            "repeatIteration": null,
            "repeatRowId": null,
            "showTitle": true,
            "tags": [
                "flows",
                "overhead",
                "flows-overhead"
            ],
            "title": "Flows Overhead",
            "titleSize": "h6"
        },
        {
```

Of course I can also open a separate PR

api/v1beta1/flowcollector_types.go
jotak: It seems like, at least for now, "health" fully overlaps with "flows" so we could keep just "flows"

controllers/infra_health_dashboard.json
jpinsonneau: Let's have it ! Thanks

controllers/flowlogspipeline/metrics_definitions/namespace_flows_total.yaml
```diff
@@ -8,6 +8,7 @@ usage:
 tags:
   - flows
   - namespaces
+  - health
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/388/commits/86282d7ef54d0518ef1c50b571fee98208d67585

jotak:
@jpinsonneau I pushed some commits:
- to fix bundle/linter issues
- to reduce metrics cardinality related to the new namespace-based metrics; FTR, initially we didn't have specific namespace-based metrics because they are a subset of the existing workload-based metrics. I think adding them makes sense, so that users can disable the "workload" tag and still be able to get namespace-based metrics; but in the default case I think we should not enable both of them (namespaces+workloads) as it is unoptimal => so, my last commit disables the "namespaces" tag and still allow to show namespace-based metrics in the dashboards via the workload metrics.

If any problem with that, let's discuss when you're back

memodi:
@jpinsonneau @jotak - I am not sure what's the intention of "Application" panel here, based on looking at panels, I find difficult to spot differences between Application and Infrastructure.

image

I was thinking Application would be something that's NOT `netobserv|openshift.*` and to have view something like this:

image

wdyt?

jotak:
Yes, we should exclude netobserv from applications, I agree

jotak:
(FYI I just rebased the branch)

jpinsonneau:
Tested back with @jotak changes and works fine:
image
image

@memodi note that @jotak excluded netobserv from `Applications` panels
image
I only see it appearing for kube api queries (default -> netobserv)

jotak:
@memodi also another change that I did is to run a top10 instead of top5, because I found top5 has a tendency to often show different metrics for different timestamps, which results in "chunking" more the timeseries. With top10 I feel like the visualization is better / more consistent; sounds good?

PR #391, opened 2023-07-06
NETOBSERV-1017 Metrics: pod or service talkers

jpinsonneau:
Requires:
- https://github.com/netobserv/network-observability-operator/pull/388
- https://github.com/netobserv/flowlogs-pipeline/pull/450 

image

### DIFF
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -329,9 +329,9 @@ type FLPMetrics struct {
 	Server MetricsServerConfig `json:"server,omitempty"`
 
 	// `ignoreTags` is a list of tags to specify which metrics to ignore. Each metric is associated with a list of tags. More details in https://github.com/netobserv/network-observability-operator/tree/main/controllers/flowlogspipeline/metrics_definitions .
-	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `pods`, `services`, `nodes-flows`, `namespaces-flows`, `workloads-flows`.
+	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `resources`, `nodes-flows`, `namespaces-flows`, `workloads-flows`, `resources-flows`.
 	// Namespace-based metrics are covered by both `workloads` and `namespaces` tags, hence it is recommended to always ignore one of them (`workloads` offering a finer granularity).
-	//+kubebuilder:default:={"egress","packets","nodes-flows","namespaces-flows","workloads-flows","namespaces", "pods"}
+	//+kubebuilder:default:={"egress","packets","nodes-flows","namespaces-flows","workloads-flows", "resources-flows","namespaces", "resources"}
 	// +optional
 	IgnoreTags []string `json:"ignoreTags"`
 
--- a/api/v1beta1/flowcollector_types.go
+++ b/api/v1beta1/flowcollector_types.go
@@ -329,9 +329,9 @@ type FLPMetrics struct {
 	Server MetricsServerConfig `json:"server,omitempty"`
 
 	// `ignoreTags` is a list of tags to specify which metrics to ignore. Each metric is associated with a list of tags. More details in https://github.com/netobserv/network-observability-operator/tree/main/controllers/flowlogspipeline/metrics_definitions .
-	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `nodes-flows`, `namespaces-flows`, `workloads-flows`.
+	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `pods`, `services`, `nodes-flows`, `namespaces-flows`, `workloads-flows`.
 	// Namespace-based metrics are covered by both `workloads` and `namespaces` tags, hence it is recommended to always ignore one of them (`workloads` offering a finer granularity).
-	//+kubebuilder:default:={"egress","packets","nodes-flows","namespaces-flows","workloads-flows","namespaces"}
+	//+kubebuilder:default:={"egress","packets","nodes-flows","namespaces-flows","workloads-flows","namespaces", "pods"}
 	// +optional
 	IgnoreTags []string `json:"ignoreTags"`
 
--- a/controllers/flowlogspipeline/flp_common_objects.go
+++ b/controllers/flowlogspipeline/flp_common_objects.go
@@ -519,7 +519,7 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 		enrichedStage.EncodePrometheus("prometheus", promEncode)
 	}
 
-	if len(topPromMetrics) > 0 && b.info != nil && b.info.AvailableAPIs != nil && b.info.AvailableAPIs.HasSvcMonitor() {
+	if len(topPromMetrics) > 0 {
 		topBytesStage := enrichedStage.ExtractTimebased("top_bytes", api.ExtractTimebased{
 			Rules: []api.TimebasedFilterRule{
 				{
--- a/controllers/flowlogspipeline/flp_common_objects.go
+++ b/controllers/flowlogspipeline/flp_common_objects.go
@@ -8,7 +8,6 @@ import (
 	"path"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/netobserv/flowlogs-pipeline/pkg/api"
@@ -363,22 +362,17 @@ func (b *builder) podTemplate(hasHostPort, hostNetwork bool, annotations map[str
 var metricsConfigEmbed embed.FS
 
 // obtainMetricsConfiguration returns the configuration info for the prometheus stages
-func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, api.PromMetricsItems, error) {
+func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, error) {
 	entries, err := metricsConfigEmbed.ReadDir(metricsConfigDir)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to access metrics_definitions directory: %w", err)
+		return nil, fmt.Errorf("failed to access metrics_definitions directory: %w", err)
 	}
 
 	// separate metrics and top metrics
-	promConfGen := confgen.NewConfGen(&confgen.Options{
+	cg := confgen.NewConfGen(&confgen.Options{
 		GenerateStages: []string{"encode_prom"},
 		SkipWithTags:   b.desired.Processor.Metrics.IgnoreTags,
 	})
-	topPromConfGen := confgen.NewConfGen(&confgen.Options{
-		GenerateStages: []string{"encode_prom"},
-		SkipWithTags:   b.desired.Processor.Metrics.IgnoreTags,
-	})
-	promConfGens := []*confgen.ConfGen{promConfGen, topPromConfGen}
 
 	for _, entry := range entries {
 		fileName := entry.Name()
@@ -389,32 +383,24 @@ func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, api.PromMe
 
 		input, err := metricsConfigEmbed.ReadFile(srcPath)
 		if err != nil {
-			return nil, nil, fmt.Errorf("error reading metrics file %s; %w", srcPath, err)
+			return nil, fmt.Errorf("error reading metrics file %s; %w", srcPath, err)
 		}
 
-		var confGeng = promConfGen
-		if strings.HasPrefix(fileName, "top_") {
-			confGeng = topPromConfGen
-		}
-		err = confGeng.ParseDefinition(fileName, input)
+		err = cg.ParseDefinition(fileName, input)
 		if err != nil {
-			return nil, nil, fmt.Errorf("error parsing metrics file %s; %w", srcPath, err)
+			return nil, fmt.Errorf("error parsing metrics file %s; %w", srcPath, err)
 		}
 	}
 
-	var metrics = make([]api.PromMetricsItems, len(promConfGens))
-	for i, confGen := range promConfGens {
-		stages := confGen.GenerateTruncatedConfig()
-		if len(stages) != 1 {
-			return nil, nil, fmt.Errorf("error generating truncated config, 1 stage expected in %v", stages)
-		}
-		if stages[0].Encode == nil || stages[0].Encode.Prom == nil {
-			return nil, nil, fmt.Errorf("error generating truncated config, Encode expected in %v", stages)
-		}
-		metrics[i] = stages[0].Encode.Prom.Metrics
+	stages := cg.GenerateTruncatedConfig()
+	if len(stages) != 1 {
+		return nil, fmt.Errorf("error generating truncated config, 1 stage expected in %v", stages)
+	}
+	if stages[0].Encode == nil || stages[0].Encode.Prom == nil {
+		return nil, fmt.Errorf("error generating truncated config, Encode expected in %v", stages)
 	}
 
-	return metrics[0], metrics[1], nil
+	return stages[0].Encode.Prom.Metrics, nil
 }
 
 func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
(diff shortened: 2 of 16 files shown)


controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -499,40 +484,58 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) (*corev
...
-			dashboardCMFile: dashboard,
-		},
+		})
+
+		// prometheus stage (encode) configuration for top metrics
+		topPromEncode := api.PromEncode{
+			Prefix:  "netobserv_",
+			Metrics: topPromMetrics,
+		}
+		topBytesStage.EncodePrometheus("top_prometheus", topPromEncode)
```
jpinsonneau, line 534: The positive side of this approach is that it requires only two stages:
- gather top X src / dst bytes per Name / Namespace / Type / HostName 
- send to prometheus
The resources consumption are quite low on a small cluster.

However, since there is no filter on `Type` prior to this stage, the top X results may contains only pods or services in the dataset and generated graphs may be empty.

image
image
image

Ideally, we could improve the timebased stage to accept grouping. Else we'll need to add 4 parallel filter stages for Src Pod, Src Service, Dst Pod and Dst Service

jpinsonneau:
Rebased on https://github.com/netobserv/network-observability-operator/pull/388

controllers/flowlogspipeline/metrics_definitions/top_pods_egress_bytes_total.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - name: top_pods_egress_bytes_total
+        type: counter
+        valuekey: Bytes
+        filters:
+        - key: FlowDirection
+          value: "1"
+        - key: Duplicate
+          value: "false"
+        labels:
+          - SrcK8S_Name
```
OlivierCazade, line 26: My understanding was that using SrcK8S_Name as label would create too much granularity for prometheus.

May be we should document than enabling this metric can create a lot of pressure on prometheus?

jotak: my take on that is that Src alone at pod granularity is ok (cardinality is N - number of pods), Dst alone is ok (cardinality is still N), but Src + Dst is too much (cardinality is N²)

After all there's already many metrics that are per pod, that would be just another one.

(I haven't looked yet at this `ExtractTimeBased`)

jotak:
@jpinsonneau I asked @memodi , we merged the split infra/apps PR so this one can be rebased and I'll go ahead with the review after that

jpinsonneau:
> @jpinsonneau I asked @memodi , we merged the split infra/apps PR so this one can be rebased and I'll go ahead with the review after that

:+1: thanks ! I cleaned the commit list. It's ready to go

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -527,6 +550,42 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 		enrichedStage.EncodePrometheus("prometheus", promEncode)
 	}
 
+	if len(topPromMetrics) > 0 && b.info != nil && b.info.AvailableAPIs != nil && b.info.AvailableAPIs.HasSvcMonitor() {
```
jotak, line 553: I don't think this should be tied to the ServiceMonitor API: exposing prometheus metrics should be possible even without it. For instance, non-openshift users may have their own prometheus scraping metrics, that doesn't use the prometheus operator hence no ServiceMonitor API

controllers/flowlogspipeline/metrics_definitions/top_pods_egress_bytes_total.yaml
jpinsonneau: I'm testing some traffic generation on services and noticed that I see the generated traffic only using Reporter Node Source + Filter on Destination Resource Service

image
image

So maybe at the end we need only one metric `top_pods_bytes_total` containing only destinations fields and mixing both FlowDirection (egress / ingress) ? WDYT ?

jpinsonneau: You can directly try https://github.com/netobserv/network-observability-operator/pull/392 to get the dashboards :wink:

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -527,6 +550,42 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
...
+		topBytesStage := enrichedStage.ExtractTimebased("top_bytes", api.ExtractTimebased{
+			Rules: []api.TimebasedFilterRule{
+				{
+					Name:          "top_src_bytes",
+					IndexKeys:     []string{"SrcK8S_Name", "SrcK8S_Namespace", "SrcK8S_Type", "SrcK8S_HostName", "FlowDirection", "Duplicate"},
+					OperationType: "avg",
```
jotak, line 559: This is the downside of having aggregation done at ingest time: we're setting in stone the operation type.
I think FLP also supports histograms, which would provide more flexibility on queries, but haven't digged into how to use it in the aggregation stage - cc @ronensc iirc you wrote this, right?

jotak: (but I guess that's not a blocker, we can use avg and see later if there's a need for something else)

jotak:
Reviewed, something I think to change wrt ServiceMonitor API, apart from that lgtm!

controllers/flowlogspipeline/flp_common_objects.go
jpinsonneau: I guess it was part of previous code before rebase.

Done https://github.com/netobserv/network-observability-operator/pull/391/commits/4175cfacf2405ef43039395df1ebe8583968cde1

jotak:
Heads up on the impact, I have done more tests: I'm comparing with an alternative solution where I directly expose the pods metric without the pre-agg topK stage; and also, trying to see the impact on prometheus itself.

Here's FLP CPU / memory:

Capture d’écran du 2023-08-03 08-14-22

=> As seen before, the Aggregation stage has a huge impact on FLP memory. New metrics without pre-aggregation shows almost no difference compared with old metrics.

Prometheus memory:
Capture d’écran du 2023-08-03 08-15-40

=> The difference isn't easily identifiable bc prometheus already takes a lot of memory that always slightly fluctuates. I didn't see a significative difference between pre-aggregated and non-aggregated.

I also looked at other prometheus performance indicators, such as the "Head chunks" increase, there's indeed an increase ~+5K but that's about the same for pre-aggregated and non-aggregated (tbh, here I would have expected an advantage for the pre-aggregated ones)

So I have the feeling it doesn't worth it to pre-aggregate with topK, as it impacts us negatively while it doesn't show clear benefit for prometheus.

jpinsonneau:
> So I have the feeling it doesn't worth it to pre-aggregate with topK, as it impacts us negatively while it doesn't show clear benefit for prometheus.

Interesting, thanks for sharing that @jotak 
Do you feel we should run some perfscale testing here to ensure we get the same results on bigger cluster ?
Else it would be straightforward to remove that stage

jotak:
yeah perf tests would always be nice
btw, my methodology for checking prom's memory is probably not very good, I'll do again a few tests

jpinsonneau:
@jotak following your feedback I removed the `topk` stage and took the opportunity to add `packets` / `flows` metrics disabled by default.
https://github.com/netobserv/network-observability-operator/pull/391/commits/6fe0612fbbc59447c90a72732f0480dc0c7db183

Let's see the impact on perfscale testing to decide if it's the right solution or if we should rollback the last commit

nathan-weinberg:
Note for QE: Do a large-scale pre-merge performance test, perhaps the 120 node scenario

jpinsonneau:
> Note for QE: Do a large-scale pre-merge performance test, perhaps the 120 node scenario

@memodi @nathan-weinberg the code is ready test as is. I'm waiting for the perf test results before merging as we have concerns about the impact on Prometheus performances.
:crossed_fingers: thanks !

jpinsonneau:
> 1. we want cluster with large number of services and pods and measure prometheus before and after we have default ignoreTags and empty ignoreTags?

We can focus on with / without `resources` tag here since it's the purpose of that PR but having all of them is also good to see the extreme config.

> 2. Do you have metrics on prometheus side we should be looking at to measure impact there?

Storage / Memory would be good enough I guess. CPU would be a plus !

> 3. In terms of scale, I am thinking a 25 worker node cluster with ~2500 pods and 100 services, wdyt?

That's perfect :ok_hand: 

> Metrics measured:
> 
> * prometheus_tsdb_head_series
> * prometheus rss
> * prometheus CPU
> 
> Measured in 3 iterations at above scale with catalog image on this PR:
> 
> 1. with default config on ignoreTags - `active TSDB: 2.05 Mil, RSS: ~10.9GB`
> 2. with just egress, packets under ignoreTags - `active TSDB: 2.44 Mil, RSS: ~13.5GB`
> 3. with everything published (empty ignoreTags)- `active TSDB: 3.77 Mil, RSS: ~18GB`  - in this case you'll find it was stable at 3:09 Mil active TSDB, I noticed some FLP pods were crashing due to memory limits, so around 17:13, I increased FLP memory to 2GB and the active TSDB increased further to 3.77 Mil.

So where is the prom CPU ? Do you see an impact on it ?
What happened to FLP & prom after the memory limit increase  ? did it stabilized or just raise infinitely ?

memodi:
@jpinsonneau @jotak - here's re-run of the test in Loki-less mode, with 3 iterations as shown in graph at same scale as previous run, in the end ignoreTags had config:
```
        ignoreTags:
        - egress
        - packets
        - nodes-flows
        - namespaces-flows
        - workloads-flows
 ```
towards the end Prometheus RSS and active TSDB were ~28.5 GB and 2.54 Mil respectively. CPU usage didn't impact as much so leaving it out.

image

IMHO As part of this feature, we restrict users to not letting them take off all tags from above list since it could make cluster prometheus unstable.

**Comment by jpinsonneau** [comment]
> IMHO As part of this feature, we restrict users to not letting them take off all tags from above list since it could make cluster prometheus unstable.

Thanks for your testings @memodi ! The impact is quite huge as expected. I wonder if the `topK` stage would solve this cc @jotak 

I agree we should improve and avoid letting the user select everything.
Here are some suggestions that could help in that way:
- change `ignoreTags` to `tags` so the user actually needs to explicitly specify expected items
- use enums:
   - `resources` / `namespaces` / `workload` to avoid picking these 3 at the same time
   - `ingress` / `egress` 
   - `packets` / `bytes`  / `flows`
 
We could also think about a less flexible approach where we only have preconfigured cases the user can pick in a single enum such as:
- nodes & pods ingress bytes
- nodes & workload ingress bytes
- nodes & namespaces ingress bytes 
...

**Comment by memodi** [comment]
would it make sense to choose one of the tags of `packets` and `bytes` to reduce cardinality? can always add if customer asks for it.

**Comment by jotak** [comment]
@memodi , sorry for chiming in late here, but I'd say what we want to compare here is the impact on prometheus between the previous version that @jpinsonneau had, and the current version; the diff being that in the previous version we had a pre-aggregate stage in FLP that basically puts more load on FLP but that could potentially be better for prometheus.
I rebuilt an image based on Julien's previous commit (c1de0784) : `quay.io/jotak/network-observability-operator:with-aggregate-stage` : so the goal would be to run test perfs with this image and compare with the current PR image.

**Comment by jotak** [comment]
@memodi @jpinsonneau on the other discussion about preventing users to remove all ignored tags: what you suggest would be a breaking change, right? What about simply writing a big WARNING in the crd doc ?
fwiw, this is how the monitoring team does, in several places they write "if you modify this setting, closely monitor the `prometheus-k8s` deployment for excessive memory usage."  e.g: https://github.com/openshift/cluster-monitoring-operator/blob/master/pkg/manifests/types.go#L496

File: controllers/flowlogspipeline/flp_common_objects.go [hunk 5]
```diff
@@ -236,14 +361,14 @@ func (b *builder) podTemplate(hasHostPort, hostNetwork bool, annotations map[str
...
-// obtainMetricsConfiguration returns the configuration info for the prometheus stage needed to
-// supply the metrics for those metrics
+// obtainMetricsConfiguration returns the configuration info for the prometheus stages
 func (b *builder) obtainMetricsConfiguration() (api.PromMetricsItems, error) {
 	entries, err := metricsConfigEmbed.ReadDir(metricsConfigDir)
 	if err != nil {
 		return nil, fmt.Errorf("failed to access metrics_definitions directory: %w", err)
 	}
 
+	// separate metrics and top metrics
```
jotak, line 371: nit: after last change this comment is irrelevant

memodi:
> What about simply writing a big WARNING in the crd doc ?
> fwiw, this is how the monitoring team does, in several places they write "if you modify this setting, closely monitor the `prometheus-k8s` deployment for excessive memory usage."

+1 that seems like best approach, we could also add similar `Warning notes` in documentation.

skrthomas:
> > What about simply writing a big WARNING in the crd doc ?
> > fwiw, this is how the monitoring team does, in several places they write "if you modify this setting, closely monitor the `prometheus-k8s` deployment for excessive memory usage."
> 
> +1 that seems like best approach, we could also add similar `Warning notes` in documentation. cc @skrthomas

@memodi @jotak @jpinsonneau we can add the warning to the documentation as well. We currently don't have a docs tracker for https://issues.redhat.com/browse/NETOBSERV-139, due to no-doc impact on all the associated epics. So if we move forward with this, I'll create a docs epic for this small piece.

memodi:
> due to no-doc impact on all the associated epics. So if we move forward with this, I'll create a docs epic for this small piece

Epic NETOBSERV-139 containing this story would definitely need doc attention, not sure where it's flagged as no-doc?

skrthomas:
> @skrthomas Epic [NETOBSERV-139](https://issues.redhat.com//browse/NETOBSERV-139) containing this story would definitely need doc attention, not sure where it's flagged as no-doc?

I guess it isn't actually labeled as no-doc, sorry for misrepresenting that. What I meant is that in documentation planning, this was discussed as not having any components with doc impact, and Julien's comment to me that the updates don't have any documentation impact. There are updates to the Health dashboard, but we don't go into specifics about the metrics shown there. Are there other stories within 139 that you think have doc impact?

skrthomas:
@jotak would the warning be with the FlowCollector api doc re-generation?

jotak:
> @jotak would the warning be with the FlowCollector api doc re-generation?

yes

api/v1beta1/flowcollector_types.go
```diff
@@ -329,9 +329,9 @@ type FLPMetrics struct {
 	Server MetricsServerConfig `json:"server,omitempty"`
 
 	// `ignoreTags` is a list of tags to specify which metrics to ignore. Each metric is associated with a list of tags. More details in https://github.com/netobserv/network-observability-operator/tree/main/controllers/flowlogspipeline/metrics_definitions .
-	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `nodes-flows`, `namespaces-flows`, `workloads-flows`.
+	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `resources`, `nodes-flows`, `namespaces-flows`, `workloads-flows`, `resources-flows`.
```
jotak, line 332:
```suggestion
	// Warning: removing tags from this list will result in more metrics creation, which can alter cluster performance and stability. It is recommended to monitor the memory and CPU usage of the `prometheus-k8s` deployment when doing so.
	// Available tags are: `egress`, `ingress`, `flows`, `bytes`, `packets`, `namespaces`, `nodes`, `workloads`, `resources`, `nodes-flows`, `namespaces-flows`, `workloads-flows`, `resources-flows`.
```

cc @skrthomas for doc review :-)

skrthomas: Small suggestion to avoid "will" because peer reviewers will nab it: 

Warning: The removal of tags from this list results in increased metrics creation, which can...`prometheus-k8s` deployment when removing tags.

jotak: > Small suggestion to avoid "will" because peer reviewers will nab it

I'll learn it eventually

skrthomas: Its allll good :) I find my own PRs having it all the time. That's what peer reviews are for!

memodi:
> 

is aggregation performed only on `netobserv_top_pods_ingress/egress_bytes_total` ? I have `workloads` enabled but I don't see any aggregation happening there.

jpinsonneau:
> is aggregation performed only on `netobserv_top_pods_ingress/egress_bytes_total` ? I have `workloads` enabled but I don't see any aggregation happening there.

Yes, the `top_bytes` stage only affect new bytes metrics for pods / services. The already existing metrics are not altered by this PR.

jpinsonneau:
Closing this as it will need to be reworked with [NETOBSERV-1284](https://issues.redhat.com/browse/NETOBSERV-1284) approach

PR #393, opened 2023-07-12
Upgrade linter - fix found issues - use go1.20

jotak:


OlivierCazade (commented):
Changes are looking good to me, but I think we should also bump the version in the go.mod file since we are not testing 1.18 anymore.

jotak:
> Changes are looking good to me, but I think we should also bump the version in the go.mod file since we are not testing 1.18 anymore.

Hmm .. i did that, not sure why it's not in the commit, let me push again :-)

OlivierCazade (approved):

PR #398, opened 2023-07-19
NETOBSERV-1201: Make empty CR work out of the box

jotak:
Fixing a couple of default values that weren't working as expected:
- metav1.Duration fields need to be pointers so their default is actually used
- deploymentModel shouldn't be required as it default to DIRECT
- Agent type default wasn't working due to missing omitempty
- Add a test to make sure empty spec works


pkg/helper/helpers.go
```diff
@@ -68,3 +70,10 @@ func MaxLabelLength(in string) string {
 	}
 	return in[:maxLabelLength]
 }
+
+func UnstructuredDuration(in *metav1.Duration) string {
```
msherif1234, line 74: pls add ut coverage for this api, I have one for u :)
```golang
func TestUnstructuredDuration(t *testing.T) {

  t.Run("nil input", func(t *testing.T) {
    var d *metav1.Duration
    got := UnstructuredDuration(d)
    want := ""
    
    if got != want {
      t.Errorf("got %q, want %q", got, want) 
    }
  })

  t.Run("valid input", func(t *testing.T) {
    d := &metav1.Duration{Duration: time.Minute}
    want := "1m0s"
    
    got := UnstructuredDuration(d)
    
    if got != want {
      t.Errorf("got %q, want %q", got, want)
    }
  })

}
```

msherif1234 (commented):
/LGTM
just suggested adding UT coverage and u need to fix rebase issues

pkg/helper/helpers.go
jotak: thanks! done

jotak:
Merging for dequeuing - I don't think specific tests are needed here, apart from usual regression tests
/approve

PR #399, opened 2023-07-25
NETOBSERV-939: Detect which dashboards to show based on ocp version

jotak:
This PR is based on https://github.com/netobserv/network-observability-operator/pull/386 - merge it first
Related console-plugin PR: - to come -

Based on OCP version, the console plugin reconciler will configure a
list of dashboard names that are available, so the the plugin will
create links for them

- Read & compare versions
- Some refactoring:
  - Permission vendor detection and Available API detection were two
    mechanisms very similar: merge Permission vendor detection in
Available APIs
  - Reading cluster info (id, version) can also considered as part of the discovery
    mechanism: merge them all in a new ClusterInfo structure, that
provides the ClusterVersion stuff and the available APIs


pkg/cluster/dashboards_list.go
```diff
@@ -0,0 +1,54 @@
...
+		availableCMNames: map[string]interface{}{
+			constants.KubernetesNetworkDashboard: nil,
+		},
+	}
+}
+
+func (d *Dashboards) CheckClusterDashboards(ctx context.Context, clusterInfo *Info) {
+	if ok, err := clusterInfo.OpenShiftVersionIsAtLeast("4.15.0"); err != nil {
+		// Log error but do not fail: it's likely a bug in code, if the openshift version cannot be found
+		log.FromContext(ctx).Error(err, "Could not get available dashboards for this cluster version. Is it OpenShift?")
```
jpinsonneau, line 34: Nit: log info including `err.Message` instead ?

jpinsonneau (approved):
Code looks good, maybe just change the log level to `info` when not on Openshift ?

pkg/cluster/dashboards_list.go
jotak: if fact this function is only called for openshift (on the caller's side, it's inside a `if c.HasOCPSecurity()` block, which is our way to detect openshift). But I agree it could be made more explicit, let me see that...

jotak:
@jpinsonneau [90498ef](https://github.com/netobserv/network-observability-operator/pull/399/commits/90498ef280011b5e5835fcd07abdb579a52c6fee) => better like that?

jotak:
closing this old PR as it's not clear how we want the dashboard list displayed, and it's quite outdated now .. it's simpler to restart from scratch when needed
I'm extracting the part about discovering ocp version here: https://github.com/netobserv/network-observability-operator/pull/817

PR #402, opened 2023-08-09
NETOBSERV-1109 remove reinterpret direction

jpinsonneau:
Following [NETOBSERV-1099](https://issues.redhat.com/browse/NETOBSERV-1099) EGRESS / INGRESS metrics are not consistent using `reinterpretDirection` function.

I suggest to simply remove it to stick with IPFIX definition. The metrics are consistent after that (see https://github.com/netobserv/network-observability-console-plugin/pull/366#issuecomment-1686139968).
Keeping it double the numbers in the `Overview` section graphs

### DIFF
--- a/controllers/flowlogspipeline/flp_common_objects.go
+++ b/controllers/flowlogspipeline/flp_common_objects.go
@@ -294,16 +294,7 @@ func (b *builder) addTransformStages(stage *config.PipelineBuilderStage) error {
 			Input:  "DstAddr",
 			Output: "DstK8S",
 			Type:   api.AddKubernetesRuleType,
-		}, {
-			Type: api.ReinterpretDirectionRuleType,
 		}},
-		DirectionInfo: api.NetworkTransformDirectionInfo{
-			ReporterIPField:    "AgentIP",
-			SrcHostField:       "SrcK8S_HostIP",
-			DstHostField:       "DstK8S_HostIP",
-			FlowDirectionField: "FlowDirection",
-			IfDirectionField:   "IfDirection",
-		},
 	})
 
 	// loki stage (write) configuration


msherif1234:
I am fine removing reinterpret logic given the issues seen
/LGTM

jpinsonneau:
> I am fine removing reinterpret logic given the issues seen /LGTM

I think I wil wait for @jotak anyways before merging this.
I'm still not convinced by the results I get using concrete download / upload testing (see https://github.com/netobserv/network-observability-console-plugin/pull/366#discussion_r1289970464)

msherif1234:
> > I am fine removing reinterpret logic given the issues seen /LGTM
> 
> I think I will wait for @jotak anyways before merging this. I'm still not convinced by the results I get using concrete download / upload testing (see [netobserv/network-observability-console-plugin#366 (comment)](https://github.com/netobserv/network-observability-console-plugin/pull/366#discussion_r1289970464))
as long as QE not blocked then that can wait

jotak:
Did you double-check that the original issue https://issues.redhat.com/browse/NETOBSERV-696 is not being reintroduced with this?
I am not sure why fixing the FlowDirection (ie use reinterpret-direction) would cause troubles, but, that said, I guess I get why it would work now: since reporters are now merged in queries, we're less affected than before by the "unsoundness" of FlowDirection.

jpinsonneau:
> Did you double-check that the original issue https://issues.redhat.com/browse/NETOBSERV-696 is not being reintroduced with this? 

That brings back the initial issue indeed:
- Without reinterpret:
image
image

- With reinterpret:
image
image

> I am not sure why fixing the FlowDirection (ie use reinterpret-direction) would cause troubles

As written in the description, it messes up the [upload / download scenarios](https://github.com/netobserv/network-observability-console-plugin/pull/366#issuecomment-1686139968). Amonts are incorrect. Here is the example for download:

- Without reinterpret (correct numbers):
image
image

- With reinterpret (wrong numbers):
image
image

jotak:
Trying to reproduce, I do indeed see an issue with the bytes volume reported but I get the same issue with or without this PR, in other words, apparently not related to reinterpret-direction

To try locate the bug, I'm looking at the raw flows stored in loki rather than the computed metrics: the raw flows themselves are wrong, ie. they always show a value smaller to what I'd expect after downloading large volumes.

Example: capture in Loki after downloading a 500MB image (sampling is 1) :
Capture d’écran du 2023-08-25 15-17-51

It shows only ~100MB while I would expect 500MB. I see the same issue keeping or removing reinterpret-direction.
My gut feeling would be something wrong in the ebpf agent aggregation logic

jpinsonneau:
Closing this in favor of "INNER" alternative

PR #409, opened 2023-09-07
NETOBSERV-314: NETOBSERV-1274 index type and duplicate (improve query perfs)

jotak:
## Description

Index duplicate field in Loki to improve query performance.
This is a very low cardinality field (2), and very used in queries, so it is an obvious improvement to do

## Dependencies

- https://github.com/netobserv/network-observability-console-plugin/pull/380

## Checklist

### DIFF
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -34,6 +34,6 @@ const (
 	EndConnectionType = "endConnection"
 )
 
-var LokiIndexFields = []string{"SrcK8S_Namespace", "SrcK8S_OwnerName", "DstK8S_Namespace", "DstK8S_OwnerName", "FlowDirection", "Duplicate"}
+var LokiIndexFields = []string{"SrcK8S_Namespace", "SrcK8S_OwnerName", "SrcK8S_Type", "DstK8S_Namespace", "DstK8S_OwnerName", "DstK8S_Type", "FlowDirection", "Duplicate"}
 var LokiConnectionIndexFields = []string{"_RecordType"}
 var FlowCollectorName = types.NamespacedName{Name: "cluster"}
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -34,6 +34,6 @@ const (
 	EndConnectionType = "endConnection"
 )
 
-var LokiIndexFields = []string{"SrcK8S_Namespace", "SrcK8S_OwnerName", "DstK8S_Namespace", "DstK8S_OwnerName", "FlowDirection"}
+var LokiIndexFields = []string{"SrcK8S_Namespace", "SrcK8S_OwnerName", "DstK8S_Namespace", "DstK8S_OwnerName", "FlowDirection", "Duplicate"}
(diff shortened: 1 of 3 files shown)


jotak:
I'll share here some performance tests ...

jotak:
Some quick perf tests below. tl;dr: it's not like chalk and cheese, but still a slight improvement...

```
Test scenario:
- Delete any FlowCollector, LokiStack, clean up AWS S3, delete any Loki PVC
- Install mesh-arena (once - no need to clean up)
- Install LokiStack
- Install FlowCollector with sampling=1
- Wait 10 minutes (it's important to wait same time between runs)
- Open UI: https://console-openshift-console.apps.jtakvori-sep07-0.devcluster.openshift.com/netflow-traffic?timeRange=900&limit=50&match=all&showDup=false&function=last&type=bytes&packetLoss=all&recordType=flowLog&filters=dst_kind%3DPod%3Bsrc_kind%3DPod%3Bdst_namespace%21%3Dopenshift-%2Cnetobserv%3Bsrc_namespace%21%3Dopenshift-%2Cnetobserv&bnf=true
- Open javascript console / monitor network events: calls to "topology"
- Refresh a couple of times to get a mean value / stats

----
Run with operator's `main`:
~1s each (min=1.08, max=1.21)

Run with operator's `5ff4147`:
~900ms each (min=805, max=954)

```

jotak:
Oh but I get muuuuch better results when I index resource types: 
```
~170ms each (min=150ms, max=192ms)
```
This one seems to be a win :-)

jpinsonneau (approved):

jotak:
> Interesting the type has so much impact. I would have not expect such !

Actually I should mention this was with the default filters, that include "SrcType=Pod && DstType=Pod" ; without those filters we wouldn't get such a difference I guess

> 
> Do you think we should introduce `Proto` ? Maybe `Interface` ? The hard part here is to identify which of these would be mostly used by the users. Maybe exposing a list of additionnal indexes in the CR would allow to cover all these cases 😄

`Proto` could possibly be interesting as you say if people commonly use it as a filter, but I don't have the impression it's a common one? Or maybe it's just me, I hardly never filter on it ....
`Interface` is likely not a good fit because it's high cardinality

> 
> We should also introduce `PktDropLatestState` / `PktDropLatestDropCause` / `DnsLatencyMs` / `DnsFlagsResponseCode` / `TimeFlowRttNs` according to enabled features; but that can be a followup.

Yeah but always keeping in mind the cardinality aspect: latencies can have pretty much any value so that would certainly generate too big indexes. But when values are bounded, like states, causes or codes, yeah that could do it.

> 
> Code looks good to me. Thanks @jotak

jotak:
as discussed in jira, this is dev-only PR
/approve

jpinsonneau:
@Amoghrd @nathan-weinberg marking this PR as breaking change as it requires this PR + https://github.com/netobserv/network-observability-console-plugin/pull/380 to work correctly.

Any pending console PR that doesn't contains these changes will show "No result found" in every tab

PR #440, opened 2023-09-29
NETOBSERV-1340: Add operator to "Observability" category in OperatorHub

stleerh:
The name says it all.  This adds the "Observability" category if it doesn't exist and adds Network Observability to this category.

observability_category

PR #451, opened 2023-10-05
NETOBSERV-1340: Add operator to "Networking" category in OperatorHub

stleerh:
There is a request to also add Network Observability to the "Networking" category.

See https://github.com/netobserv/network-observability-operator/pull/440 for related details.


jotak:
@stleerh : I had to revert your previous PR adding to Observability category, as this isn't a recognized category.
Do you know if there is an ongoing change somewhere (OLM?) to create this category? I guess we need first to have this category created before we can add it here - not doing so generates validation failures (unstable downstream builds).

Adding to "Networking" is fine I guess, since this category already exists; you'll need to rebase the PR to fix conflict though

jotak:
(I resolved the conflict via github)

PR #464, opened 2023-10-17
NETOBSERV-1245: NETOBSERV-1304: to enable DNS we no longer require privileged

msherif1234:
## Description

to enable DNS we no longer have to run as privileged
## Dependencies
https://github.com/netobserv/netobserv-ebpf-agent/pull/206

## Checklist

### DIFF
--- a/controllers/ebpf/agent_controller.go
+++ b/controllers/ebpf/agent_controller.go
@@ -206,9 +206,9 @@ func (c *AgentController) desired(ctx context.Context, coll *flowslatest.FlowCol
 		volumeMounts = append(volumeMounts, volumeMount)
 	}
 
-	if helper.IsFeatureEnabled(&coll.Spec.Agent.EBPF, flowslatest.PacketDrop) || helper.IsFeatureEnabled(&coll.Spec.Agent.EBPF, flowslatest.DNSTracking) {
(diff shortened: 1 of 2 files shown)


jpinsonneau (dismissed):
Awesome ! thanks @msherif1234 

@skrthomas we will need to remove the mention about `privileged: true` [in DNS section](https://docs.openshift.com/container-platform/4.13/networking/network_observability/observing-network-traffic.html#network-observability-dns-tracking_nw-observe-network-traffic) when this PR will be merged

jpinsonneau:
If the performances are similar to what we currently have without DNS tracking, we may think about enabling the feature by default as it's an important one :thinking:

msherif1234:
> If the performances are similar to what we currently have without DNS tracking, we may think about enabling the feature by default as it's an important one 🤔

there is additional CPU and memory cycles used by DNS code and maps but I agree we can take another look at scale numbers with and without DNS and decide

msherif1234:
cc'd @Amoghrd

Amoghrd:
Privileged check has to be removed here [flowcollector.go#L270](https://github.com/netobserv/network-observability-operator/blob/main/pkg/helper/flowcollector.go#L270)

skrthomas:
@jpinsonneau @msherif1234 Ok I can update the `privileged: true` requirement for 1.5 as part of this. Just @ me again if you decide to make DNS default behavior.

PR #467, opened 2023-10-20
NETOBSERV-1092 Move CRD fields to advanced

jpinsonneau:
## Description

This PR move unused API fields to debug sections under each component.
See [NETOBSERV-1092](https://issues.redhat.com/browse/NETOBSERV-1092) for the list of affected fields.

## Dependencies

n/a

## Checklist

  * [X] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.


jpinsonneau:
@msherif1234 Should we consider the eBPF cache fields as debug ?

```
	// `cacheActiveTimeout` is the max period during which the reporter aggregates flows before sending.
	// Increasing `cacheMaxFlows` and `cacheActiveTimeout` can decrease the network traffic overhead and the CPU load,
	// however you can expect higher memory consumption and an increased latency in the flow collection.
	//+kubebuilder:validation:Pattern:=^\d+(ns|ms|s|m)?$
	//+kubebuilder:default:="5s"
	CacheActiveTimeout string `json:"cacheActiveTimeout,omitempty"`

	// `cacheMaxFlows` is the max number of flows in an aggregate; when reached, the reporter sends the flows.
	// Increasing `cacheMaxFlows` and `cacheActiveTimeout` can decrease the network traffic overhead and the CPU load,
	// however you can expect higher memory consumption and an increased latency in the flow collection.
	//+kubebuilder:default:=100000
	CacheMaxFlows int32 `json:"cacheMaxFlows,omitempty"`
```

msherif1234:
> @msherif1234 Should we consider the eBPF cache fields as debug ?
> 
> ```
> 	// `cacheActiveTimeout` is the max period during which the reporter aggregates flows before sending.
> 	// Increasing `cacheMaxFlows` and `cacheActiveTimeout` can decrease the network traffic overhead and the CPU load,
> 	// however you can expect higher memory consumption and an increased latency in the flow collection.
> 	//+kubebuilder:validation:Pattern:=^\d+(ns|ms|s|m)?$
> 	//+kubebuilder:default:="5s"
> 	CacheActiveTimeout string `json:"cacheActiveTimeout,omitempty"`
> 
> 	// `cacheMaxFlows` is the max number of flows in an aggregate; when reached, the reporter sends the flows.
> 	// Increasing `cacheMaxFlows` and `cacheActiveTimeout` can decrease the network traffic overhead and the CPU load,
> 	// however you can expect higher memory consumption and an increased latency in the flow collection.
> 	//+kubebuilder:validation:Minimum=1
> 	//+kubebuilder:default:=100000
> 	CacheMaxFlows int32 `json:"cacheMaxFlows,omitempty"`
> ```

I don't think so those knob if user decided to tune will be because resource limitation or excess and will change and stay so they are more of config operation knobs not debug IMHO

api/v1beta2/flowcollector_types.go
```diff
@@ -635,37 +591,10 @@ type FlowCollectorLoki struct {
 	// Set `enable` to `true` to store flows in Loki. It is required for the OpenShift Console plugin installation.
 	Enable *bool `json:"enable,omitempty"`
 
-	//+kubebuilder:default:="1s"
```
jotak, line 638: Like for the ebpf agent, I think batchWait and batchSize could be good to keep not in Debug, for performance tuning?
Especially BatchSize is something users need to keep consistent with the msg max size defined on loki server-side.

jotak:
Just a remark that we should probably take back loki batch settings , other than that LGTM!

api/v1beta2/flowcollector_types.go
jpinsonneau: Well I based that on the Jira you initially wrote :thinking: 
If you changed your mind I can revert. Just let me know.

jpinsonneau: On my side I would feel better as a user to have that elsewhere but `debug` is a strong word here. 
Maybe we could just concider renaming the `debug` sections to something more generic such as `advanced` ?

jpinsonneau:
> Just a remark that we should probably take back loki batch settings , other than that LGTM!

Sure, I moved loki `batch wait`, `batch size` and `timeout` to FLP since these are only related to `processor`.
Since we now have 3 loki related fields and 4 kafka ones I can created subsections under `processor`. WDYT ?

api/v1beta2/flowcollector_types.go
```diff
@@ -883,15 +817,118 @@ type SASLConfig struct {
...
+
+	//+kubebuilder:validation:Minimum=0
+	//+kubebuilder:default:=2
+	// `maxRetries` is the maximum number of retries for client connections.
+	MaxRetries *int32 `json:"maxRetries,omitempty"`
+
+	//+kubebuilder:default:={"app":"netobserv-flowcollector"}
+	// +optional
+	// `staticLabels` is a map of common labels to set on each flow.
+	StaticLabels map[string]string `json:"staticLabels"`
```
jotak, line 901: I'm fine with your suggestion to move flp-related loki config under processor;
For those 4 fields I guess they should be prefixed with "Loki" like you did for the others

jpinsonneau: Renamed while rebasing https://github.com/netobserv/network-observability-operator/compare/3ae36b13d67536bd3a8ca834c04813f9c72aad31..93e67291a0ab7be328dbe506037dd39c8e5b852a

api/v1beta1/flowcollector_webhook_test.go
```diff
@@ -29,7 +29,6 @@ func TestBeta1ConversionRoundtrip_Loki(t *testing.T) {
 					Enable:             true,
 					InsecureSkipVerify: true,
 				},
-				BatchSize: 1000,
```
jpinsonneau, line 32: For `beta1` -> `beta2` -> `beta1` cycle, we loose `BatchSize` value since this field moved from `Loki` to `Processor` spec.

If we want to manage such, we'll need to manually override `Convert_v1beta1_FlowCollector_To_v1beta2_FlowCollector` and `Convert_v1beta2_FlowCollector_To_v1beta1_FlowCollector` to copy the fields. 

Is it worth the price ? @jotak @msherif1234

jotak: I think it's necessary to do it, else we loose the ability to set this field, since the conversions always happen between the served version (like v1beta2) and the stored version (v1beta1) ?

jotak: (and same for all the other fields that are moved)

jotak: also I tested deploying the sample flowcollector doesn't work, I think it still has all the old fields

jpinsonneau: Hm, that's annoying since it always runs the conversion webhook, you end with `debug` fields in the yaml all the time even if you directly `v1beta2` sample:
```
spec:
  agent:
    ebpf:
      ...
      debug: {}
  consolePlugin:
    ...
    debug:
      port: 9001
      register: true
  processor:
    ...
    debug:
      port: 2055
      lokiMaxRetries: 2
      lokiMaxBackoff: 5s
      conversationTerminatingTimeout: 5s
      conversationEndTimeout: 10s
      lokiStaticLabels:
        app: netobserv-flowcollector
      enableKubeProbes: true
      lokiMinBackoff: 1s
      healthPort: 8080
      dropUnusedFields: true
```

I would prefer to omit these when default.

jpinsonneau:
- Added v1beta1 <> v1beta2 FlowCollector manual conversion to move Loki / Processor fields
- Added webhook tests
- Updated v1beta2 sample

api/v1beta1/flowcollector_webhook_test.go
jpinsonneau: I found a way to load defaults from CRD and show custom values only when set.
That required a bit of refactoring and creating a new [crd helper](https://github.com/netobserv/network-observability-operator/pull/467/files#diff-73ccb26d626eeea081eea7a3e9a8dd2fe3553bf80a746a1762a0c98467eff7d3) but it works very well and the user still have autocompletion in the yaml.

For simplicity, I have created [getters in flowcollector helper](https://github.com/netobserv/network-observability-operator/pull/467/files#diff-05471c458d8215e64e74c8e9123420dc94a81bddf0560d1536365c1ac95d68f9) that sets defaults and load custom values in their related config so we can rely on these everywhere in the code and we don't need to check for nils.

jpinsonneau:
Rebased with new approach:
- loading defaults in code from CRD
- improved helpers to rely on always fully defined stucts
- clearing CR debug fields / sections everytime possible

jpinsonneau:
Rebased without changes

jpinsonneau:
Found a bug on conversion webhook: https://github.com/netobserv/network-observability-operator/pull/467/commits/04269b8e0664df4c6e011658786908e5e5e2d9c9

jotak (approved):

jpinsonneau:
I've rebased again since some changes were introduced in https://github.com/netobserv/network-observability-operator/pull/501 requiring https://github.com/netobserv/network-observability-operator/pull/467/commits/83166a87b297edef70956f8461498ca3b91d30f6 update

jotak (approved):

PR #394, opened 2023-07-12
NETOBSERV-1110: Enable support for Flow RTT

dushyantbehl:
Related console PR: https://github.com/netobserv/network-observability-console-plugin/pull/365

PR #474, opened 2023-10-23
NETOBSERV-764: Loki v1beta2 model follow-up

jotak:
## Description

API changes:
- Improve default values in monolithic and microservices modes to stick with our guides (e.g. http://loki-distributor:3100/ as the ingester URL)
- Rename modes so that they match upstream Loki naming
- In microservices mode, the querier URL isn't optional (?)
- Add ability to configure tenant in these modes
- Add documentation in places where it was missing; fix some typos
- Strong-type LokiMode; set the enum a CamelCase as this is recommended; set it as a unionDiscriminator

Implementation changes:
- Use a "Hub" struct to convert any Loki mode into something alike Manual config; this avoids having to define many getter helpers each doing a switch/case.
- Adapt some tests to better cover nominal use case, e.g. certificates_test now uses LokiStack mode

Also, when a lokistack is referenced, the corresponding role YAML is now automatically be created/updated. It sets the correct FLP service account depending on whether kafka is used. 

## Dependencies

n/a

## Checklist


api/v1beta2/flowcollector_types.go
```diff
@@ -562,74 +562,95 @@ type LokiManualParams struct {
...
+
 const (
-	LokiModeManual      = "MANUAL"
-	LokiModeDistributed = "DISTRIBUTED"
-	LokiModeMonolith    = "MONOLITH"
-	LokiModeLokiStack   = "LOKISTACK"
+	LokiModeManual        LokiMode = "Manual"
+	LokiModeLokiStack     LokiMode = "LokiStack"
+	LokiModeMonolithic    LokiMode = "Monolithic"
+	LokiModeMicroservices LokiMode = "Microservices"
```
jpinsonneau, line 617: Why these should be camel case when we use uppercase for `agent type`, `deployment mode`, `HPA status`, `Loki auth`, `SASL`, `Exporter type` ?

pkg/helper/loki_config.go
```diff
@@ -0,0 +1,102 @@
...
+		MaxRetries:   spec.MaxRetries,
+		StaticLabels: spec.StaticLabels,
+	}
+	switch spec.Mode {
+	case flowslatest.LokiModeLokiStack:
+		dotNamespace := ""
+		if len(spec.LokiStack.Namespace) > 0 {
+			dotNamespace = "." + spec.LokiStack.Namespace
+		}
+		gatewayURL := fmt.Sprintf("https://%s-gateway-http%s.svc:8080/api/logs/v1/network/", spec.LokiStack.Name, dotNamespace)
```
jpinsonneau, line 38: We could fallback on `FlowCollectorSpec.namespace` here to have something easier to read and consistent

jpinsonneau (approved):
Code looks good at first look, some suggestions

api/v1beta2/flowcollector_types.go
jotak: cf https://github.com/netobserv/network-observability-operator/pull/394#discussion_r1304927133
We did it wrong on the others.
Now we have a mix, we should eventually make them all consistent, but from now on, like we did on Features, we should adopt the correct convention

pkg/helper/loki_config.go
jotak: this would add a dependency on `spec.namespace` in the function input, which would break the logic in `Convert_v1beta2_FlowCollectorLoki_To_v1alpha1_FlowCollectorLoki` (https://github.com/netobserv/network-observability-operator/pull/474/files#diff-23e9a27bd609a9414ced41fdaf8063ade03a51b3154d85ad4edfdae228dd5531R139-R150) where we need to run this conversion only with `v1beta2.FlowCollectorLoki` as an input, not the whole spec. So removing the dependency on `spec.namespace` actually helps there.

jotak:
I added the changes to automatically create the roles/binding for reading/writing loki logs (these things: https://github.com/netobserv/documents/blob/main/examples/loki-stack/role.yaml )
FYI @nathan-weinberg @Amoghrd that's surely something you do in automation scripts, that can be removed once this PR is merged

api/v1beta2/flowcollector_types.go
jpinsonneau: so to avoid having a mix I would suggest to fix everything at once for `v1beta2` since it's already an API change. WDYT ?

controllers/flowlogspipeline/flp_common_objects.go
```diff
@@ -773,6 +774,57 @@ func (b *builder) clusterRoleBinding(ck ConfKind, mono bool) *rbacv1.ClusterRole
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: constants.LokiCRReader,
+				},
+				Rules: []rbacv1.PolicyRule{{
+					APIGroups:     []string{"loki.grafana.com"},
+					Resources:     []string{"network"},
+					ResourceNames: []string{"logs"},
+					Verbs:         []string{"get"},
+				}},
+			},
```
jpinsonneau, line 801: Should we put that guy on console plugin side when enabled ?
Also the role binding only mention the writer, not the reader :thinking:

api/v1beta2/flowcollector_types.go
jotak: I agree to do it before releasing v1beta2, but rather on a different PR to not mix it up

jotak: I created a task: https://issues.redhat.com/browse/NETOBSERV-1374
We should add there everything that needs to be cleaned up before we do the first release, ie. any breaking change that we might still find before 1.5

jotak:
@jpinsonneau the update-bundle check [fails](https://github.com/netobserv/network-observability-operator/actions/runs/6625871263/job/17997681952?pr=474) apparently because I'm using a different version of `yq` than what the CI has.
In your previous PR, I didn't see how you installed yq on CI .. or is it already present in the runner image? We should find a way to pin the version...

controllers/flowlogspipeline/flp_common_objects.go
jotak: (done creating a `loki` package via [4d5bc3e](https://github.com/netobserv/network-observability-operator/pull/474/commits/4d5bc3eefb374ea173cc7872dc7057b6b71d578d))

jpinsonneau: thanks !

jpinsonneau (approved):
LGTM, thanks @jotak

jpinsonneau:
Can you please remove `flows_v1beta2_flowcollector_lokistack.yaml` ? It seems not relevent anymore

jotak:
> Can you please remove `flows_v1beta2_flowcollector_lokistack.yaml` ? It seems not relevent anymore

@jpinsonneau done
(and rebased)

jpinsonneau (approved):

PR #476, opened 2023-10-25
NETOBSERV-754: Optimize cache usage

jotak:
## Description

Cut down memory usage; controller-runtime may cache whole cluster configmap / secret for a single `client.Get` call: prevent that.

## Dependencies

n/a

## Checklist


controllers/flowcollector_controller.go
```diff
@@ -231,10 +230,8 @@ func (r *FlowCollectorReconciler) handleNamespaceChanged(
 func (r *FlowCollectorReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&flowslatest.FlowCollector{}).
-		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.DaemonSet{}).
-		Owns(&ascv2.HorizontalPodAutoscaler{}).
```
OlivierCazade, line 237: My understanding of removing this is that we will not get notification for reconciliation if the  HPA is manually modified.
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/builder#Builder.Owns

main.go
```diff
@@ -132,6 +133,11 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "7a7ecdcd.netobserv.io",
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				DisableFor: []client.Object{&corev1.ConfigMap{}, &corev1.Secret{}},
```
OlivierCazade, line 138: This will reduce cache memory but will increase a lot queries to kubeAPI.
Are we sure about this? This make many GET request to kubeAPI each reconciliation loop.

jotak: Yes but it will DRASTICALLY reduce cache memory especially on clusters with many cm/secrets, because this caches the whole cluster configmaps/secrets. I think the additional GET are peanuts compared to this (cf slack thread https://app.slack.com/client/T027F3GAJ/C02939DP5L5/thread/C02939DP5L5-1698237974.684409 )

controllers/flowcollector_controller.go
jotak: Yeah tbh I'm not sure why I removed HPA, my point is really about configmap, I'll put HPA back.
It's a trade-off, might be loosing some reconcile events, but the gain overcomes the disadvantages

main.go
jotak: for info, look what the others are doing: https://github.com/search?q=org%3Aopenshift%20DisableFor%3A&type=code ;-)

jotak:
Here's the test I'm doing:

- Deploy operator + default flowcollector
- Run this script that creates 500 big configmaps:

```bash
oc new-project test
for i in {1..500}; do k create cm test-cm-$i -n test --from-file=bundle/manifests/flows.netobserv.io_flowcollectors.yaml ; done
```

**Results**
On current `main`, I'm seeing operator memory flying from 95 MB to almost 400 MB
Capture d’écran du 2023-10-25 11-39-40

With this PR, memory stays flat at 33 MB
Capture d’écran du 2023-10-25 17-44-59

jotak:
@OlivierCazade I'll investigate creating a custom cache instead of removing it entirely for CM/Secrets

jotak:
So, after investigating doing a custom cache, it's not satisfying, I'm going back to totally disabling cache for CM and secrets: improvements go far beyond losses, and this is even better when considering Kube API traffic. Let me detail:

I'm tracking here the operator <-> kube API traffic with our metrics (dogfooding FTW)

**Scenario using current operator**

Capture d’écran du 2023-10-27 07-27-12

You can see huge traffic when I'm creating 200 CMs (900kBps during 2 minutes), and also noticeable traffic at operator startup and when updating 10 CMs (~150-200 kBps for 30s or so).

**Now with this PR** (notice the scale)
Capture d’écran du 2023-10-27 07-37-55

You see: much less traffic at startup (35kBps - because it's not initializing CM informers cache), and no traffic when CMs are created / updated.

The question that remains is: what happens during reconciles? Well, yes, there is a bit more traffic generated: around 7kBps versus 3kBps with cache. We can expect more when we need to watch loki/kafka certificates. But that's still far below the huge difference made by all the unneeded configmap changes being watched all the time cluster-wide.

So what's the difference with our current `main` if I create a custom cache ? No difference here, we'll still have this huge traffic with Kube API. It only allows to save memory, but not traffic, because the way it's done, it's still going to track all cluster CMs - we can only play on eliminating undesired data to not store them in cache, but we'll still be informed of every CM map change.

jotak:
FTR here's my WIP attempt to create a custom cache, but I prefer to stop here since it's still not as good as disabling cache: https://github.com/jotak/network-observability-operator/commit/26c3fbc4699bb4f7a33b31427d7ac55e6fef6f14

OlivierCazade (approved):
LGTM, thank for the investigations Joel

jotak:
Thanks @OlivierCazade 
@nathan-weinberg I guess it's on you :-)

nathan-weinberg:
I ran a **node-density-heavy** job with this PR image, comparing it with our current baseline tracked by UUID **46dbc801-3b5d-471e-8172-6a71c0eedccb**

In that baseline run[1], we see the following values for the memory of the controller pod:

memoryUsageRSS | netobserv-controller-manager-54bd69cc9f-gwjp6 | avg(value) | 157 MB
-- | -- | -- | --

memoryUsageWorkingSet | netobserv-controller-manager-54bd69cc9f-gwjp6 | avg(value) | 193 MB
-- | -- | -- | --

In the new run[2], we see the following values:

memoryUsageRSS | netobserv-controller-manager-7b4887475d-89444 | avg(value) | 156 MB
-- | -- | -- | --

memoryUsageWorkingSet | netobserv-controller-manager-7b4887475d-89444 | avg(value) | 196 MB
-- | -- | -- | --

While the values do not look that much improved, the second run processed many more flows (almost double) than the original run[3]

nFlowsProcessedPerMinuteTotals | avg(value) | 95.48% | 584710.4091 | 1143012.915
-- | -- | -- | -- | --

@jotak let me know your thoughts around this data - another idea I had was running our [cluster-density test case](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-56617) which typically requires raising the controller memory limit to 800MB and seeing if we can run it using your PR image with the default limit of 400MB

[1]https://docs.google.com/spreadsheets/d/11_ppiK42Y1KIssmmYLt9L678EtZEiQ0Qn3EdIOFD4SA/edit?usp=sharing
[2]https://docs.google.com/spreadsheets/d/1Bqvyyog2H0WKPgJZOZ2SvtKnQ8m773f5tu5Qrorzi9k/edit?usp=sharing
[3]https://docs.google.com/spreadsheets/d/1p72gIC8i4T4Rh-7qu5XDCjuQ2nzRgs6N_78OVZJ7ZJI/edit?usp=sharing

jotak:
So it seems like there are side-effects that I didn't see I first - I did my tests with monolithic Loki, and it seems it's doesn't work correctly when using the loki operator. @nathan-weinberg you might be blocked by that.
I'm working on it ...

jotak:
PS: good idea to test with cluster-density !

jotak:
@OlivierCazade FYI I've changed a lot of things in my last commit. Disabling cache like I did before was breaking our certificate watching. So I had to implement a whole new thing which should be optimal: still using cache, but no more than what is needed.
My concern about it is that's it's a whole new piece that needs to be battle-tested. There must be solid perf/consumption advantage to make it worth it.

nathan-weinberg:
@jotak can you rebase so we can run the regressions?

OlivierCazade (approved):
Nice! Thanks Joel.
LGTM

PR #477, opened 2023-10-25
NETOBSERV-1150 Console plugin columns & filters as config

jpinsonneau:
## Description

- import static config for frontend config containing `columns` and `filters` definitions.
- move console plugin args related to `server` and `loki` to configmap for consistency

It also allow devs/users to remove the `ownerReferences` metadata of any resource created by FlowCollector to skip update on these.

The controller will log the following every reconcile loop:
```
2023-10-25T16:12:22.874Z INFO SKIP *v1.ConfigMap update since not owned {"controller": "flowcollector", "controllerGroup": "flows.netobserv.io", "controllerKind": "FlowCollector", "FlowCollector": {"name":"cluster"}, "namespace": "", "name": "cluster", "reconcileID": "41a05182-d7ba-4aeb-8a84-c03e3cdd80c2", "Namespace": "netobserv", "Name": "console-plugin-config"}
```

To quick reload the configmap, you need to restart the associated pod. 

TODO: check console arguments that can be moved to configmap

## Dependencies

https://github.com/netobserv/network-observability-console-plugin/pull/417

## Checklist

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -1,11 +1,11 @@
 package consoleplugin
 
 import (
+	_ "embed"
 	"fmt"
 	"hash/fnv"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	osv1alpha1 "github.com/openshift/api/console/v1alpha1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	flowslatest "github.com/netobserv/network-observability-operator/api/v1beta2"
+	config "github.com/netobserv/network-observability-operator/controllers/consoleplugin/config"
 	"github.com/netobserv/network-observability-operator/controllers/constants"
 	"github.com/netobserv/network-observability-operator/pkg/helper"
 	"github.com/netobserv/network-observability-operator/pkg/volumes"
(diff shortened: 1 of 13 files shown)


jpinsonneau:
It's a bit hard to maintain this PR and continue working in parallel on columns. I've included the following PRs in the config:
https://github.com/netobserv/network-observability-console-plugin/pull/423
https://github.com/netobserv/network-observability-console-plugin/pull/422

It would be great to avoid keeping this PR open too long

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -162,94 +163,49 @@ func (b *builder) deployment(cmDigest string) *appsv1.Deployment {
...
+	// ensure volumes are up to date
+	if b.loki.TLS.Enable && !b.loki.TLS.InsecureSkipVerify {
+		b.volumes.AddCACertificate(&b.loki.TLS, "loki-certs")
+	}
+	if b.loki.StatusTLS.Enable && !b.loki.StatusTLS.InsecureSkipVerify {
+		b.volumes.AddMutualTLSCertificates(&b.loki.StatusTLS, "loki-status-certs")
+	}
+	if b.loki.UseHostToken() {
+		b.volumes.AddToken(constants.PluginName)
+	}
```
jotak, line 208: it seems like the volumes are added twice: here and from `setLokiConfig`. Is there a reason? Ideally if we could keep them from `setLokiConfig` it would be better, so we only have in a single place code related to checking TLS config

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -340,38 +300,97 @@ func (b *builder) metricsService() *corev1.Service {
...
+	err := yaml.Unmarshal(staticFrontendConfig, &config.Frontend)
+	if err != nil {
+		return nil, "", err
 	}
+	b.setFrontendConfig(&config.Frontend)
 
-	configStr := "{}"
-	if bs, err := yaml.Marshal(config); err == nil {
+	var configStr string
+	bs, err := yaml.Marshal(config)
```
jotak, line 389: good thing to add error handling here ... since you did that you could also do the same line 407 below :-) 
(error currently ignored on `hasher.Write` )

controllers/consoleplugin/consoleplugin_objects.go
jpinsonneau: All of these call the `insertOrReplace` function behind the scene
https://github.com/netobserv/network-observability-operator/blob/5fa07177c2ad1c3c9c39d88fdeb9eee40bfb1f35/pkg/volumes/builder.go#L133-L144

From what I seen, without that piece, the volumes are not up to date

controllers/consoleplugin/consoleplugin_objects.go
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/477/commits/9e37a34b77e30ded8990f5df1cd31a323fa217d0

controllers/consoleplugin/consoleplugin_objects.go
jotak: hmm, this is weird ... I'd prefer to keep the builder process as straightforward as possible, we should need this logic IMO .. I can take a look at that outside of your PR

jotak (approved):
/lgtm
I'll take a closer look at the volume creation in parallel

PR #480, opened 2023-10-31
NETOBSERV-1383: do not turn off conversion-gen

jotak:
conversion-gen was turned off on some fields because it didn't work well with some alpha1 missing fields.

But the problem is that it's also turned off for beta1<->beta2 conversions, which we don't want.

This is turning them back on, however, adding the missing fields in alpha1 (which is deprecated) so that it converts without trouble

## Dependencies

n/a

## Checklist


msherif1234:
/lgtm
its expected the older version won't match new version so at somepoint we probably need to findout what is the best practice to do this instead of adding new fields to older version but not urgent

PR #486, opened 2023-11-08
NETOBSERV-1398: fix loki url when same namespace

jotak:
## Description

Host like "loki-gateway-http.svc" seem invalid

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/helper/loki_config.go
+++ b/pkg/helper/loki_config.go
@@ -19,7 +19,7 @@ type LokiConfig struct {
 	StaticLabels map[string]string
 }
 
-func NewLokiConfig(spec *flowslatest.FlowCollectorLoki) LokiConfig {
+func NewLokiConfig(spec *flowslatest.FlowCollectorLoki, namespace string) LokiConfig {
 	loki := LokiConfig{
 		BatchWait:    spec.BatchWait,
 		BatchSize:    spec.BatchSize,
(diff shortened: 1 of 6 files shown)


pkg/helper/loki_config.go
```diff
@@ -31,15 +31,15 @@ func NewLokiConfig(spec *flowslatest.FlowCollectorLoki) LokiConfig {
...
-			dotNamespace = "." + spec.LokiStack.Namespace
+			ns = spec.LokiStack.Namespace
 		}
-		gatewayURL := fmt.Sprintf("https://%s-gateway-http%s.svc:8080/api/logs/v1/network/", spec.LokiStack.Name, dotNamespace)
+		gatewayURL := fmt.Sprintf("https://%s-gateway-http.%s.svc:8080/api/logs/v1/network/", spec.LokiStack.Name, ns)
 		loki.LokiManualParams = flowslatest.LokiManualParams{
 			QuerierURL:  gatewayURL,
 			IngesterURL: gatewayURL,
-			StatusURL:   fmt.Sprintf("https://%s-query-frontend-http%s.svc:3100/", spec.LokiStack.Name, dotNamespace),
+			StatusURL:   fmt.Sprintf("https://%s-query-frontend-http.%s.svc:3100/", spec.LokiStack.Name, ns),
```
msherif1234, line 42: when ns is `""` there is no need to have `.` in the url name right ?

jotak: When namespace is "", the URL won't be correct, no matter if there's the `.` or not, this is precisely the issue that this PR is fixing. So here, an empty namespace will still result in an invalid URL but it is ultimately replaced with the netobserv's namespace which fixes it all.

PR #489, opened 2023-11-10
NETOBSERV-1334: DNS metrics and dashboards

jotak:
## Description

- Add DNS metrics as predefined metric (per node/namespace/workload)
- namespace_dns_latency_seconds is enabled by default
- update metrics doc
- add dashboards: per-code dns request rate, and dns latencies histogram
- to add histogram feature to the dashboards I had to refactor dashboards builder:
  - define a clear dashboard model, that abstracts Grafana and hopefully should be reusable later for perses
  - separate business logic (dashboard.go) from that model (model.go)
- since histograms is now there I could add the missing RTT dashboard
- fixed a bug where drops dashboards were displayed even though the feature is disabled
- shorten titles
- also fixes [NETOBSERV-1405](https://issues.redhat.com//browse/NETOBSERV-1405)

image

## Dependencies

n/a

## Checklist

* [x] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [x] If so, make sure it is described in the JIRA ticket.


api/v1beta1/flowcollector_types.go
```diff
@@ -234,9 +234,7 @@ type FlowCollectorEBPF struct {
 	// - `PacketDrop`: enable the packets drop flows logging feature. This feature requires mounting
 	// the kernel debug filesystem, so the eBPF pod has to run as privileged.
 	// If the `spec.agent.eBPF.privileged` parameter is not set, an error is reported.<br>
-	// - `DNSTracking`: enable the DNS tracking feature. This feature requires mounting
-	// the kernel debug filesystem hence the eBPF pod has to run as privileged.
-	// If the `spec.agent.eBPF.privileged` parameter is not set, an error is reported.<br>
+	// - `DNSTracking`: enable the DNS tracking feature.<br>
```
msherif1234, line 237: we need to probably say here to get proper DNS tracking records we need to run with sample: 1

config/samples/flows_v1beta2_flowcollector.yaml
```diff
@@ -36,10 +36,12 @@ spec:
...
-      includeList:
-        - node_ingress_bytes_total
-        - workload_ingress_bytes_total
-        - namespace_flows_total
+      # includeList:
+      #   - "node_ingress_bytes_total"
+      #   - "workload_ingress_bytes_total"
+      #   - "namespace_flows_total"
+      #   - "namespace_drop_packets_total"
+      #   - "namespace_rtt_seconds"
```
nathan-weinberg, line 44: @jotak can you explain this change? we've already started updating our CRDs to include those three defaults

jotak: yes there is a subtlety. The default metrics are still these ones, but in CRD, the default is to not have the `includeList` specified. The reason is that, when you set values in includeList, even just keeping this exact same default metrics list, it's considered as a setting explictly set. In a following release, imagine that we add new metrics to the default list: people starting from scratch would get this default, but people upgrading would have the explicit list in settings, hence they would not get them.
And in addition it is also fixing a problem when converting from beta2 to beta1 (NETOBSERV-1405) , where setting default values in this sample was preventing to get updated defaults.

jotak: but you can leave them in your CR, it's still the defaults metrics that we expect, and they are defined in code: https://github.com/netobserv/network-observability-operator/blob/main/pkg/metrics/predefined_metrics.go#L37-L42

pkg/metrics/predefined_metrics.go
```diff
@@ -125,6 +126,21 @@ func init() {
...
+		predefinedMetrics = append(predefinedMetrics, taggedMetricDefinition{
+			PromMetricsItem: flpapi.PromMetricsItem{
+				Name:     fmt.Sprintf("%s_dns_latency_seconds", groupTrimmed),
+				Type:     "histogram",
+				ValueKey: "DnsLatencyMs",
+				Filters: []flpapi.PromMetricsFilter{
+					{Key: "DnsId", Type: flpapi.PromFilterPresence},
+				},
+				Labels:     dnsLabels,
+				ValueScale: 1000, // ms => s
```
jpinsonneau, line 140: This scale makes it really difficult to read to me:
image

Why don't we show `ms` as same as in plugin ?

jpinsonneau (commented):
While testing this PR, I saw that we don't use `enum` / `type` on `IncludeList` field.
That result in not suggesting the dashboards:
image

It would be helpful to add a type and set `+kubebuilder:validation:Enum` on this field

jotak:
It wasn't trivial to just use an enum .. but it's done :-)

image

jpinsonneau:
> It wasn't trivial to just use an enum .. but it's done :-)
> 
> image

Awesome ! Thanks a lot, it will really help customers !

jpinsonneau (approved):

jotak:
thanks @Amoghrd !
/approve

PR #502, opened 2023-11-27
NETOBSERV-627: merge Agent and FLP

jotak:
## Description

Do not deploy FLP: instead, configure the Agent to start FLP in-process. This only works without Kafka.

~~This is a PoC: consider merging only if it improves performances.~~ It does

## Dependencies

n/a

## Checklist


jotak:
@msherif1234 ok I finally have all working with mounted certificates etc.
There's also a new agent image: https://github.com/netobserv/netobserv-ebpf-agent/pull/268

controllers/ebpf/agent_controller.go
```diff
@@ -64,7 +66,7 @@ const (
 
 const (
 	exportKafka        = "kafka"
-	exportGRPC         = "grpc"
+	exportFLP          = "direct-flp"
```
jpinsonneau, line 69: Is the plan to force users not using kafka to have `direct-flp` here ?

I wonder if in some cases we want them separated:
- For security pusposes for example when using ebpf privileged mode
- For resources consumption repartition if the observability stack must have a very low impact on a particular node
- For small clusters where a single FLP could be enough

jotak: I'd say removing the non-encrypted grpc connection will always be more secure than having it?
I'm not sure to understand the two other points, as having merged agent+FLP is more lightweight than having both of them as two different pods (they're both daemonsets / one per node)

PR #503, opened 2023-11-28
NETOBSERV-1358: splitting controllers

jotak:
## Description

- Start with a new "Monitoring" controllers that deals with ServiceMonitors / Dashboards etc
- Create a Status Manager that gathers all statuses from each controller
- Each status becomes eventually converted in a Condition (k8s status API), plus there is a global "Ready" status that is a merge of each component status

_edit:_ a next commit also extracted FLP reconciler as a new controller

### Previously

We had only 1 controller, the FlowCollector controller, which managed all deployed components (FLP, console plugin, agents and a few other resources such as the monitoring dashboards)

The code was structured with each component managed by a "Reconciler". In terms of code structure, "reconcilers" are similar to "controllers", but at runtime it differs a lots because all "reconcilers" are managed synchronously, called from the same Reconcile loop, sharing reconcile events / cache & watches / etc.

Each reconciler were amending a global Status via hooks called `SetChanged` / `SetInProgress`. On any error, the single whole reconcile loop would return, setting an error in status conditions.

The FlowCollector status was amended at the end of the single reconcile loop to reflect any new status.

### Now, with this PR

We have 2 controllers, and we plan to create more in follow-up work. There's still the main / legacy FlowCollector controller, from which I extracted functions related to the monitoring stack (dashboards, roles, annotating namespace, etc.). These functions are now managed by the new controller called Monitoring.

So each controller have their own configuration related to cache / watches, their own reconcile loop and requests queue. Their reconciliation code is run asynchronously from each other.

Status management still needs to be coordinated, since each controller may need to change the global status. This is achieved with a new Status Manager that holds a map keeping track of each controller's status, in a thread-safe way.

Each controller can call status manager's functions to update their component's status, such as with `r.status.SetFailure("MonitoringError", err.Error())` or `r.status.SetReady()`.

Typically, in a reconcile loop, a controller would:
- start setting status as Unknown, or Unset if this controller isn't used (e.g. a future IPFIX controller when agent.type is EBPF)
- call `defer r.status.Commit(ctx, r.Client)` to commit (synchronize) status at the end of Reconcile
- wire any deployment/daemonset it installs with status hooks `CheckDaemonSetProgress` and `CheckDeploymentProgress` (they are called in shared functions `ReconcileDeployment` / `ReconcileDaemonSet`)
- call `r.status.SetFailure` or `r.status.Error` (<= syntactic sugar) on error

When the `Commit` function is called, the Status Manager will merge all statuses into a global one, and write it down to the FlowCollector CR

## Dependencies

n/a

## Checklist


pkg/manager/status/status_manager.go
```diff
@@ -0,0 +1,231 @@
...
+	FlowCollectorLegacy ComponentName = "FlowCollectorLegacy"
+	Monitoring          ComponentName = "Monitoring"
+)
+
+var allNames = []ComponentName{FlowCollectorLegacy, Monitoring}
+
+type Manager struct {
+	sync.Mutex
+
+	statuses map[ComponentName]ComponentStatus
```
msherif1234, line 31: can u use syncMap here ?

jotak: sounds good ... sometimes I don't like using sync.Map as the API is less user-friendly, and they don't always offer better performances ; but I just tried here and it plays nicely

controllers/flowcollector_controller.go
```diff
@@ -90,271 +89,105 @@ func NewFlowCollectorReconciler(client client.Client, scheme *runtime.Scheme, co
...
-		cpReconciler = consoleplugin.NewReconciler(&reconcilersInfo, r.config.ConsolePluginImage)
+	if r.mgr.HasConsolePlugin() {
+		cpReconciler = consoleplugin.NewReconciler(reconcilersInfo.NewInstance(r.mgr.Config.ConsolePluginImage, r.status))
 	}
 
 	// Check namespace changed
 	if ns != previousNamespace {
-		if err := r.handleNamespaceChanged(ctx, previousNamespace, ns, desired, &flpReconciler, &cpReconciler); err != nil {
-			return ctrl.Result{}, r.failure(ctx, conditions.CannotCreateNamespace(err), desired)
+		if previousNamespace != "" && r.mgr.HasConsolePlugin() {
```
OlivierCazade, line 142: Not related to this PR.

Handling namespace change create some complexity that is not necessary IMO.

Apparently it is possible to make a field immutable with kubebuilder pattern. I think it could be a nice compromise.
What do you think?

pkg/manager/manager.go
```diff
@@ -0,0 +1,106 @@
...
+//+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors/finalizers,verbs=update
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,resourceNames=hostnetwork,verbs=use
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;create;update;watch
+//+kubebuilder:rbac:groups=apiregistration.k8s.io,resources=apiservices,verbs=list;get;watch
+//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;prometheusrules,verbs=get;create;delete;update;patch;list;watch
+//+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=loki.grafana.com,resources=network,resourceNames=logs,verbs=get;create
+//+kubebuilder:rbac:urls="/metrics",verbs=get
```
OlivierCazade, line 33: How about declaring rbac rule in each controller?

So we can track permissions needed by each controller. This will need a bit more of maintenance so if you prefer to keep all permissions here, I am fine with it.

controllers/flp/flp_ingest_reconciler.go
```diff
@@ -32,58 +28,54 @@ type ingestOwnedObjects struct {
...
-	cmn.Managed.AddManagedObject(RoleBindingName(ConfKafkaIngester), owned.roleBinding)
-	cmn.Managed.AddManagedObject(configMapName(ConfKafkaIngester), owned.configMap)
+	rec := ingesterReconciler{
+		Instance:       cmn,
+		daemonSet:      cmn.Managed.NewDaemonSet(name),
+		promService:    cmn.Managed.NewService(promServiceName(ConfKafkaIngester)),
+		serviceAccount: cmn.Managed.NewServiceAccount(name),
+		configMap:      cmn.Managed.NewConfigMap(configMapName(ConfKafkaIngester)),
+		roleBinding:    cmn.Managed.NewCRB(RoleBindingName(ConfKafkaIngester)),
+	}
```
OlivierCazade, line 40: The new Managed API make this part more readable, this is nice.

controllers/flp/flp_controller.go
```diff
@@ -0,0 +1,228 @@
...
+
+	ns := helper.GetNamespace(&fc.Spec)
+	previousNamespace := r.status.GetDeployedNamespace(fc)
+	loki := helper.NewLokiConfig(&fc.Spec.Loki)
+	cmn := r.newCommonInfo(clh, ns, previousNamespace, &loki)
+
+	r.watcher.Reset(ns)
+
+	// obtain default cluster ID - api is specific to openshift
+	if r.mgr.IsOpenShift() && r.clusterID == "" {
```
OlivierCazade, line 106: May be we can move it to the start function ?

OlivierCazade (commented):
Just a few comments.

If some comments are related to code you moved but did not change, please ignore them.

controllers/flowcollector_controller.go
jotak: I agree that the namespace thing is a pain.. We can make it immutable yes - we then loose the ability to reconfigure it on the fly, but I guess it's quite a rare operation anyway, and the workaround would just be to manually delete then re-install the flowcollector at the desired namespace. That would interrupt flow collection, but it's also interrupted when you modify namespace, so.. yeah I think I'm good with that proposal :-)

pkg/manager/manager.go
jotak: That was what I did at first, but the rules are then duplicated in the resource yamls :-( 
I mean, it still works, but it makes the yamls quite harder to read

controllers/flp/flp_controller.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type subReconciler interface {
+	context(ctx context.Context) context.Context
+	cleanupNamespace(ctx context.Context)
+	reconcile(ctx context.Context, desired *flowslatest.FlowCollector) error
+	getStatus() *status.Instance
+}
+
+func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
jotak: I can also rename `reconcile` but I fell short of inspiration

controllers/flp/flp_ingest_reconciler.go
jotak: yeah I don't know why I didn't do that earlier :)

controllers/flp/flp_controller.go
jotak: I also tried that :) and it doesn't work because  in `Start` the cache hasn't started yet, so all client calls fail

controllers/flp/flp_controller.go
```diff
@@ -0,0 +1,228 @@
...
+		if err := r.Client.Get(ctx, key, cversion); err != nil {
+			log.Error(err, "unable to obtain cluster ID")
+		} else {
+			r.clusterID = string(cversion.Spec.ClusterID)
+		}
+	}
+
+	// Create sub-reconcilers
+	reconcilers := []subReconciler{
+		newMonolithReconciler(cmn.NewInstance(r.mgr.Config.FlowlogsPipelineImage, r.mgr.Status.ForComponent(status.FLPMonolith))),
```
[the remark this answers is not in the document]
jotak: lol again something I tried :-)
The thing is `cmn` is bound to the FlowCollector object, for setting the SetControllerReference hook.
I started to refactor /decouple that but a I was going down the rabbit hole, the refactoring grew bigger and bigger, I preferred to leave that for perhaps later and not touch the `reconcilers.Common` struct now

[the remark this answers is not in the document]
jotak: adding a TODO comment

controllers/flp/flp_controller.go
jotak: (Given it's cached, it's a negligible performance penalty)

pkg/helper/flowcollector.go
```diff
@@ -38,11 +38,11 @@ func HasKafkaExporter(spec *flowslatest.FlowCollectorSpec) bool {
 }
 
 func HPADisabled(spec *flowslatest.FlowCollectorHPA) bool {
-	return spec.Status == flowslatest.HPAStatusDisabled
+	return spec == nil || spec.Status == flowslatest.HPAStatusDisabled
```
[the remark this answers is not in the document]
jotak: I'll actually keep only `HPAEnabled`, to make it more clear

OlivierCazade (dismissed):
LGTM!

jotak:
/approve
thanks @OlivierCazade @nathan-weinberg !

OlivierCazade (approved):

PR #517, opened 2023-12-12
NETOBSERV-1244: Added flow type transformation rule in FLP configuration

OlivierCazade:
## Description

Added flow type transformation rule in FLP configuration

## Dependencies

Related FLP PR:
https://github.com/netobserv/flowlogs-pipeline/pull/554

## Checklist


controllers/flp/flp_pipeline_builder.go
```diff
@@ -72,6 +72,9 @@ func (b *PipelineBuilder) AddProcessorStages() error {
 			Type:   api.AddKubernetesRuleType,
 		}, {
 			Type: api.ReinterpretDirectionRuleType,
+		}, {
+			Type:       "add_kubernetes_infra",
```
OlivierCazade, line 76: TODO once FLP PR is merged: update FLP dependency and change this string using FLP API

OlivierCazade:
/hold
We need to merge FLP PR first and then update this one

api/v1beta1/flowcollector_types.go
```diff
@@ -664,7 +664,7 @@ type FlowCollectorConsolePlugin struct {
 	// `portNaming` defines the configuration of the port-to-service name translation
 	PortNaming ConsolePluginPortConfig `json:"portNaming,omitempty"`
 
-	//+kubebuilder:default:={{name:"Applications",filter:{"src_namespace!":"openshift-,netobserv","dst_namespace!":"openshift-,netobserv"},default:true},{name:"Infrastructure",filter:{"src_namespace":"openshift-,netobserv","dst_namespace":"openshift-,netobserv"}},{name:"Pods network",filter:{"src_kind":"Pod","dst_kind":"Pod"},default:true},{name:"Services network",filter:{"dst_kind":"Service"}}}
+	//+kubebuilder:default:={{name:"Applications",filter:{"flow_type":"app"},default:true},{name:"Infrastructure",filter:{"flow_type":"infra"}},{name:"Pods network",filter:{"src_kind":"Pod","dst_kind":"Pod"},default:true},{name:"Services network",filter:{"dst_kind":"Service"}}}
```
jpinsonneau, line 667: :+1:  love it ! It will simplify the display a lot !

controllers/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -703,6 +709,11 @@ filters:
     component: autocomplete
     placeholder: 'E.g: Ingress, Egress, Inner'
     hint: Specify the direction of the Flow observed at the Node observation point.
+  - id: flow_type
+    name: Flow type
+    component: autocomplete
```
jpinsonneau, line 714: To use `autocomplete` component you will need to implement `getOptions` on plugin side:
https://github.com/netobserv/network-observability-console-plugin/blob/be4ea4532962056f32080cd29bc99ce1fe77ca89/web/src/utils/filter-definitions.ts#L253-L297

We could even improve that to something more generic for `Namespaces`, `Names`, `Kinds` and this new `FlowType` since all of these rely on [autoCompleteCache](https://github.com/netobserv/network-observability-console-plugin/blob/eadc9beff97a2deef33d0494fc080ed3e31e7612/web/src/utils/autocomplete-cache.ts).

That would require a new route, sending field name as argument and refactor [resources.go](https://github.com/netobserv/network-observability-console-plugin/blob/0c2cebc3a144b9e3c24ad266f9af826e5b5aedad/pkg/handler/resources.go) accordingly. WDYT ?

OlivierCazade: I changed to text. I think this is enough since the placeholder already hint the user that this is only infra or app.

And I agree that as a follow up to the generic column, we could have a follow up task to have generic auto complete for generic column.

OlivierCazade:
Updated operator to use this PR:
https://github.com/netobserv/flowlogs-pipeline/pull/554

Until it is merged, to build this branch it is required to use go replace to point at the FLP branch.

jpinsonneau (approved):
Code looks good to me, thanks @OlivierCazade !

jotak:
I'm also glad to see these awful infra/app filters looking so much better now :-)

One thing that is not a blocker but I think I can do that quickly as a follow-up, if you all agree, is to find another name than "flow type" ... there's so many ways to interpret "type", it's too overloaded and could mean totally different things in different context. I think "layer" is quite common to refer to things like infrastructure / applications, wdyt?

jotak (approved):

jpinsonneau:
Changes looks good, flow layer will be better indeed :smiley_cat:

jotak (approved):

jotak:
Thinking about the upgrade process, users will still have their old quick filters. We might have to add in the release note that they could do the change manually in their config (although the old filters will continue to work)

OlivierCazade:
As discussed in scrum today, this will be tested post merge.

PR #529, opened 2024-01-19
NETOBSERV-1407 - Loki labels for deduper merge

jpinsonneau:
## Description

This PR removes `FlowDirection` and `Duplicate` labels when deduper merge mode is enabled.

It also centralize labels array in a single function (`K8S_ClusterName` was not sent to console plugin).

## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -22,6 +22,7 @@ import (
 	"github.com/netobserv/network-observability-operator/controllers/constants"
 	"github.com/netobserv/network-observability-operator/controllers/ebpf"
 	"github.com/netobserv/network-observability-operator/pkg/helper"
+	"github.com/netobserv/network-observability-operator/pkg/loki"
 	"github.com/netobserv/network-observability-operator/pkg/volumes"
 )
 
@@ -310,12 +311,7 @@ func (b *builder) setLokiConfig(lconf *config.LokiConfig) {
 	if lconf.URL != statusURL {
 		lconf.StatusURL = statusURL
 	}
-	// check for connection traking to list indexes
-	indexFields := constants.LokiIndexFields
-	if b.desired.Processor.LogTypes != nil && *b.desired.Processor.LogTypes != flowslatest.LogTypeFlows {
-		indexFields = append(indexFields, constants.LokiConnectionIndexFields...)
-	}
-	lconf.Labels = indexFields
+	lconf.Labels = loki.GetLokiLabels(b.desired)
 	lconf.TenantID = b.loki.TenantID
 	lconf.ForwardUserToken = b.loki.UseForwardToken()
 	if b.loki.TLS.Enable {
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -344,7 +344,8 @@ func (b *builder) setLokiConfig(lconf *config.LokiConfig) {
 }
 
 func (b *builder) setFrontendConfig(fconf *config.FrontendConfig) {
-	var dedupJustMark, dedupMerge bool
+	dedupJustMark, _ := strconv.ParseBool(ebpf.DedupeJustMarkDefault)
+	dedupMerge, _ := strconv.ParseBool(ebpf.DedupeMergeDefault)
 	if helper.UseEBPF(b.desired) {
 		if helper.IsPktDropEnabled(&b.desired.Agent.EBPF) {
 			fconf.Features = append(fconf.Features, "pktDrop")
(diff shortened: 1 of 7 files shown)


jpinsonneau:
Rebased without changes

jpinsonneau:
As a followup, we should try to entirely get rid of `FlowDirection` and `Interface` fields:
- need to rework the way flp `NetworkTransformDirectionInfo` works to manage FlowDirections array
- decide what to do with prometheus metrics (keep current filtering or switch to arrays)
- remove old columns from loki storage / console display

controllers/flowcollector_controller_certificates_test.go
```diff
@@ -146,6 +146,13 @@ func flowCollectorCertificatesSpecs() {
 			DeploymentModel: flowslatest.DeploymentModelKafka,
 			Agent: flowslatest.FlowCollectorAgent{
 				Type: "eBPF",
+				EBPF: flowslatest.FlowCollectorEBPF{
+					Advanced: &flowslatest.AdvancedAgentConfig{
+						Env: map[string]string{
+							"DEDUPER_JUST_MARK": "true",
```
msherif1234, line 152: isn't that the default ? why need to set it ?

pkg/loki/labels.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if desired.Processor.LogTypes != nil && *desired.Processor.LogTypes != flowslatest.LogTypeFlows {
+		indexFields = append(indexFields, constants.LokiConnectionIndexFields...)
+	}
+
+	if desired.Processor.MultiClusterDeployment != nil && *desired.Processor.MultiClusterDeployment {
+		indexFields = append(indexFields, constants.ClusterNameLabelName)
+	}
+
+	if helper.UseEBPF(desired) {
```
msherif1234, line 23: so you are using justMark bool as indication if dupMerge or not ? why not check dupMerge directly ?

controllers/flowcollector_controller_certificates_test.go
jpinsonneau: That's the default indeed !

However removing it seems to affect the reconcile loop and makes the test fail. As far as I remember it says the CR failed to update because it's not up to date (probably from status update code). It would require a dummy update in the tests to make it work.

Everything seems to work fine on a real cluster :smile_cat:

pkg/loki/labels.go
jpinsonneau: /!\ This file only manages loki labels ! 

For now, I only add / remove `FlowDirection`  and `Duplicate` labels according to `justMark` flag: 
```
var LokiDeduperMarkIndexFields = []string{"FlowDirection", "Duplicate"}
```

For `merge` flag, we don't need label on related arrays since there is no deduplication on plugin side. That means queries are running straight forward without making any fancy filtering on these :sunglasses:

jpinsonneau:
While double checking console flags, I figured out that the defaults values were not correctly set in current code when `Advanced` is `nil` (I guess it was a rebase issue).

I fixed that in https://github.com/netobserv/network-observability-operator/pull/529/commits/894cf303cb8a5c80fca9f553243549f95008dd39 but that doesn't solve the `certificates_test` issue.

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -365,14 +362,10 @@ func (b *builder) setFrontendConfig(fconf *config.FrontendConfig) {
 		if b.desired.Agent.EBPF.Advanced != nil {
 			if v, ok := b.desired.Agent.EBPF.Advanced.Env[ebpf.EnvDedupeJustMark]; ok {
 				dedupJustMark, _ = strconv.ParseBool(v)
-			} else {
-				dedupJustMark, _ = strconv.ParseBool(ebpf.DedupeJustMarkDefault)
 			}
 
 			if v, ok := b.desired.Agent.EBPF.Advanced.Env[ebpf.EnvDedupeMerge]; ok {
 				dedupMerge, _ = strconv.ParseBool(v)
```
msherif1234, line 368: not related to recent changes but why we ignore errors from `ParseBool` ?

controllers/flowcollector_controller_certificates_test.go
msherif1234: yeah this is just test code

controllers/consoleplugin/consoleplugin_objects.go
jpinsonneau: These booleans are constants from `agent_controller.go`. We are not supposed to have errors here right ?
If you prefer we can make reconcile loop crash :thinking:

msherif1234: what do u mean ParseBool is what we will get from FC object for the constants u are referring to someone could set the env var to something bad which will trigger
https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/strconv/atob.go;l=17
specially we have no API validation for this envvar try setting to "foo" and see what will happen :)

jpinsonneau: oh I thought we were talking about defaults values here; that's right, I should manage parse error from `Env` input !

jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/529/commits/c891c2f47b81fff1abb4613fda0bd3ef57e60512

PR #535, opened 2024-01-24
swap-image-csv.sh updates and minor fixes to set-plugin-image

memodi:
## Description

Updating `swap-image-csv.sh` script such that DOWNSTREAM_DEPLOYMENT env variable could be patched in CSV and minor fixes to set-plugin-image.

## Dependencies

n/a

## Checklist

### DIFF
--- a/.mk/development.mk
+++ b/.mk/development.mk
@@ -178,7 +178,8 @@ else
 endif
 	@echo -e "\n==> Redeploying..."
 	kubectl rollout status -n $(OPERATOR_NS) --timeout=60s deployment netobserv-controller-manager
-	kubectl wait -n $(releaseRATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
+	kubectl wait -n $(OPERATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
+	@echo -e "\n==> Wait a moment before plugin pod is fully redeployed"
 
 .PHONY: set-release-kind-downstream
 set-release-kind-downstream:
--- a/.mk/development.mk
+++ b/.mk/development.mk
@@ -178,13 +178,15 @@ else
 endif
 	@echo -e "\n==> Redeploying..."
 	kubectl rollout status -n $(OPERATOR_NS) --timeout=60s deployment netobserv-controller-manager
-	kubectl wait -n $(OPERATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
-	kubectl rollout status -n $(NAMESPACE) --timeout=60s deployment netobserv-plugin
-	kubectl wait -n $(NAMESPACE) --timeout=60s --for condition=Available=True deployment netobserv-plugin
+	kubectl wait -n $(releaseRATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
 
 .PHONY: set-release-kind-downstream
 set-release-kind-downstream:
+ifeq ("", "$(CSV)")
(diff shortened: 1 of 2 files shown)


.mk/development.mk
```diff
@@ -179,12 +179,15 @@ endif
 	@echo -e "\n==> Redeploying..."
 	kubectl rollout status -n $(OPERATOR_NS) --timeout=60s deployment netobserv-controller-manager
 	kubectl wait -n $(OPERATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
-	kubectl rollout status -n $(NAMESPACE) --timeout=60s deployment netobserv-plugin
-	kubectl wait -n $(NAMESPACE) --timeout=60s --for condition=Available=True deployment netobserv-plugin
+	@echo -e "\n==> Wait a moment before plugin pod is fully redeployed"
 
 .PHONY: set-release-kind-downstream
 set-release-kind-downstream:
+ifeq ("", "$(CSV)")
```
msherif1234, line 186: This is for kind cluster deploy which doesn't use CSV

hack/swap-image-csv.sh
```diff
@@ -6,13 +6,19 @@ rel_name=$3
...
+fi
 
 idx_env=$(oc get csv $csv_name -n $namespace -o json | jq --arg NAME "$env_name" '.spec.install.spec.deployments[0].spec.template.spec.containers[0].env | map(.name == $NAME) | index(true)')
 patch2="{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/template/spec/containers/0/env/$idx_env', 'value': {'name': '$env_name', 'value': '$new_image'}}"
 
-oc patch csv $csv_name -n $namespace --type='json' -p "[$patch1, $patch2]"
+if [[ ! -z ${patch1+x} ]] ; then
+  oc patch csv $csv_name -n $namespace --type='json' -p "[$patch1, $patch2]"
+else
+  oc patch csv $csv_name -n $namespace --type='json' -p "[$patch2]"
```
msherif1234, line 20: is this suppose to be `patch1` ?

.mk/development.mk
```diff
@@ -179,12 +179,15 @@ endif
 	@echo -e "\n==> Redeploying..."
 	kubectl rollout status -n $(OPERATOR_NS) --timeout=60s deployment netobserv-controller-manager
 	kubectl wait -n $(OPERATOR_NS) --timeout=60s --for condition=Available=True deployment netobserv-controller-manager
-	kubectl rollout status -n $(NAMESPACE) --timeout=60s deployment netobserv-plugin
```
msherif1234, line 182: did u intentionally remove those 2 lines ?

memodi: yes, since it assumes netobserv-plugin is already deployed, while csv could be patched before any plugin is deployed and there would no netobserv-plugin deployment to check rollout status on. This is in line with what's currently in there for ebpf/flp make targets

hack/swap-image-csv.sh
memodi: so patch2 is what we want in all cases, patch1 is conditional depending upon whether rel_name is set patch1 var will be set.

.mk/development.mk
jotak: I think there's a confusion :-) It's not for KIND clusters, it's "release kind" like in "downstream kind of release"

.mk/development.mk
jotak: I'm good with that, this wasn't really useful actually

PR #539, opened 2024-01-25
bug fix to use OPERATOR_NS instead

memodi:
fixes following #535


jpinsonneau (approved):
:+1: good catch @memodi ! Thanks

PR #550, opened 2024-01-30
Doc: regenerate API refs (FlowCollector + Flow Format)

jotak:
Flow format is now generated from a fields.yaml file stored here, instead of the console plugin's type definition previously used, as it became more likely to be outdated

A check is added to make sure any field referenced in the frontend config is also defined in fields.yaml


jotak:
FYI @skrthomas 
(let's wait a bit that the dev team reviews it, but that should set the ground for 1.5 API doc)
Also, you'll see that I completely changed the layout of the Flow Format spec: this is because I don't use anymore the previous tooling and now we're more free for designing the way we want. So I've moved away from the previous flat text format, to use a table instead. Pls let me know how it plays with the doc integration...

jotak:
to see the new Flow Format spec in github's renderer: https://github.com/netobserv/network-observability-operator/blob/a5606eef057a8e40c798d1cc2609cc8998468ef0/docs/flows-format.adoc

docs/fields.yaml
```diff
@@ -0,0 +1,129 @@
+# Fields definition, used to generate documentation
```
jpinsonneau, line 1: Can we add this to the static-frontend-config ?

I would like to get rid of `ipfix.ts` file from console plugin repo in the end. 
We can also rely on these definitions to check if field is a label and add types since the query engine check for numbers.

jotak: I can surely move that to the config, ok, but it will stay unused in code for the time being, right? at least for this PR

jotak: well then I can also change my `flow-format-validation` script into a go test

docs/flows-format.adoc
```diff
@@ -3,489 +3,277 @@
...
-Egress::
-
-• *Egress* = `"1"`
-
-Outgoing traffic, from the network interface observation point
+This is the specification of the network flows format. That format is used when a Kafka exporter is configured, for Prometheus metrics labels as well as internally for the Loki store.
+
+The "Filter ID" column shows which related name to use when defining Quick Filters (see `spec.consolePlugin.quickFilters` in the `FlowCollector` specification).
+
+The "Loki label" column is useful when querying Loki directly: label fields need to be selected using link:https://grafana.com/docs/loki/latest/logql/log_queries/#log-stream-selector[stream selectors].
```
skrthomas, line 10: Do we still need to write this too:
> If you are reading this specification as a reference for the Kafka export feature, you must treat all _Labels_ and _Fields_ as regular fields and ignore any distinctions between them that are specific to Loki.

jotak: No I removed that because now it's not organized by Fields and Labels anymore - it's directly written for Kafka exporter users

skrthomas: Thanks Joel. I re-copied your latest IBMSG tweaks to my PR.

controllers/flowcollector_controller_certificates_test.go
```diff
@@ -486,15 +484,17 @@ func flowCollectorCertificatesSpecs() {
...
-			Eventually(func() error {
-				err := k8sClient.Get(ctx, crKey, &flowCR)
-				if err == nil {
-					err = fmt.Errorf("CR is still present. Status: %s", status.ConditionsToString(flowCR.Status.Conditions))
-				}
-				return err
-			}, timeout, interval).Should(MatchError(`flowcollectors.flows.netobserv.io "cluster" not found`))
-		})
+		// Flaky assertion here, quite annoying, some weirdness with `envtest` maybe?
+		// Anyway it doesn't bring much value. Disabling it for the time being
```
jpinsonneau, line 488: I wonder if this is not tied to the parallelization of the tests. Currently, we are not setting any parameter wich means tests from different packages can run concurently.
```
	-parallel n
	    Allow parallel execution of test functions that call t.Parallel, and
	    fuzz targets that call t.Parallel when running the seed corpus.
	    The value of this flag is the maximum number of tests to run
	    simultaneously.
	    While fuzzing, the value of this flag is the maximum number of
	    subprocesses that may call the fuzz function simultaneously, regardless of
	    whether T.Parallel is called.
	    By default, -parallel is set to the value of GOMAXPROCS.
	    Setting -parallel to values higher than GOMAXPROCS may cause degraded
	    performance due to CPU contention, especially when fuzzing.
	    Note that -parallel only applies within a single test binary.
	    The 'go test' command may run tests for different packages
	    in parallel as well, according to the setting of the -p flag
	    (see 'go help build').
```

Adding `-p 1` to the `go test` command ensure each tests to be run separately but highly increase the testing time.
We could also try to use the `t.Parallel` call in each test checking for `Cleanup`
Lastly, we can also force different FlowCollector instance name. It's hacky but may work here.

controllers/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -803,3 +803,184 @@ filters:
...
+  - name: TimeFlowRttNs
+    type: number
+    description: TCP Smoothed Round Trip Time (SRTT), in nanoseconds
+  - name: K8S_ClusterName
+    type: string
+    description: Cluster name or identifier
+    lokiLabel: true
+  - name: _RecordType
+    type: string
+    description: "Type of record: 'flowLog' for regular flow logs, or 'allConnections', 'newConnection', 'heartbeat', 'endConnection' for conversation tracking"
```
jpinsonneau, line 982:
```suggestion
  - name: _RecordType
    type: string
    description: "Type of record: 'flowLog' for regular flow logs, or 'newConnection', 'heartbeat', 'endConnection' for conversation tracking"
```

controllers/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -803,3 +803,184 @@ filters:
...
+    type: number
+    description: Number of packets dropped by the kernel
+  - name: PktDropLatestState
+    type: string
+    description: TCP state on last dropped packet
+    filter: pkt_drop_state # couldn't guess from config
+  - name: PktDropLatestDropCause
+    type: string
+    description: Latest drop cause
+    filter: pkt_drop_cause # couldn't guess from config
```
jpinsonneau, line 954:
```suggestion
  - name: PktDropLatestState
    type: string
    description: TCP state on last dropped packet
  - name: PktDropLatestDropCause
    type: string
    description: Latest drop cause
```

We keep filter definition in the columns right ?

controllers/consoleplugin/config/config.go
```diff
@@ -62,6 +62,14 @@ type FilterConfig struct {
 	Placeholder            string `yaml:"placeholder,omitempty" json:"placeholder,omitempty"`
 }
 
+type FieldConfig struct {
+	Name        string `yaml:"name" json:"name"`
+	Type        string `yaml:"type" json:"type"`
+	Description string `yaml:"description" json:"description"`
+	LokiLabel   bool   `yaml:"lokiLabel,omitempty" json:"lokiLabel,omitempty"`
+	Filter      string `yaml:"filter,omitempty" json:"filter,omitempty"`
+}
```
jpinsonneau, line 71:
```suggestion
type FieldConfig struct {
	Name        string `yaml:"name" json:"name"`
	Type        string `yaml:"type" json:"type"`
	Description string `yaml:"description" json:"description"`
	LokiLabel   bool   `yaml:"lokiLabel,omitempty" json:"lokiLabel,omitempty"`
}
```

I don't think we should define filters in the field configuration. Filters are already defined in their own section and tied to columns when needed.

controllers/flowcollector_controller_certificates_test.go
jpinsonneau: FYI [this PR](https://github.com/netobserv/network-observability-operator/pull/510
) moved FLP and Monitoring controller tests outside of `controllers` package so we parallelize creation / deletion of `FlowCollector` between the 3 packages.

controllers/consoleplugin/config/config.go
jotak: they aren't defined from there, it's only a reference. In fact, for almost all the fields I didn't need that reference as I could [retrieve them from the columns](https://github.com/netobserv/network-observability-operator/pull/550/files#diff-1b179606477c1b69fe935bd9c8537d29e36b9996c3bb6959af1d6c21c113b53bR24). But there are two exceptions: for drop state and drop cause. These ones don't have the filter defined in column, I guess they must be some kind of special cases in the front implementation? Because of them, as I want to document the existence of these filters, I had to add an explicit reference.

jotak: I can rename the field `filterRef` to clarify

hack/asciidoc-flows-gen.sh
```diff
@@ -1,38 +1,44 @@
...
+  desc=$(printf "$entry" | yq ".description")
+  filter=$(printf "$entry" | yq ".filter")
+  if [[ "$filter" == "null" ]]; then
+    filter=$(yq ".columns[] | select(.field == \"$name\").filter" $SOURCE | sed 's/null//')
+    if [[ "$filter" == "" ]]; then
+      filter="n/a"
+    else
+      filter="\`$filter\`"
+    fi
+  fi
```
jpinsonneau, line 30:
```suggestion
  else 
    filter="\`$filter\`"
  fi
```

image

controllers/flowcollector_controller_certificates_test.go
jotak: So, after many tests, I didn't find issues with parallelism (although I don't guarantee that there aren't any). My guess is that it's ok with parallelism because each controller set up its own envtest environment so they run in isolation. And tbh if this isn't the case, I'm quite surprised to not see many more failures at random places, whereas here this is always the same single test that is failing.
Anyway my understanding of this test failure is finally that this is a timing issue, the CR taking sometimes more time to be actually deleted than the test timeout. So I have this PR for a fix: https://github.com/netobserv/network-observability-operator/pull/557 (and I'll remove my change here)

controllers/consoleplugin/config/config.go
jotak: I don't see filters for `TimeFlowStartMs` / `TimeFlowEndMs` or `Duplicate`, while they exist for PktDrop stuff (in the `filters` yaml)

jotak: Perhaps a refactoring will be necessary moving forward, but for this PR I didn't intend to do any code change, just extracting whatever data I can for the doc

controllers/flowcollector_controller_certificates_test.go
jotak: I reverted the change here, cf the mentioned PR above

jotak:
@jpinsonneau PR updated (doc regenerated after your updates, reverted flaky test change)

jpinsonneau (approved):
Looks good for the doc, thanks @jotak !

jotak:
thanks
/approve

docs/flows-format.adoc
skrthomas: re-copied and pushed latest to my PR.

PR #554, opened 2024-02-01
NETOBSERV-1458: reduce loki streams, index FlowLayer

jotak:


jotak:
cc @msherif1234 @nathan-weinberg  : I would like to compare perf with this set of loki index, against the previous runs and also against this other PR https://github.com/netobserv/network-observability-operator/pull/552

jotak:
Some frontend perf test --
I've been following the same steps that I ran previously and described here: https://github.com/netobserv/network-observability-operator/pull/409#issuecomment-1709875727

```
---------------------
1st RUN: main
---------------------
streams: ~2K
---------------------
min: 171ms
max: 393ms
avg: ~210ms
=====================
---------------------
2nd RUN: 88ec5d6
Remove some labels
---------------------
streams: ~1.2K
---------------------
min: 1321ms
max: 2784ms
avg: ~1800ms
=====================
---------------------
3rd RUN: 405439d
New indexes
---------------------
streams: ~1.3K
---------------------
min: 202ms
max: 399ms
avg: ~210ms
=====================
---------------------
4th RUN: local build
Only removing Duplicate
---------------------
streams: ~1.3K
---------------------
min: 253ms
max: 493ms
avg: ~300ms
=====================
```

To summarize: it's clear that the labels introduced in 1.5 caused an increased number of streams (RUN 1). We can remove some labels to reduce that number:
- Removing both Duplicate and Src/DstType (RUN 2), like it was in 1.4, degrades too much the query performance (x10 slower)
- Removing only Duplicate (RUN 4) makes us retrieve almost the same number of streams than previously, while keeping pretty good query performance
- Removing Duplicate and Adding FlowLayer (RUN 3 / this PR) is comparable to RUN 4 for streams but with a slightly better query performance

nathan-weinberg:
@jotak ack will test this! see table in [NETOBSERV-1470](https://issues.redhat.com/browse/NETOBSERV-1470) for comparisons

jpinsonneau (approved):
Sounds good to me; let's wait @nathan-weinberg feedback on the perfs before merging

PR #557, opened 2024-02-02
Fix envtests with FlowCollector deletion

jotak:
There was is timing issue with FlowCollector taking more time than the test timeout to be deleted. One option could have been to increase the timeout, but here we're simply checking that the DeletionTimestamp field is set, which means the resource has been marked for deletion.

Another issue found while testing, was the eBPF agent potentially calling `status.CheckDaemonSetProgress` with a nil parameter, which triggered an error. Adding nil-checks to StatusManager functions.


jotak:
this time, I think I've done it... damn test!

jpinsonneau (approved):
Let's give a try in the CI. Code looks good and tests are passing locally, thanks !

PR #566, opened 2024-02-12
NETOBSERV-1470: reduce memory usage in agent due to kafka batches [1.5 backport]

jotak:
Backport of https://github.com/netobserv/network-observability-operator/pull/567

As shown [here](https://docs.google.com/spreadsheets/d/1qakBaK1dk_rERO30k1cSR4W-Nn0SXW4A3lqQ1sZC4rE/edit#gid=1155551118) in run11, this tuning drastically reduces memory used due to kafka batching, retrieving similar level than in 1.4 (even slightly better).

It should be verified on large-scale perf tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ On the Loki server side, configuration differs depending on how Loki was install
 
 More performance fine-tuning is possible when using Kafka, ie. with `spec.deploymentModel` set to `KAFKA`:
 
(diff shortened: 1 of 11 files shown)


nathan-weinberg:
I will test this once regression testing is complete, please attach and make a note of it in NETOBSERV-1470 so it can be properly tracked as discussed this morning, thanks

PR #568, opened 2024-02-13
doc feedback

jotak:
@skrthomas FYI
cc @memodi @nathan-weinberg


jotak:
(For info it's not needed to backport this PR, it can be merged just in main)

skrthomas (commented):
Thanks for making these updates Joel. I'm glad it doesn't need to be backported 😅 lgtm!

skrthomas:
For the GitHub link: https://github.com/openshift/openshift-docs/pull/71033#discussion_r1486324614, since the style of product docs doesn't really play nicely for the upstream documentation, I can amend the link in the product doc version only. Otherwise, I'm double checking with the docs team about this issue. I feel its a really good example for why we need to be able to put links in modules. It sucks because we literally have this same reference replicated in product docs but we can't create a proper link :(

jotak:
@skrthomas thanks!
I added @nathan-weinberg 's suggestion about "use at your own risk", so I think we're good now

nathan-weinberg (approved):
LGTM!

PR #569, opened 2024-02-16
NETOBSERV-1508 add selector, afinity and priority to components

jpinsonneau:
## Description

This PR expose `NodeSelector`, `Affinity` and `PriorityClassName` on each netobserv component in their relative `advanced` sections.
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector

Example on a 2 Nodes cluster:

```bash
oc get nodes
NAME                         STATUS   ROLES    AGE   VERSION
ip-10-0-1-175.ec2.internal   Ready    worker   52m   v1.27.9+e36e183
ip-10-0-1-238.ec2.internal   Ready    worker   51m   v1.27.9+e36e183

oc get pods -n netobserv-privileged
NAME                         READY   STATUS    RESTARTS   AGE
netobserv-ebpf-agent-qs4f7   1/1     Running   0          4s
netobserv-ebpf-agent-t2b92   1/1     Running   0          4s
```

Set a custom annotation on a node:
```yaml
kind: Node
apiVersion: v1
metadata:
  name: ip-10-0-1-238.ec2.internal
  labels:
    netobserv: 'true'
...
```

Set nodeSelector on eBPF agent config:
```yaml
apiVersion: flows.netobserv.io/v1beta2
kind: FlowCollector
spec:
  agent:
    ebpf:
      advanced:
        nodeSelector:
          netobserv: 'true'
...
```

```bash
oc get pods -n netobserv-privileged
NAME                         READY   STATUS    RESTARTS   AGE
netobserv-ebpf-agent-hwp4c   1/1     Running   0          7m46s
```

## Dependencies

n/a

## Checklist


apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -864,6 +864,26 @@ type AdvancedAgentConfig struct {
 	// in edge debug or support scenarios.
 	//+optional
 	Env map[string]string `json:"env,omitempty"`
+
+	// NodeSelector is a selector which must be true for the pod to fit on a node.
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	// +mapType=atomic
+	NodeSelector map[string]string `json:"nodeSelector,omitempty" protobuf:"bytes,7,rep,name=nodeSelector"`
```
msherif1234, line 873: I would use `metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"` type for nodeSelector
`NodeSelector metav1.LabelSelector`

Makefile
```diff
@@ -361,7 +361,7 @@ deploy: kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/c
 	$(SED) -i -r 's~ebpf-agent:.+~ebpf-agent:main~' ./config/manager/manager.yaml
 	$(SED) -i -r 's~flowlogs-pipeline:.+~flowlogs-pipeline:main~' ./config/manager/manager.yaml
 	$(SED) -i -r 's~console-plugin:.+~console-plugin:main~' ./config/manager/manager.yaml
-	$(KUSTOMIZE) build config/openshift | sed -r "s/openshift-netobserv-operator\.svc/${NAMESPACE}.svc/" | kubectl apply -f -
+	$(KUSTOMIZE) build config/openshift | sed -r "s/openshift-netobserv-operator\.svc/${NAMESPACE}.svc/" | kubectl create -f -
```
msherif1234, line 364: kubectl folks recommend using apply over create IIRC they want to find ways to remove `create`

apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -924,6 +944,26 @@ type AdvancedProcessorConfig struct {
 	//+optional
 	// `conversationTerminatingTimeout` is the time to wait from detected FIN flag to end a conversation. Only relevant for TCP flows.
 	ConversationTerminatingTimeout *metav1.Duration `json:"conversationTerminatingTimeout,omitempty"`
+
+	// NodeSelector is a selector which must be true for the pod to fit on a node.
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	// +mapType=atomic
+	NodeSelector map[string]string `json:"nodeSelector,omitempty" protobuf:"bytes,7,rep,name=nodeSelector"`
```
msherif1234, line 953: same here

apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -981,6 +1021,26 @@ type AdvancedPluginConfig struct {
 	//+optional
 	// `port` is the plugin service port. Do not use 9002, which is reserved for metrics.
 	Port *int32 `json:"port,omitempty"`
+
+	// NodeSelector is a selector which must be true for the pod to fit on a node.
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	// +mapType=atomic
+	NodeSelector map[string]string `json:"nodeSelector,omitempty" protobuf:"bytes,7,rep,name=nodeSelector"`
```
msherif1234, line 1030: and here

Makefile
jpinsonneau: We fall in the following issue keeping `apply` here:
```
metadata.annotations: Too long: must have at most 262144 bytes
```

msherif1234: this is weird which md is it referring to here svc which version of kubectl u have ?

jpinsonneau: I was on an old version. Just updated to latest and still have the same issue:
```
Client Version: v1.29.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.9+e36e183
```

```
USER=jpinsonn VERSION=1508 make deploy
cd config/manager && /home/julien/dev/me/network-observability-operator/bin/kustomize edit set image controller=quay.io/jpinsonn/network-observability-operator:1508-3
sed -i -r 's~ebpf-agent:.+~ebpf-agent:main~' ./config/manager/manager.yaml
sed -i -r 's~flowlogs-pipeline:.+~flowlogs-pipeline:main~' ./config/manager/manager.yaml
sed -i -r 's~console-plugin:.+~console-plugin:main~' ./config/manager/manager.yaml
/home/julien/dev/me/network-observability-operator/bin/kustomize build config/openshift | sed -r "s/openshift-netobserv-operator\.svc/netobserv.svc/" | kubectl apply -f -
namespace/netobserv created
customresourcedefinition.apiextensions.k8s.io/flowmetrics.flows.netobserv.io created
serviceaccount/netobserv-controller-manager created
role.rbac.authorization.k8s.io/netobserv-leader-election-role created
role.rbac.authorization.k8s.io/netobserv-openshift-netobserv-operator-prometheus created
clusterrole.rbac.authorization.k8s.io/netobserv-manager-role created
clusterrole.rbac.authorization.k8s.io/netobserv-proxy-role created
rolebinding.rbac.authorization.k8s.io/netobserv-leader-election-rolebinding created
rolebinding.rbac.authorization.k8s.io/netobserv-openshift-netobserv-operator-prometheus created
clusterrolebinding.rbac.authorization.k8s.io/netobserv-manager-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/netobserv-proxy-rolebinding created
configmap/netobserv-manager-config created
service/netobserv-metrics-service created
service/netobserv-webhook-service created
deployment.apps/netobserv-controller-manager created
servicemonitor.monitoring.coreos.com/netobserv-metrics-monitor created
validatingwebhookconfiguration.admissionregistration.k8s.io/netobserv-validating-webhook-configuration created
The CustomResourceDefinition "flowcollectors.flows.netobserv.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```

apis/flowcollector/v1beta2/flowcollector_types.go
jpinsonneau: As discussed I'm setting `corev1.PodSpec` attributes for each components. These take `map[string]string` for `NodeSelector`.

Makefile
jotak: that must be the `alm-example`, I guess? Perhaps we should try to shrink that a bit

apis/flowcollector/v1beta1/flowcollector_webhook.go
```diff
@@ -52,6 +52,32 @@ func (r *FlowCollector) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.Loki.Microservices = restored.Spec.Loki.Microservices
 	dst.Spec.Loki.Manual = restored.Spec.Loki.Manual
 
+	if restored.Spec.Agent.EBPF.Advanced != nil {
```
jotak, line 55: Rather than adding more complex conversion logic (sometimes error prone), I find it generally easier to just backport a feature to previous versions. As long as we're not modifying or deleting existing parts of the API, I think it's just fine to backport it to v1beta1?

pkg/helper/comparators.go
```diff
@@ -57,6 +92,28 @@ func annotationsChanged(old, new *corev1.PodTemplateSpec, report *ChangeReport)
 	return false
 }
 
+func assignationChanged(old, new *corev1.PodTemplateSpec, report *ChangeReport) bool {
+	if !equality.Semantic.DeepDerivative(old.Spec.NodeSelector, new.Spec.NodeSelector) {
```
jotak, line 96: `DeepDerivative` normally takes the "new" param as a first argument and the "old" as the second; and it checks for stuff in "new" that wasn't  in "old", but not the other way around. That's typically useful for annotations/labels, when other parts of the kube infra might add their own labels / annotations on pods and we don't want to interfere with that.

So here I believe either we want to use `DeepDerivative` but then we should change the arguments order, or perhaps we actually want `DeepEqual` here ?

Makefile
jotak: Like we shrinked alpha1 sample to its simplest minimal form, I guess we can do the same with beta1:

```yaml
apiVersion: flows.netobserv.io/v1beta1
kind: FlowCollector
metadata:
  name: cluster
spec:
  deploymentModel: DIRECT
```

Keeping all the details and examples in beta2 only

jpinsonneau: I would rather remove the old samples from `kustomization.yaml` than simplifying them in that case.
We will gain more bytes and we'll keep valid examples in the repo.

WDYT ?

apis/flowcollector/v1beta1/flowcollector_webhook.go
jpinsonneau: Are you suggesting to add the whole `advanced` sections to every components here ?

Currently in `v1beta1`, only eBPF and flp components have `debug` sections containing a single `Env` map.

pkg/helper/comparators.go
jpinsonneau: I gave a try with `DeepEqual` here and got a infinite reconcile loop. It seemed to work fine with `DeepDerivative` but I'll give another try inverting the two args to see if I get differences.

Makefile
jotak: We can't remove them, OLM asks that every served version has an example. Actually, I'm removing alpha1 via [this](https://issues.redhat.com/browse/NETOBSERV-1513) ; but beta1 has to stay

apis/flowcollector/v1beta1/flowcollector_webhook.go
jotak: no that's not what I was suggesting ... iirc `advanced` brought some breaking changes so it had to go via a new API and could not be backported; I was more thinking about adding these fields at the component root like we used to do in beta1, but if you think it won't simplify the things, feel free to keep it like you did

pkg/helper/comparators.go
jpinsonneau: I have addressed your feedback and simplified the code in this commit: https://github.com/netobserv/network-observability-operator/pull/569/commits/4178913ec45ffaab1183d4941bf84583389f61bc

Thanks !

Makefile
jpinsonneau: @jotak I get the same error after shrinking `v1beta1` as you suggested

jpinsonneau: The `corev1.Affinity` object is actually too complex and adds **1071 rows** in the manifest for each components.
We can remove this object to solve the issue here but keep in mind that this may force customers to set custom labels on their nodes just for us.
Also, `Pod Affinity / Anti Affinity` helps to get or avoid two pods in the same node. This is complementary to `NodeSelector`.

I feel we will fall into that situation anyway in the future. WDYT ?

jotak: what if you rebase on https://github.com/netobserv/network-observability-operator/pull/577 ?

jpinsonneau: Just tested and same :disappointed:

jpinsonneau:
Rebased without changes

Makefile
jotak: FWIW I did some search around this problem and this is something other people has brought up, apparently a common approach to mitigate that is to remove the description fields on embedded APIs (like Affinity or Autoscaler); there's a ticket for allowing this via kubebuilder but it's currently still open https://github.com/kubernetes-sigs/controller-tools/issues/441 (since 2020).

I tried this simple workaround that seems to work well:

(in our Makefile under `manifests`)
```bash
##@ Code / files generation
manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
	$(YQ) -i 'del(.spec.versions[].schema.openAPIV3Schema.properties.spec.properties.processor.properties.kafkaConsumerAutoscaler.properties.metrics.items | .. | select(has("description")) | .description)' config/crd/bases/flows.netobserv.io_flowcollectors.yaml
	$(YQ) -i 'del(.spec.versions[].schema.openAPIV3Schema.properties.spec.properties.consolePlugin.properties.autoscaler.properties.metrics.items | .. | select(has("description")) | .description)' config/crd/bases/flows.netobserv.io_flowcollectors.yaml
	$(YQ) -i 'del(.spec.versions[].schema.openAPIV3Schema.properties.spec.properties.agent.properties.ebpf.properties.advanced.properties.affinity.properties | .. | select(has("description")) | .description)' config/crd/bases/flows.netobserv.io_flowcollectors.yaml
	$(YQ) -i 'del(.spec.versions[].schema.openAPIV3Schema.properties.spec.properties.processor.properties.advanced.properties.affinity.properties | .. | select(has("description")) | .description)' config/crd/bases/flows.netobserv.io_flowcollectors.yaml
	$(YQ) -i 'del(.spec.versions[].schema.openAPIV3Schema.properties.spec.properties.consolePlugin.properties.advanced.properties.affinity.properties | .. | select(has("description")) | .description)' config/crd/bases/flows.netobserv.io_flowcollectors.yaml

```

It makes the CRD shrinks from ~592kB to ~321kB so not too bad .. although this isn't a definitive solution

jotak: (and for the record, the issue comes from `kubectl apply` injecting the `last-applied-configuration`, which `kubectl create` does not; it's this annotation that is too big since it contains the whole resource)

jpinsonneau:
> @jpinsonneau @msherif1234 : on the `apply` vs `create`, I'm proposing this workaround: [jpinsonneau#4](https://github.com/jpinsonneau/network-observability-operator/pull/4) This removes external/embedded resource descriptions, which allows shrinking the CRD & annotation size and makes the `apply` pass again.

I guess we loose autocompletion in the UI with this approach right ? 
But if it's making the CRD lightweight I'm fine with that. We just need to document with examples these sections so the users are not lost

jotak:
> I guess we loose autocompletion in the UI with this approach right ? But if it's making the CRD lightweight I'm fine with that. We just need to document with examples these sections so the users are not lost

Yes .. or perhaps what we could do, instead of remove the descriptions, is to put a link to these object's doc, such as https://docs.openshift.com/container-platform/4.15/rest_api/autoscale_apis/horizontalpodautoscaler-autoscaling-v2.html

jpinsonneau:
Rebased without changes

jotak (approved):

PR #582, opened 2024-02-26
NETOBSERV-557: add eBPF agent metrics for troubleshooting

msherif1234:
## Description

adding promo svc to ebpf agent for debugging metrics support
## Dependencies

n/a

## Checklist


apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -239,6 +239,9 @@ type FlowCollectorEBPF struct {
 	// - `FlowRTT` [unsupported (*)]: enable flow latency (RTT) calculations in the eBPF agent during TCP handshakes. This feature better works with `sampling` set to 1.<br>
 	// +optional
 	Features []AgentFeature `json:"features,omitempty"`
+	// Metrics server endpoint configuration for Prometheus scraper
+	// +optional
+	Server MetricsServerConfig `json:"server,omitempty"`
```
jotak, line 244: could it be called `metricsServer`? Either that, or have an intermediate `metrics` struct (it might be useful if we add more metrics settings later)

controllers/ebpf/agent-metrics.go
```diff
@@ -0,0 +1,90 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func (c *AgentController) ReconcileMetricsService(ctx context.Context, target *flowslatest.FlowCollectorEBPF) error {
+	report := helper.NewChangeReport("EBPF Agent prometheus service")
+	defer report.LogIfNeeded(ctx)
+
+	if err := c.ReconcileService(ctx, c.promSvc, c.promService(target), &report); err != nil && !errors.IsAlreadyExists(err) {
```
jotak, line 22: why `IsAlreadyExists` is checked here? Sounds like not propagating this error could cause missed reconcile events?

controllers/ebpf/agent_controller.go
```diff
@@ -134,7 +142,10 @@ func (c *AgentController) Reconcile(ctx context.Context, target *flowslatest.Flo
 	if err != nil {
 		return err
 	}
-
+	err = c.reconcilePrometheusService(ctx, &target.Spec.Agent.EBPF)
```
jotak, line 145: there's not only the service to reconcile here, but also the agent pod config (ENV) to pass the configured server port.
I would also add a config to enable metric
And I think we should set the metrics prefix to "netobserv_agent_"

jotak:
FYI I also added a follow-up to implement TLS: https://issues.redhat.com/browse/NETOBSERV-1532

apis/flowcollector/v1beta1/flowcollector_types.go
msherif1234: done

controllers/ebpf/agent_controller.go
msherif1234: yeah was holding on that piece till I get the promo pieces in I added that it now

controllers/ebpf/agent-metrics.go
msherif1234: agree removed the check for alreadyexsists

jotak:
@msherif1234 I don't think this is for debugging only, all our other components provide some metrics even by default in production, TBH I don't think it's a big overhead but we can see & measure that. All these metrics are low cardinality they're just a bunch of float64's values fetched from time to time, it shouldn't be big

For now we can go ahead with that disabled by default but if it doesn't have a visible overhead we can think about enabling it by default

PR #593, opened 2024-03-21
NETOBSERV-1568: reintroduce finalizer check to remove any remaining finalizer

jotak:
## Description

Since removing ipfix, we don't use finalizers anymore. However users upgrading from a 1.5 will still have one set in their CR. So we need, for a release, to continue checking for the finalizer, and remove it if found.

## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/flowcollector_controller.go
+++ b/controllers/flowcollector_controller.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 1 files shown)


memodi:
@jotak - I tried to upgrade test with pre-merge catalog, but unable with downstream 1.5 and upstream pre-merge, QE testing would have to occur on downstream bundle for this change.

jotak:
> @jotak - I tried to upgrade test with pre-merge catalog, but unable with downstream 1.5 and upstream pre-merge, QE testing would have to occur on downstream bundle for this change.

What I did to test was installing downstream 1.5, then downstream 1.6 (the one that is buggy) and then I manually changed the operator image from the installed CSV to the upstream one.
Otherwise, to switch from a downstream catalog to an upstream one, IIRC there's a couple of more things to do like replacing the "replaces" key in CSV

memodi:
> What I did to test was installing downstream 1.5, then downstream 1.6 (the one that is buggy) and then I manually changed the operator image from the installed CSV to the upstream one. And then I was able to delete the FlowCollector.

thanks @jotak . I was able to verify with this approach.

/label qe-approved

PR #613, opened 2024-04-10
 NETOBSERV-739: Add prometheus

jotak:
## Description

Adds a new config section in CRD to allow reading metrics via prometheus ( / thanos) :

- Auto mode: (when used in openshift, automatically uses monitoring's thanos)

```yaml
  prometheus:
    querier:
      enable: true
      mode: Auto
      timeout: 30s
```

- Manual mode: to manually configure prometheus endpoint / TLS etc.

This is enabled by default.

## Dependencies

- https://github.com/netobserv/network-observability-console-plugin/pull/516

## Checklist


apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -60,6 +60,9 @@ type FlowCollectorSpec struct {
 	// `loki`, the flow store, client settings.
 	Loki FlowCollectorLoki `json:"loki,omitempty"`
 
+	// `prometheusClient` defines Prometheus client settings, used to fetch metrics from the Console plugin.
+	PrometheusClient FlowCollectorPrometheusClient `json:"prometheusClient,omitempty"`
+
```
jpinsonneau, line 65:
```suggestion
	// `prometheus` defines Prometheus client settings, used to fetch metrics from the Console plugin.
	Prometheus FlowCollectorPrometheus `json:"prometheus,omitempty"`
```

For consistency, I feel it would be better to simply have `prometheus` as same as `loki`

jpinsonneau: that could contains FLP prometheus settings in future

jpinsonneau: That makes sense indeed. And we should also end up allowing users to disable prometheus export from FLP too indeed

jotak: the current way to do this would be to set an empty `includeList` of metrics

Makefile
```diff
@@ -364,7 +364,7 @@ deploy: kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/c
 	$(SED) -i -r 's~ebpf-agent:.+~ebpf-agent:main~' ./config/manager/manager.yaml
 	$(SED) -i -r 's~flowlogs-pipeline:.+~flowlogs-pipeline:main~' ./config/manager/manager.yaml
 	$(SED) -i -r 's~console-plugin:.+~console-plugin:main~' ./config/manager/manager.yaml
-	$(KUSTOMIZE) build config/openshift | sed -r "s/openshift-netobserv-operator\.svc/${NAMESPACE}.svc/" | kubectl apply -f -
+	$(KUSTOMIZE) build config/openshift | sed -r "s/openshift-netobserv-operator\.svc/${NAMESPACE}.svc/" | kubectl create -f -
```
jotak, line 367: hey @jpinsonneau @msherif1234 @OlivierCazade : I'm hitting again the same issue that Julien had previously, with `kubectl apply` failing due to send too large annotations. (and `create` passes)
So temporarily I'm using create; but we should find a more durable fix for that (previously we did remove some of the CRD descriptions to make it pass, but it was just a matter of time to see the problem rise again)

Are we too verbose in the doc? Is the CRD too big? idk how to fix that. Or perhaps we should just work a bit more with the Makefile to do `create` / `delete` with the CRD (instead of `apply`) and keep `apply` for all other resources... Any idea?

msherif1234: try with `kubectl apply --server-side` see if that helps I added more comments in
https://issues.redhat.com/browse/NETOBSERV-1613

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -716,6 +724,64 @@ type FlowCollectorLoki struct {
...
+
+// `PrometheusQuerier` defines the desired state for querying Prometheus (client...)
+type PrometheusQuerier struct {
+	// Set `enable` to `true` to make the Console plugin querying flow metrics from Prometheus instead of Loki whenever possible.
+	// The Console plugin can use either Loki or Prometheus as a data source for metrics (see also `spec.loki`), or both.
+	// Not all queries are transposable from Loki to Prometheus. Hence, if Loki is disabled, some features of the plugin are disabled as well,
+	// such as getting per-pod information or viewing raw flows.
+	// If both Prometheus and Loki are enabled, Prometheus takes precedence and Loki is used as a fallback for queries that Prometheus cannot handle.
+	// If they are both disabled, the Console plugin is not deployed.
+	//+kubebuilder:default:=true
```
msherif1234, line 763: // +optional

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -716,6 +724,64 @@ type FlowCollectorLoki struct {
...
+	Enable *bool `json:"enable,omitempty"`
+
+	// `mode` must be set according to the type of Prometheus installation that stores NetObserv metrics:<br>
+	// - Use `Auto` to try configuring automatically. In OpenShift, it uses the Thanos querier from OpenShift Cluster Monitoring<br>
+	// - Use `Manual` for a manual setup<br>
+	//+unionDiscriminator
+	//+kubebuilder:validation:Enum=Manual;Auto
+	//+kubebuilder:default:="Auto"
+	//+kubebuilder:validation:Required
+	Mode PrometheusMode `json:"mode,omitempty"`
```
msherif1234, line 773: same ?

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -347,9 +356,77 @@ func (b *builder) setLokiConfig(lconf *cfg.LokiConfig) {
...
+	// Default config = manual
+	tls := b.desired.Prometheus.Querier.Manual.TLS
+	config := cfg.PrometheusConfig{
+		URL:              b.desired.Prometheus.Querier.Manual.URL,
+		ForwardUserToken: b.desired.Prometheus.Querier.Manual.ForwardUserToken,
+	}
+	if b.desired.Prometheus.Querier.Mode == flowslatest.PromModeAuto {
+		if b.info.UseOpenShiftSCC /* aka IsOpenShift */ {
+			config = cfg.PrometheusConfig{
+				URL:              "https://thanos-querier.openshift-monitoring.svc:9091/",
```
msherif1234, line 379: r u sure the svc will always be at `9091` any better way than hardcoding it ?

controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -347,9 +356,77 @@ func (b *builder) setLokiConfig(lconf *cfg.LokiConfig) {
...
+				config.CAPath = caPath
+			}
+		}
+	}
+
+	if b.desired.Prometheus.Querier.Timeout != nil {
+		config.Timeout = api.Duration{Duration: b.desired.Prometheus.Querier.Timeout.Duration}
+	} else {
+		config.Timeout = api.Duration{Duration: 30 * time.Second}
+	}
```
msherif1234, line 410: nit: avoid else
```go
config.Timeout = api.Duration{Duration: 30 * time.Second}
if b.desired.Prometheus.Querier.Timeout != nil {
   config.Timeout = api.Duration{Duration: b.desired.Prometheus.Querier.Timeout.Duration}
}
```

controllers/consoleplugin/consoleplugin_objects.go
jotak: We could query for the service object, but we cannot avoid some hardcoded stuff (if not the port number, it will be the port name in YAML, and the service name itself). So, I'm not sure it's worth it. And anyway that's just the "Auto" mode: if the hardcoded config doesn't work anymore later, users can switch to manual config.

msherif1234: I guess we can start like that and change when we see issues ports tends to get reused by different app sometimes can also change on different arch like we have seen with metrics on ppc64le, since u have a log stating we can't do auto mode and fall back to manual that will be good enough for now

apis/flowcollector/v1beta1/flowcollector_types.go
jotak: fields are optional by default (you can make the test, marking this field as optional won't do anything when regenerating the crd openapi) - see also https://book.kubebuilder.io/reference/markers/crd-validation.html
=>
> // +optional
> specifies that this field is optional, **if fields are required by default**.

jotak:
@memodi FYI: last build adds fix for NETOBSERV-1654 and also a small perf improvement, using exact match with the default quick filters instead of partial matching

config/samples/flows_v1beta2_flowcollector.yaml
```diff
@@ -137,19 +142,19 @@ spec:
...
       filter:
-        src_kind: 'Pod'
-        dst_kind: 'Pod'
+        src_kind: '"Pod"'
+        dst_kind: '"Pod"'
       default: true
     - name: Services network
       filter:
-        dst_kind: 'Service'
+        dst_kind: '"Service"'
```
msherif1234, line 157: this is bit ugly and not user friendly why did u change the APIs
https://github.com/netobserv/network-observability-operator/pull/613/files#diff-1aad53ab28e869c35c03e94f19d4e37402b869c34eabfd55317931ab5e2b1436R863 ?

msherif1234:
another than the API question above
/LGTM

config/samples/flows_v1beta2_flowcollector.yaml
msherif1234: cc'd @memodi

memodi:
> @memodi FYI: last build adds fix for [NETOBSERV-1654](https://issues.redhat.com//browse/NETOBSERV-1654) and also a small perf improvement, using exact match with the default quick filters instead of partial matching

this would need update to our tests that verifies "filters"

config/samples/flows_v1beta2_flowcollector.yaml
jotak: it's a perf improvement for queries, have quotes means doing an exact match instead of a partial match, which is more performant

jotak:
> this would need update to our tests that verifies "filters"

If you feel this shouldn't be in this PR I can remove it, but it seems to me it's a quick & easy way to improve query perf

memodi:
> > this would need update to our tests that verifies "filters"
> 
> If you feel this shouldn't be in this PR I can remove it, but it seems to me it's a quick & easy way to improve query perf

no worries, it's okay to have it in this PR, I just wanted to make a note of it.

memodi:
@jotak the other thing I am noticing is below when flowcollector is created newly, plugin goes from Running --> Terminating --> Running state and I see below events for plugin before it stabilizes,  not sure why it's scale down and back up:
```
netobserv-plugin-7d76c88489-89cf5   1/1     Running             0          5s
netobserv-plugin-bf7d457d8-rgv2z    1/1     Terminating         0          7s
netobserv-plugin-bf7d457d8-rgv2z    0/1     Terminating         0          7s
netobserv-plugin-bf7d457d8-rgv2z    0/1     Terminating         0          8s
netobserv-plugin-bf7d457d8-rgv2z    0/1     Terminating         0          8s
netobserv-plugin-bf7d457d8-rgv2z    0/1     Terminating         0          8s

$ oc get events
26s         Normal    ScalingReplicaSet        deployment/netobserv-plugin              Scaled up replica set netobserv-plugin-bf7d457d8 to 1
24s         Normal    ScalingReplicaSet        deployment/netobserv-plugin              Scaled up replica set netobserv-plugin-7d76c88489 to 1
21s         Normal    ScalingReplicaSet        deployment/netobserv-plugin              Scaled down replica set netobserv-plugin-bf7d457d8 to 0 from 1
```

jotak:
@memodi I don't reproduce this either :-(
Can you share the operator logs when that happens? And the FlowCollector config?
And if possible get the plugin's pod YAML before and after restart

memodi:
> @memodi I don't reproduce this either :-( Can you share the operator logs when that happens? And the FlowCollector config? And if possible get the plugin's pod YAML before and after restart

@jotak here's the must-gather tar file: https://drive.google.com/file/d/1O6eduOKzr8SqoxZyylJ1HKxsG8u9eZm9/view?usp=drive_link

memodi:
thanks @jotak , yes, we usually work with defaults in our tests as much as possible and turn on/off only if we need to for our testing. Last commit fixed the bug.

/label qe-approved

jotak:
unit tests are failing weirdly on CI, they work fine locally - I'll see on Monday if that's all clean before merging

PR #621, opened 2024-04-17
README / markdown updates

jotak:
Refresh a few outdated things in our readme and descriptions

Make it clear that Loki isn't a requirement.


OlivierCazade (approved):
LGTM, thanks!

PR #622, opened 2024-04-18
NETOBSERV-1614: Metrics API validation webhook

msherif1234:
## Description

Validation webhook for flow metrics resourcce
## Dependencies

n/a

## Checklist


apis/flowmetrics/v1alpha1/flowmetric_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+// based on the docs : https://docs.openshift.com/container-platform/latest/observability/network_observability/json-flows-format-reference.html.
+var validFieldsNames = []string{
+	"Bytes", "DnsErrno", "DnsFlags", "DnsFlagsResponseCode", "DnsId", "DnsLatencyMs",
+	"Dscp", "DstAddr", "DstK8S_HostIP", "DstK8S_HostName", "DstK8S_Name", "DstK8S_Namespace", "DstK8S_OwnerName",
+	"DstK8S_OwnerType", "DstK8S_Type", "DstK8S_Zone", "DstMac", "DstPort", "Duplicate", "Flags", "FlowDirection",
+	"IcmpCode", "IcmpType", "IfDirection", "Interface", "K8S_ClusterName", "K8S_FlowLayer", "Packets",
+	"PktDropBytes", "PktDropLatestDropCause", "PktDropLatestFlags", "PktDropLatestState", "PktDropPackets",
+	"Proto", "SrcAddr", "SrcK8S_HostIP", "SrcK8S_HostName", "SrcK8S_Name", "SrcK8S_Namespace", "SrcK8S_OwnerName", "SrcK8S_OwnerType",
+	"SrcK8S_Type", "SrcK8S_Zone", "SrcMac", "SrcPort", "TimeFlowEndMs", "TimeFlowStartMs", "TimeReceived", "TimeFlowRttNs",
+}
```
jotak, line 29: Instead of this hard-coded list, we can leverage the embedded frontend config which already contains the fields list: see here https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/consoleplugin_objects.go#L414-L415

If we create a `config.GetStaticConfig` func in consoleplugin package, we can get that same fields list, and we keep a single source of truth

jotak: cc @OlivierCazade I guess the same approach would be valid on your task

pkg/helper/helpers.go
```diff
@@ -88,3 +91,17 @@ func UnstructuredDuration(in *metav1.Duration) string {
 	}
 	return in.ToUnstructured().(string)
 }
+
+func FindFilter(name string) bool {
+	var cfg config.FrontendConfig
+	err := yaml.Unmarshal(config.StaticFrontendConfig, &cfg)
```
jotak, line 97: this unmarshal is called repeatedly (in `for` loops in `validateFlowMetric`), can we try have a single call for it?

apis/flowmetrics/v1alpha1/flowmetric_webhook.go
```diff
@@ -0,0 +1,77 @@
...
+
+// log is for logging in this package.
+var flowmetriclog = logf.Log.WithName("flowmetric-resource")
+
+type FlowMetricWebhook struct {
+	FlowMetric
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-netobserv-io-v1alpha1-flowmetric,mutating=false,failurePolicy=fail,sideEffects=None,groups=netobserv.io,resources=flowmetrics,versions=v1alpha1,name=flowmetricvalidationwebhook.netobserv.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:verbs=create;update,path=/validate-netobserv-io-v1beta2-flowcollector,mutating=false,failurePolicy=fail,groups=netobserv.io,resources=flowcollectors,versions=v1beta2,name=flowcollectorconversionwebhook.netobserv.io,sideEffects=None,admissionReviewVersions=v1
```
jotak, line 24: was that line added automatically with operator-sdk?
The first line is for flowmetric, but the second is for flowcollector, for which we don't have a validation webhook, so should we remove it?

msherif1234: w/o this 2nd line the generated manifest will have only one webhook so yes it seems related with conversion we can get both

memodi (commented):
@msherif1234 this overalls works well for fields set in labels and filters. I think we should apply validation for `valueField` as well, since it also takes values from one of the numeric fields from the flow or if could be left out to just count number of flows.

image

what do you think?

memodi (approved):
/label qe-approved

thanks @msherif1234

PR #628, opened 2024-04-25
NETOBSERV-1619: use dedicated metrics ports for netobserv to avoid conflicts

msherif1234:
## Description

use dedicated metrics ports for netobserv

## Dependencies

https://github.com/openshift/enhancements/pull/1612

## Checklist


msherif1234:
image

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -387,9 +387,8 @@ type MetricsServerConfig struct {
 
 	//+kubebuilder:validation:Minimum=1
 	//+kubebuilder:validation:Maximum=65535
-	//+kubebuilder:default:=9102
 	// The prometheus HTTP port
-	Port int32 `json:"port,omitempty"`
+	Port *int32 `json:"port,omitempty"`
```
OlivierCazade, line 391: I am not sure to understand why we have to change the approach here and to use a pointer instead of `kubebuilder:default`.

Most of the integer here are not pointer, I think it would be better to stay consistent but may be I am missing something.

msherif1234: it make the check if the field is provisioned or not so u can install default vs checking for port value of `0`

OlivierCazade: But isn't it the goal of `kubebuilder:default` to check if the field is provisioned and to set it to default if it is not?

OlivierCazade: I am not seeing the difference here, to my understanding, in both case, the user does not set the port, and in one case kubebuilder does the work of setting it to the default value and in the other one we check if it nil and set the default value manually.

OlivierCazade: Our messages crossed, thanks, for the explanation.

OlivierCazade (approved):

PR #629, opened 2024-04-26
NETOBSERV-1447: Added a check to log a warning if some labels are used in metric defi…

OlivierCazade:
## Description

Added a check to log a warning if some labels are used in metric definition

## Dependencies

n/a

## Checklist

### DIFF
--- a/apis/flowmetrics/v1alpha1/flowmetric_webhook.go
+++ b/apis/flowmetrics/v1alpha1/flowmetric_webhook.go
@@ -61,6 +61,16 @@ func (r *FlowMetricWebhook) ValidateDelete(_ context.Context, _ runtime.Object) 
 	return nil, nil
 }
 
+func checkFlowMetricCartinality(fMetric *FlowMetric) {
+	for _, label := range fMetric.Spec.Labels {
+		if helper.LabelIsHighCardinality(label) {
+			//No warning level logging as info since it is not an error
+			flowmetriclog.Info("Warning: metric label has high cardinality, please limit it by using some filters", "labelName", label)
+			return
+		}
+	}
+}
+
 func validateFlowMetric(_ context.Context, fMetric *FlowMetric) error {
 	var str []string
 	var allErrs field.ErrorList
@@ -95,5 +105,6 @@ func validateFlowMetric(_ context.Context, fMetric *FlowMetric) error {
 			schema.GroupKind{Group: GroupVersion.Group, Kind: FlowMetric{}.Kind},
 			fMetric.Name, allErrs)
 	}
+	checkFlowMetricCartinality(fMetric)
 	return nil
 }
--- a/pkg/helper/helpers.go
+++ b/pkg/helper/helpers.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/netobserv/network-observability-operator/controllers/consoleplugin/config"
 
-	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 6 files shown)


controllers/constants/constants.go
```diff
@@ -54,3 +54,4 @@ var EnvNoHTTP2 = corev1.EnvVar{
 	Name:  "GODEBUG",
 	Value: "http2server=0",
 }
+var MetricHighCardinalityLabels = []string{"DstAddr", "DstK8S_Name", "DstMac", "SrcAddr", "SrcK8S_Name", "SrcMac"}
```
msherif1234, line 57: can u instead of hardcoding them to look for
https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/config/config.go#L83 with cardinalty `avoid ` similar to how I look for valueField for field of type `number` ?

apis/flowmetrics/v1alpha1/flowmetric_webhook.go
```diff
@@ -61,6 +62,25 @@ func (r *FlowMetricWebhook) ValidateDelete(_ context.Context, _ runtime.Object)
...
+	}
+	return false
+}
+
+func checkFlowMetricCartinality(fMetric *FlowMetric) {
+	for _, label := range fMetric.Spec.Labels {
+		if labelIsHighCardinality(label) {
+			//No warning level logging as info since it is not an error
+			flowmetriclog.Info("Warning: metric label has high cardinality, please limit it by using some filters", "labelName", label)
+			return
```
msherif1234, line 79: pls add a test case for this check

controllers/constants/constants.go
OlivierCazade: Done!

memodi:
building cluster with this change for testing.

memodi:
I was able to see warning in Operator logs:

```
2024-05-07T16:47:43.402Z	INFO	flowmetric-resource	Warning: metric label has high cardinality, please limit it by using some filters	{"labelName": "DstAddr"}
```
I have some feedback for this warning to be more visible, we can have this PR merged.

/label qe-approved

pkg/helper/helpers.go
```diff
@@ -124,3 +124,18 @@ func FindFilter(labels []string, isNumber bool) bool {
 
 	return true
 }
+
+func LabelIsHighCardinality(label string) bool {
+	pluginCfg := config.PluginConfig{}
+	err := yaml.Unmarshal(config.LoadStaticFrontendConfig(), &pluginCfg.Frontend)
```
jotak, line 130: It would be good to cache the PluginConfig rather than unmarshaling it every time (not just here but also done in other places). Especially here that's done in a `for` loop for each label

msherif1234: we could leverage https://github.com/netobserv/network-observability-operator/blob/main/pkg/helper/helpers.go#L95 remove the for loop and pass in another pool `isHighCardinality` this will do unmarshall once ?

pkg/helper/helpers.go
```diff
@@ -124,3 +122,17 @@ func FindFilter(labels []string, isNumber bool) bool {
...
+	if err != nil {
+		return false
+	}
+	for _, cfgLabel := range frontendCfg.Fields {
+		if label == cfgLabel.Name {
+			return cfgLabel.CardinalityWarn == config.CardinalityWarnCareful ||
+				cfgLabel.CardinalityWarn == config.CardinalityWarnAvoid
+		}
+	}
+	return false
```
msherif1234, line 137: this works for me thank you!!

pkg/helper/helpers.go
OlivierCazade: I modified `LoadStaticFrontendConfig` to unmarshall inside the function and only once.

PR #632, opened 2024-04-29
NETOBSERV-1625: Add ebpf altering for flows drop

msherif1234:
## Description

Add alert for ebpf flows drop to detect when ebpf hash table is full

## Dependencies

n/a

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [X] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [X] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [X] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [ ] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #640, opened 2024-05-10
NETOBSERV-1625: (follow-up) mention other possible cause for ebpf drops

jotak:
## Dependencies

n/a

## Checklist

### DIFF
--- a/apis/flowcollector/v1beta1/flowcollector_types.go
+++ b/apis/flowcollector/v1beta1/flowcollector_types.go
@@ -159,7 +159,7 @@ const (
 
 // Name of an eBPF agent alert.
 // Possible values are:<br>
-// `NetObservDroppedFlows`, which is triggered when the eBPF agent is dropping flows, such as when the BPF hashmap is full.<br>
+// `NetObservDroppedFlows`, which is triggered when the eBPF agent is dropping flows, such as when the BPF hashmap is full or the capacity limiter being triggered.<br>
 // +kubebuilder:validation:Enum:="NetObservDroppedFlows"
 type EBPFAgentAlert string
 
@@ -178,7 +178,7 @@ type EBPFMetrics struct {
 
 	// `disableAlerts` is a list of alerts that should be disabled.
 	// Possible values are:<br>
-	// `NetObservDroppedFlows`, which is triggered when the eBPF agent is dropping flows, such as when the BPF hashmap is full.<br>
+	// `NetObservDroppedFlows`, which is triggered when the eBPF agent is dropping flows, such as when the BPF hashmap is full or the capacity limiter being triggered.<br>
 	// +optional
 	DisableAlerts []EBPFAgentAlert `json:"disableAlerts"`
 }
--- a/apis/flowcollector/v1beta1/flowcollector_types.go
+++ b/apis/flowcollector/v1beta1/flowcollector_types.go
@@ -157,9 +157,9 @@ const (
 	FlowRTT     AgentFeature = "FlowRTT"
 )
 
-// Name of a ebpf agent alert.
+// Name of an eBPF agent alert.
 // Possible values are:<br>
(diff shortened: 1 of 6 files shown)


jotak:
hey @msherif1234 : I think https://github.com/netobserv/network-observability-operator/pull/632 missed some potential causes for drops. In agent, the drops metric is used in several scenarios:
- https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L541-L545
- https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/flow/limiter.go#L33

On the first one, I see another problem: the alert would be triggered if there are filtered-out flows with the new filter feature. I think that filter feature should use a different metric, wdyt?

msherif1234:
> hey @msherif1234 : I think #632 missed some potential causes for drops. In agent, the drops metric is used in several scenarios:
> 
> * https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L541-L545
> * https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/flow/limiter.go#L33
> 
> On the first one, I see another problem: the alert would be triggered if there are filtered-out flows with the new filter feature. I think that filter feature should use a different metric, wdyt?

Alert uses https://github.com/netobserv/network-observability-operator/blob/main/controllers/ebpf/agent-metrics.go#L130 unlike filter metrics uses `filtered_flows_total` to query and this alert is specific to the hashmap full drop reason as that what customer case was they were sending many flows and not seeing the right number back

msherif1234:
I am fine with the apis edit but as I said above there is no issue for filter metrics, regarding the limiter drops I don't recall adding this not sure what real cases can lead to it

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -157,9 +157,9 @@ const (
 	FlowRTT     AgentFeature = "FlowRTT"
 )
 
-// Name of a ebpf agent alert.
+// Name of an eBPF agent alert.
 // Possible values are:<br>
-// - `NetObservDroppedFlows`, which is triggered when eBPF agent hashmap table is full.<br>
+// `NetObservDroppedFlows`, which is triggered when the eBPF agent is dropping flows, such as when the BPF hashmap is full.<br>
```
msherif1234, line 162: can we state the other possible reason for drop here i.e limiter capacity exceeded and possible recovery config if any ? from what I see I dn't think there is anything can be done to avoid limiter more of internal go limits ?

jotak:
> Alert uses https://github.com/netobserv/network-observability-operator/blob/main/controllers/ebpf/agent-metrics.go#L130 unlike filter metrics uses `filtered_flows_total` to query and this alert is specific to the hashmap full drop reason as that what customer case was they were sending many flows and not seeing the right number back

Oh yes good point I didn't notice

apis/flowcollector/v1beta1/flowcollector_types.go
jotak: done

jotak: the capacity limiter is not related to go limits (if you're talking about the GOMEMLIMIT), it's more some sort of backpressure management, cf the warning text that is logged when it's triggered: https://github.com/netobserv/netobserv-ebpf-agent/blob/bf91cbef0008f6d6bc8b1c748729c18ec6b14d35/pkg/flow/limiter.go#L50-L53

msherif1234: Cool Thanks!!

msherif1234: cc'd @skrthomas this might need to be doc under the ebpf agent alerting WDYT?

jotak:
thanks @msherif1234 
/approve

PR #650, opened 2024-05-21
NETOBSERV-1648: Make high cardinality check more visible

jotak:
## Description

- Add Status/Conditions to FlowMetrics CRD
- Add a Ready status
- Add a Cardinality status
- Add both of them as printColumns

## Dependencies

n/a

## Checklist


memodi:
thanks @jotak , this works as expected:

image

```
$ oc get flowmetrics
NAME           METRIC NAME           STATUS   CARDINALITY
addr-metrics   service_addr_total    Ready    avoid
port-metrics   service_ports_total   Ready    careful
```

/label qe-approved

controllers/flp/fmstatus/flowmetrics_status.go
```diff
@@ -0,0 +1,113 @@
...
+func Reset() {
+	mapStatuses = make(map[types.NamespacedName]*metav1.Condition)
+	mapCards = make(map[types.NamespacedName]*metav1.Condition)
+}
+
+func SetPending(fm *metricslatest.FlowMetric) {
+	nsname := types.NamespacedName{Name: fm.Name, Namespace: fm.Namespace}
+	mapStatuses[nsname] = &metav1.Condition{
+		Type:    "Ready",
+		Reason:  "Ready",
```
msherif1234, line 30: "Pending" ?

controllers/flp/fmstatus/flowmetrics_status.go
```diff
@@ -0,0 +1,113 @@
...
+		if cardinalityCond != nil {
+			meta.SetStatusCondition(&fm.Status.Conditions, *cardinalityCond)
+		}
+		return c.Status().Update(ctx, &fm)
+	})
+
+	if err != nil {
+		log.Error(err, "failed to update FlowMetrics status")
+	}
+}
```
msherif1234, line 113: it might be a good idea adding UT for this while u are here

pkg/manager/manager.go
```diff
@@ -15,24 +15,6 @@ import (
...
-//+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowcollectors/finalizers,verbs=update
-//+kubebuilder:rbac:groups=flows.netobserv.io,resources=flowmetrics,verbs=get;list;watchh;create;update;patch;delete
-//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,resourceNames=hostnetwork,verbs=use
-//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;create;update;watch
-//+kubebuilder:rbac:groups=apiregistration.k8s.io,resources=apiservices,verbs=list;get;watch
-//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;prometheusrules,verbs=get;create;delete;update;patch;list;watch
-//+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions;networks,verbs=get;list;watch
-//+kubebuilder:rbac:groups=loki.grafana.com,resources=network,resourceNames=logs,verbs=get;create
-//+kubebuilder:rbac:urls="/metrics",verbs=get
-
```
msherif1234, line 35: I don't think we should change what the sdk operator generated template, now if we added new API or webhook we will have to move the newly generated ones to the permissions.go while this is none functional change but I prefer sticking with SDK generated fmt

config/rbac/role.yaml
```diff
@@ -132,18 +132,11 @@ rules:
   - flows.netobserv.io
   resources:
   - flowcollectors/status
+  - flowmetrics/status
   verbs:
   - get
   - patch
   - update
```
msherif1234, line 139: u dropped `watch` verb that was used with fm is that no longer needed ?

pkg/manager/manager.go
jotak: this is because [here](https://github.com/netobserv/network-observability-operator/pull/622/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R261) you changed the controller-gen call to check only the `./apis` path and not the whole repo, so that broke the rbac generation; we need either to move that to `apis` or to revert that change on controller-gen path, did you have a specific reason to change the path?

config/rbac/role.yaml
jotak: no `flowmetrics` still has `watch` look at line 124, it's now having same perms as `flowcollectors`

controllers/flp/fmstatus/flowmetrics_status.go
jotak: done

controllers/flp/fmstatus/flowmetrics_status.go
jotak: done

pkg/manager/manager.go
msherif1234: no we can revert the controller-gen path changes I was debugging something and left that change around

jotak: ok, done

PR #664, opened 2024-05-29
NETOBSERV-1642: Ovs monitoring feature

msherif1234:
## Description

changes to support ovs monitoring config and console bits

FC config to enable ovs monitoring fearure
```yaml
agent:
    type: eBPF
    ebpf:   
      privileged: true
      features:
       - "OvsMonitor"

```
## Dependencies

https://github.com/netobserv/netobserv-ebpf-agent/pull/286

## Checklist

### DIFF
--- a/apis/flowcollector/v1beta1/flowcollector_types.go
+++ b/apis/flowcollector/v1beta1/flowcollector_types.go
@@ -150,14 +150,16 @@ type FlowCollectorIPFIX struct {
 // Agent feature, can be one of:<br>
 // - `PacketDrop`, to track packet drops.<br>
 // - `DNSTracking`, to track specific information on DNS traffic.<br>
-// - `FlowRTT`, to track TCP latency. [Unsupported (*)].<br>
-// +kubebuilder:validation:Enum:="PacketDrop";"DNSTracking";"FlowRTT"
+// - `FlowRTT`, to track TCP latency [Unsupported (*)].<br>
+// - `NetworkEvents`, to track Network events.<br>
+// +kubebuilder:validation:Enum:="PacketDrop";"DNSTracking";"FlowRTT";"NetworkEvents"
 type AgentFeature string
 
 const (
-	PacketDrop  AgentFeature = "PacketDrop"
-	DNSTracking AgentFeature = "DNSTracking"
-	FlowRTT     AgentFeature = "FlowRTT"
+	PacketDrop    AgentFeature = "PacketDrop"
+	DNSTracking   AgentFeature = "DNSTracking"
+	FlowRTT       AgentFeature = "FlowRTT"
+	NetworkEvents AgentFeature = "NetworkEvents"
 )
 
 // Name of an eBPF agent alert.
@@ -325,6 +327,8 @@ type FlowCollectorEBPF struct {
 	// If the `spec.agent.ebpf.privileged` parameter is not set, an error is reported.<br>
 	// - `DNSTracking`: enable the DNS tracking feature.<br>
 	// - `FlowRTT`: enable flow latency (sRTT) extraction in the eBPF agent from TCP traffic.<br>
+	// - `NetworkEvents`: enable the Network events monitoring feature. This feature requires mounting
+	// the kernel debug filesystem, so the eBPF pod has to run as privileged.
 	// +optional
 	Features []AgentFeature `json:"features,omitempty"`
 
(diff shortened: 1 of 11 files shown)


controllers/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -1087,6 +1087,22 @@ fields:
     type: number
     description: TCP Smoothed Round Trip Time (SRTT), in nanoseconds
     cardinalityWarn: avoid
+  - name: OvsMonitorMD[0]
+    type: string
+    description: OVS monitor metadata for the 1st event
+    cardinalityWarn: avoid
```
jpinsonneau, line 1093: You should also add these into `columns` and `filters` such as:

```
columns:
  - id: OvsMonitorMD0
    name: 1st OVS metadata
    tooltip: OVS monitor metadata for the 1st event
    field: OvsMonitorMD[0]
    filter: ovs_metadata_0
    default: true
    width: 5
    feature: OvsMonitor
...
filters:
  - id: ovs_metadata_0
    name: OVS monitor metadata for the 1st event
    component: text
    hint: Specify a single metadata.
...
```

jpinsonneau: if we want to make somether smarter that concatenate the events, it will require some code changes in the plugin

controllers/constants/constants.go
```diff
@@ -42,9 +42,10 @@ const (
 	MonitoringServiceAccount = "prometheus-k8s"
 
 	// Loki roles
-	LokiCRWriter  = "netobserv-writer"
-	LokiCRBWriter = "netobserv-writer-flp"
-	LokiCRReader  = "netobserv-reader"
+	LokiCRWriter        = "netobserv-writer"
+	LokiCRBWriter       = "netobserv-writer-flp"
+	LokiCRReader        = "netobserv-reader"
+	MaxOvsMonitorEvents = 4
```
jpinsonneau, line 48: I'm not sure to get the impact of this yet. What would be the advantage to get more / less events metadatas ?

msherif1234: for the same flow we can have multiple events md if we wanted to trace different feature for example we wanted to trace ACL, NAT, QoS for example we will see array populated with 4 metadata strings each for one of the above features, while if we trace ACL only for example then only one event md will be populated

controllers/consoleplugin/config/static-frontend-config.yaml
msherif1234: I believe there is still a need for plugin work anyway to handle the new feature if that is accurate optimizing the above can be part of that too ? for now I will see how far this takes me

controllers/constants/constants.go
jpinsonneau: ok makes sense. So are we going to expose which features to trace in the FlowCollector CR ?

If so we could adapt the console plugin config dynamically and bind each field with its proper name / type etc

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -150,14 +150,16 @@ type FlowCollectorIPFIX struct {
 // Agent feature, can be one of:<br>
 // - `PacketDrop`, to track packet drops.<br>
 // - `DNSTracking`, to track specific information on DNS traffic.<br>
-// - `FlowRTT`, to track TCP latency. [Unsupported (*)].<br>
-// +kubebuilder:validation:Enum:="PacketDrop";"DNSTracking";"FlowRTT"
+// - `FlowRTT`, to track TCP latency [Unsupported (*)].<br>
+// - `NetworkEventsMonitor`, to track Network events.<br>
```
jotak, line 154: just wondering if naming it just `NetworkEvents` would be sufficient and easier to remember, no ?

jotak: Also, maybe add more details right here because it's not super obvious what are network events. like ".. such as network policy verdicts" ?

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -325,6 +327,7 @@ type FlowCollectorEBPF struct {
 	// If the `spec.agent.ebpf.privileged` parameter is not set, an error is reported.<br>
 	// - `DNSTracking`: enable the DNS tracking feature.<br>
 	// - `FlowRTT`: enable flow latency (sRTT) extraction in the eBPF agent from TCP traffic.<br>
+	// - `NetworkEventsMonitor`: enable the Network events monitoring feature.<br>
```
jotak, line 330: like in PacketDrop, we should mention here if Privileged is required
Also, maybe add more details right here because it's not super obvious what are network events. like ".. such as network policy verdicts" ?

PR #669, opened 2024-06-05
NETOBSERV-1647: Do not load loki cert when loki disabled

jotak:
## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/consoleplugin/consoleplugin_objects.go
+++ b/controllers/consoleplugin/consoleplugin_objects.go
@@ -209,15 +209,18 @@ func (b *builder) podTemplate(cmDigest string) *corev1.PodTemplateSpec {
 	}
 
 	// ensure volumes are up to date
+	// TODO/FIXME: why? The same is done in `getLokiConfig`, why doing it in two places? If there's a reason it should be commented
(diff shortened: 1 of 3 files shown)


controllers/consoleplugin/consoleplugin_objects.go
```diff
@@ -208,18 +208,6 @@ func (b *builder) podTemplate(cmDigest string) *corev1.PodTemplateSpec {
 		})
 	}
 
-	// ensure volumes are up to date
```
jotak, line 211: I think this code is not necessary but I don't want to introduce a regression so close to the release, so I'm only removing it from `main`, [not in the backport PR](https://github.com/netobserv/network-observability-operator/pull/670/files), to play safe.
Note that we discussed this already a while ago: https://github.com/netobserv/network-observability-operator/pull/477/files?diff=split&w=1#r1384571479 - I don't see why this code is duplicated, and if there's a reason I can bring it back but it should be better documented and ideally covered by a test that show-case it.

memodi:
/label qe-approved 

(based on testing for backport PR)

PR #674, opened 2024-06-10
NETOBSERV-1688: relax ebpf drop alert

jotak:
reformulate some of the text for ebpf drop alert, and relax threshold a little bit.

## Dependencies

n/a

## Checklist


jotak:
@msherif1234 @memodi please let me know also if the new doc & alert text is better like that. The changes I'm making are to convey:
- that the root cause might be the map being "busy" (EBUSY errno) , which in my understanding *can* stand for being full, but not necessarily
- that this could be "just" packets missed but not necessarily missing flows (ie. it's just the bytes/packets counter in an existing flow that won't be updated in case of miss, but the flow will still appear)
- also I prefer the word "missing" rather than "dropping" because I don't want user could think we could actually **drop a packet** ie. interfere with the data

controllers/ebpf/agent-metrics.go
```diff
@@ -124,10 +124,10 @@ func (c *AgentController) agentPrometheusRule(target *flowslatest.FlowCollectorE
 		rules = append(rules, monitoringv1.Rule{
 			Alert: string(flowslatest.AlertDroppedFlows),
 			Annotations: map[string]string{
-				"description": "NetObserv eBPF agent is not able to process new flows. Possible reasons are the BPF hashmap being full, or the capacity limiter being triggered. Both causes can be worked around by increasing cacheMaxFlows value in Flowcollector resource.",
-				"summary":     "NetObserv eBPF is not able to process any new flows",
+				"description": "NetObserv eBPF agent is missing packets or flows. The metric netobserv_agent_dropped_flows_total provides more information on the cause. Possible reasons are the BPF hashmap being busy or full, or the capacity limiter being triggered. This may be worked around by increasing cacheMaxFlows value in Flowcollector resource.",
+				"summary":     "NetObserv eBPF agent is missing packets or flows",
 			},
-			Expr: intstr.FromString("sum(rate(netobserv_agent_dropped_flows_total[1m])) > 0"),
+			Expr: intstr.FromString("sum(rate(netobserv_agent_dropped_flows_total[1m])) > 100"),
```
msherif1234, line 130: just put this 100 limit in the constants

jotak: done

controllers/ebpf/agent_controller.go
```diff
@@ -79,14 +79,15 @@ const (
...
-	bpfNetNSMountPath  = "/var/run/netns"
+	exportKafka                = "kafka"
+	exportGRPC                 = "grpc"
+	kafkaCerts                 = "kafka-certs"
+	averageMessageSize         = 100
+	bpfTraceMountName          = "bpf-kernel-debug"
+	bpfTraceMountPath          = "/sys/kernel/debug"
+	bpfNetNSMountName          = "var-run-netns"
+	bpfNetNSMountPath          = "/var/run/netns"
+	droppedFlowsAlertThreshold = 100
```
memodi, line 90: the threshold of 100 , is it based on some testing or an arbitrary value that we think is good (not too high or not too low)?

memodi (changes requested):

controllers/ebpf/agent_metrics.go
```diff
@@ -124,10 +124,10 @@ func (c *AgentController) agentPrometheusRule(target *flowslatest.FlowCollectorE
 		rules = append(rules, monitoringv1.Rule{
 			Alert: string(flowslatest.AlertDroppedFlows),
 			Annotations: map[string]string{
-				"description": "NetObserv eBPF agent is not able to process new flows. Possible reasons are the BPF hashmap being full, or the capacity limiter being triggered. Both causes can be worked around by increasing cacheMaxFlows value in Flowcollector resource.",
-				"summary":     "NetObserv eBPF is not able to process any new flows",
+				"description": "NetObserv eBPF agent is missing packets or flows. The metric netobserv_agent_dropped_flows_total provides more information on the cause. Possible reasons are the BPF hashmap being busy or full, or the capacity limiter being triggered. This may be worked around by increasing cacheMaxFlows value in Flowcollector resource.",
```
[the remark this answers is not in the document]
jotak: the capacity limiter is something different than the hashmap capacity , it's something that comes later in the process (just before flows are about to be sent over the network to kafka or FLP)

controllers/ebpf/agent_controller.go
msherif1234: random number atm

controllers/ebpf/agent_metrics.go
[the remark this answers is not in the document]
msherif1234: I am fine with new description

controllers/ebpf/agent-metrics-test.go
```diff
@@ -1,56 +0,0 @@
-package ebpf
```
msherif1234, line 1: any reason for deleting this unit-test ?

controllers/ebpf/agent_controller.go
jotak: That's arbitrary, but I don't see how to not make it arbitrary, unless by making it a configurable threshold. People who would like to set another limit can still copy/paste the rule to a new one, and turn off this alert in netobserv. The rule is still useful in that sense.

Still, I did test it on my cluster.
With my "medium" hey-ho test intensity (which actually is quite high already), I'm between 10 and 15 drops/sec. With my "high" intensity I'm around 25-30.

Bigger clusters would likely see more drops/sec under stress.

controllers/ebpf/agent-metrics-test.go
jotak: Yes I explained it in the commit message :)
[dc257fb](https://github.com/netobserv/network-observability-operator/pull/674/commits/dc257fb3931d98487f120e44d1e83f916ef660cc)

controllers/ebpf/agent_metrics.go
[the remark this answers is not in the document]
memodi: I see, so increasing the cacheMaxFlows value will help when flows/packets are dropped due to hashmap being full but it wouldn't really help with if the capacity limit is triggered, correct?

controllers/ebpf/agent_controller.go
memodi: got it, thanks!

controllers/ebpf/agent_metrics.go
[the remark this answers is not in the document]
jotak: Actually, no. The capacity limiter also obeys to cacheMaxFlows setting, so it will also be a solution in that case.
The case where cacheMaxFlows won't be a solution is the very one we've seen here: when packets are missed due to hashmap being "busy", but not full. Unfortunately we haven't found so far a way to distinguish cases between this sort of "busy" error versus map being full, that would be easily visible to users. One thing they might do is correlating this error with other metrics that provide the current size usage of the map, and see whether or not it's close to its max capacity.

PR #676, opened 2024-06-11
NETOBSERV-1649: add drop cause & state labels to metrics

jotak:
## Description

Add more metrics info to make prometheus datasource suiting for replacing loki datasource in overview page

- Add drop cause & TCP state labels to the drop metrics
- When only prometheus is enabled, turn on "workload_ingress_packets_total" by default

## Dependencies

n/a

## Checklist

  * [x] If so, fill in "Release Note Text" in the JIRA.

### DIFF
--- a/pkg/metrics/predefined_metrics.go
+++ b/pkg/metrics/predefined_metrics.go
@@ -46,6 +46,7 @@ var (
 	DefaultIncludeListLokiDisabled = []string{
 		"node_ingress_bytes_total",
 		"workload_ingress_bytes_total",
+		"workload_ingress_packets_total",
 		"workload_flows_total",
 		"workload_drop_bytes_total",
 		"workload_drop_packets_total",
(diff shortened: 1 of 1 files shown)


jotak:
fyi this makes the plugin overview page work, without anything more to do on plugin side:

image

OlivierCazade (approved):

memodi (commented):
@jotak I could see drops reasons, however some filter options in UI needs to be disabled for packetdrop such as for state and drops since they're not part of the labels when only prom. is used:

Screenshot 2024-06-20 at 4 50 37 PM

jotak:
> @jotak I could see drops reasons, however some filter options in UI needs to be disabled for packetdrop such as for state and drops since they're not part of the labels when only prom. is used:

@memodi : actually, these filters are available on purpose, because they can be used in _some_ queries. But it doesn't always work, because what matters is the combination of labels. For instance, you can filter on Node individually, you can filter on Namespace individually, but you cannot filter on Node _and_ Namespace. It's the same reason why, in the topology view, if you have a filter on namespaces and then switch to the Node aggregation, you will get an error telling you to enable Loki blah blah. This is similar with your example with drops here, the filter alone is supported, but not in combination with other filters.

I can suggest 2 things:
- I opened a new JIRA to improve the UX: https://issues.redhat.com/browse/NETOBSERV-1719 . One important thing I want to address there, is the ability to still display some data when parts of the queries fail, but not all. Especially in the overview, many queries are run depending on the selected panels to display; some of them may work but not all. Errors should be displayed per-panel, not globally.
- I also opened a PR to fix some of the most annoying aspects: https://github.com/netobserv/network-observability-console-plugin/pull/549 . It won't fix everything, but it covers the error that you mentioned (using a drop cause filter)

memodi:
this looks good, tested along with PR: netobserv/network-observability-console-plugin/pull/549. Thanks @jotak 

/label qe-approved

PR #689, opened 2024-06-25
force conflicts on deploy

jotak:
with `kubectl apply --server-side`, sometimes we get conflict errors, such as after using the make target `set-xxx-image` which amends the deployment env variables. Using `--force-conflicts` allows to ignore such conflict, and makes the new deployment prevail.


jotak:
thanks @msherif1234 
/approve

PR #728, opened 2024-08-01
 NETOBSERV-1786: Add support to build file based catalog

OlivierCazade:
## Description

This catalog will replace the catalog we currently use, for now, we have Makefile target for both.

## Dependencies

n/a

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,10 @@ REPO ?= quay.io/$(IMAGE_ORG)
 
 # Component versions to use in bundle / release (do not use $VERSION for that)
 PREVIOUS_VERSION ?= v1.6.0-community
+
 BUNDLE_VERSION ?= 1.6.1-community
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -429,6 +437,14 @@ bundle-build: ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	$(OCI_BIN) push ${BUNDLE_IMAGE};
 
+.PHONY: file-based-catalog-build
```
msherif1234, line 440: just curious why did  u added those two new targets since konflux pipeline will use the containerfile directly ?

OlivierCazade: We currently have target to build a regular catalog, but if I understand correctly at some point this catalog will be replaced by file based catalog.

I wanted to introduce the new targets without replacing the old ones for now.

PR #732, opened 2024-08-07
NETOBSERV-1790 Mac enrichment feature gate

jpinsonneau:
## Description

Put Mac address enrichment behind feature gate

You will need to processor env to:
`EnrichFromMac: true`

## Dependencies

https://github.com/netobserv/flowlogs-pipeline/pull/674

## Checklist


jpinsonneau:
@jotak it seems prometheus lib behavior changed here
https://github.com/netobserv/network-observability-operator/actions/runs/10285077714/job/28462500419?pr=732

Do you have an idea on how to bypass that ?

jpinsonneau:
updated prometheus common dependency
https://github.com/netobserv/network-observability-operator/pull/732/commits/2f7c1f7602a6845197a8d70d3e8632ca88f39759

go.mod
```diff
@@ -83,4 +82,6 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
-replace github.com/prometheus/common v0.48.0 => github.com/netobserv/prometheus-common v0.48.0-netobserv
+replace github.com/prometheus/common v0.48.0 => github.com/netobserv/prometheus-common v0.55.0-netobserv
```
jotak, line 85: curious that it works, it should be:

```suggestion
replace github.com/prometheus/common v0.55.0 => github.com/netobserv/prometheus-common v0.55.0-netobserv
```

controllers/flp/flp_pipeline_builder.go
```diff
@@ -66,21 +66,29 @@ func (b *PipelineBuilder) AddProcessorStages() error {
 	allLabels = append(allLabels, b.detectedSubnets...)
 	flpLabels := subnetLabelsToFLP(allLabels)
 
+	var srcMacInput, dstMacInput string
```
jotak, line 69: I think we will have an issue with flows on pods identified out of MAC: the app/infra layer label won't be set, because it uses IP address only to search in index and figure out the namespace. cf line 100 this file.
I see 2 options:
- Either we provide MacInputs as well via `K8sInfraRule`
- Or we change the way to get namespace, and instead of doing an informer lookup in FLP like it's currently done [here](https://github.com/netobserv/flowlogs-pipeline/blob/f5d2cefe616774d46ac94eba94a62217ee536b46/pkg/pipeline/transform/kubernetes/enrich.go#L134), we change that to look at the namespace field already enriched in the flow GenericMap.

I tend to prefer option 2, but it means the enriching stage becomes a pre-requisite for the Infra-layer-enrich stage.

@jpinsonneau @OlivierCazade wdyt?

jotak: I updated https://github.com/netobserv/flowlogs-pipeline/pull/674 with that 2d option.
It's a breaking change so I need to update this PR as well

memodi:
@jpinsonneau when EnrichFromMac is set to true, will enrich using mac address for all the pod traffic or just for secondary interfaces?

jpinsonneau:
> @jpinsonneau when EnrichFromMac is set to true, will enrich using mac address for all kinds of network traffic or just for secondary interfaces?

We will check for MAC matching first on any flow. This is not stricly tied to secondary interfaces as we grab the MAC addresses from the `k8s.v1.cni.cncf.io/network-status` annotation which could be filled by any network implementation; not only multus.
However, if no match is found using MAC, we fallback on IP matching as previously. 
Note that the related FLP PR also grab extra IPs from the same annotation:
https://github.com/netobserv/flowlogs-pipeline/pull/674/files#diff-0b5bec06046ff7cd7790a626db2b43a0361278be4265b8f19833f402b3968678R41-R59

memodi:
> > @jpinsonneau when EnrichFromMac is set to true, will enrich using mac address for all kinds of network traffic or just for secondary interfaces?
> 
> We will check for MAC matching first on any flow. This is not stricly tied to secondary interfaces as we grab the MAC addresses from the `k8s.v1.cni.cncf.io/network-status` annotation which could be filled by any network implementation; not only multus. However, if no match is found using MAC, we fallback on IP matching as previously. Note that the related FLP PR also grab extra IPs from the same annotation: https://github.com/netobserv/flowlogs-pipeline/pull/674/files#diff-0b5bec06046ff7cd7790a626db2b43a0361278be4265b8f19833f402b3968678R41-R59

Sounds good, thanks @jpinsonneau !

PR #756, opened 2024-09-06
1.6 hotfix consolealpha1

jotak:


jotak:
@memodi @Amoghrd I haven't had the opportunity to test that yet, it's not guaranteed to work .. We must test on a 4.17 cluster.

The goal is that, when console v1alpha1 isn't available, the operator still works and deploys agents+FLP without errors (it won't try to deploy the console plugin)
Then, users will have to do some manual steps if they want the console plugin back.

memodi:
@jotak couldn't we use console/v1 api instead for 1.6.2? it seems to available for 4.12+ https://docs.openshift.com/container-platform/4.12/rest_api/console_apis/consoleplugin-console-openshift-io-v1.html

jotak:
closing, replaced by https://github.com/netobserv/network-observability-operator/pull/762

PR #763, opened 2024-09-09
NETOBSERV-1811 set namespace ownership

jpinsonneau:
## Description

Set namespace ownership using a new `netobserv-managed: true` label so users can bypass reconcile loop if they want to customize labels by setting it to false.

The `ownerReference` is not set to namespace object to avoid unexpected deletions.

## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/monitoring/monitoring_controller.go
+++ b/controllers/monitoring/monitoring_controller.go
@@ -102,7 +102,7 @@ func (r *Reconciler) reconcile(ctx context.Context, clh *helper.Client, desired 
 		if err != nil {
 			return err
 		}
-	} else if !helper.IsSubSet(nsExist.ObjectMeta.Labels, desiredNs.ObjectMeta.Labels) {
+	} else if !helper.SkipOwnership(nsExist) && !helper.IsSubSet(nsExist.ObjectMeta.Labels, desiredNs.ObjectMeta.Labels) {
 		err = r.Update(ctx, desiredNs)
 		if err != nil {
 			return err
--- a/pkg/helper/client_helper.go
+++ b/pkg/helper/client_helper.go
@@ -47,6 +47,7 @@ func (c *Client) CreateOwned(ctx context.Context, obj client.Object) error {
 		log.Error(err, "Failed to set controller reference")
 		return err
 	}
+	AddOwnedLabel(obj)
 	kind := reflect.TypeOf(obj).String()
 	log.Info("CREATING a new "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
 	err = c.Create(ctx, obj)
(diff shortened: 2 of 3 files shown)


jpinsonneau:
> @jpinsonneau - I tested this and it works. The custom labels and modifications does persist after this change.
> 
> Do you think it would be useful to add regression test for this bug?

I don't feel it's mandatory but it's good to have especially since we are going to start documenting aroud this behavior to solve the customer case linked in the ticket.

FYI this apply on any yaml owned by the controller, not only namespace.

jotak:
Hold on, what happens with this change if we delete the FlowCollector => the namespace will be deleted as well, right? That could be too brutal. E.g. if a user installed Loki in the same namespace, they'd loose it, including its PVC etc.

jotak:
An alternative to setting ownership could be [here](https://github.com/netobserv/network-observability-operator/blob/main/controllers/monitoring/monitoring_controller.go#L105-L106) to just extract the extra labels and reinject them in the desired namespace object ? wdyt?

jpinsonneau:
> An alternative to setting ownership could be [here](https://github.com/netobserv/network-observability-operator/blob/main/controllers/monitoring/monitoring_controller.go#L105-L106) to just extract the extra labels and reinject them in the desired namespace object ? wdyt?

The console usually asks you if you want to cascade or not the deletion but I don't think it's the case using kubectl / oc commands.
Can't we implement a smart finalizer here ? :thinking: 
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#ownership-and-finalizers
https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/

jotak:
> > An alternative to setting ownership could be [here](https://github.com/netobserv/network-observability-operator/blob/main/controllers/monitoring/monitoring_controller.go#L105-L106) to just extract the extra labels and reinject them in the desired namespace object ? wdyt?
> 
> So if you simply invert the `DownstreamDeployment` flag it will not reconcile anymore ?

Not sure to get what you mean.. If we change `openshift.io/cluster-monitoring` it would still be reconciled and reverted to what we want, I don't mean any regression about that. Only the labels that we don't set ourselves would be reinjected. Something like that:

```go
		for k, v := range nsExist.ObjectMeta.Labels {
			if _, exists := desiredNs.ObjectMeta.Labels[k]; !exists {
				desiredNs.ObjectMeta.Labels[k] = v
			}
		}
```

> In that case maybe just add a `netobserv-managed` label in the namespace and check of its value before reconciling ? Why not doing that for any object we manage ?

I'm not against that but I think it's a different thing. Here reading @stleerh 's ticket, I don't think he wants to make the namespace unmanaged. It's more a matter of having it co-managed, we want to manage the bits that we need but without deleting stuff written by others. In fact, we already do that for some resources, like Services: on update, we don't recreate them totally from scratch, we keep some existing bits. The 5 lines of code above would just do that.

jpinsonneau:
@jotak following our discussion I reworked the code to implement a `netobserv-managed` label.
Reconciliation is ignored if it's explicitly set to `false`.
Let me know if that works better for you !

@memodi sorry we changed the way it's implemented here. Let's validate the approach before doing QE validation.

jotak (approved):

jotak:
merging for branch cut, to be verified post-merge

PR #768, opened 2024-09-11
NETOBSERV-1856: new role for metrics reading permissions

jotak:
## Description

A new role is created for reading metrics. This role is separate from netobserv-reader role, because it is meant to be namespace-scoped and not cluster-wide (whereas for Loki multi-tenancy, netobserv-reader has to be bound cluster-wide).

To use this role, just write: 

```bash
oc adm policy add-role-to-user netobserv-metrics-reader test -n my-namespace
```

This will grant user "test" access to the netflow metrics in the "my-namespace" namespace, which can be used both in the developer view and in the adminitrator view.

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/resources/static_resources.go
+++ b/pkg/resources/static_resources.go
@@ -7,9 +7,9 @@ import (
 	"github.com/netobserv/network-observability-operator/controllers/constants"
 )
 
-var NetObservWriterCR = rbacv1.ClusterRole{
+var LokiWriterCR = rbacv1.ClusterRole{
 	ObjectMeta: metav1.ObjectMeta{
-		Name: constants.CRWriter,
+		Name: constants.LokiCRWriter,
 	},
 	Rules: []rbacv1.PolicyRule{{
 		APIGroups:     []string{"loki.grafana.com"},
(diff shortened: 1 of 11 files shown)


pkg/resources/static_resources.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+var PromReaderCR = rbacv1.ClusterRole{
+	ObjectMeta: metav1.ObjectMeta{
+		Name: constants.PromCRReader,
+	},
+	Rules: []rbacv1.PolicyRule{{
+		APIGroups: []string{"metrics.k8s.io"},
+		Resources: []string{"pods"},
+		Verbs:     []string{"create"},
```
jpinsonneau, line 41: Should we put both `get` & `create` here to handle the future scenario where kube rbac will be fixed ?

jpinsonneau (approved):
LGTM, just a small comment

skrthomas:
Just commenting so I can keep an eye on updates here.

jotak:
@Amoghrd I tested on my side and it worked for me, although I observed some temporary issue with sometimes 403 being returned or sometimes it worked upon successive refreshes, before eventually having all requests working within a minute. I guess prometheus proxy has some cached permissions which expire after ~ a minute or so.
Also, did you make sure to use cluster monitoring (make set-release-kind-downstream) and not user workload mon? We know it doesn't work with user workload.

Amoghrd:
oh maybe I missed the downstream part. will do that and retry. thanks!

PR #769, opened 2024-09-11
Fix podman builds

jotak:
@OlivierCazade do you know what we need to check to make sure this won't break anything?


jotak:
Well I get this warning 
```
[Warning] one or more build args were not consumed: [BUILDPLATFORM]
```
So that's probably wrong...

OlivierCazade (approved):
LGTM

PR #799, opened 2024-10-02
NETOBSERV-1893: fix netobserv-mertics-reader role creation

jotak:
Prevent undesired mutation of the role by copying it before use.

## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/flp/flp_controller.go
+++ b/controllers/flp/flp_controller.go
@@ -269,8 +269,8 @@ func reconcileDataAccessRoles(ctx context.Context, r *reconcilers.Common, b *bui
 			}
 		}
 	}
-	// Install netobserv-metrics-reader role; copy to avoid any undesired mutation
-	cr := resources.PromReaderCR
+	// Install netobserv-metrics-reader role
+	cr := resources.PromReaderCR()
 	return r.ReconcileClusterRole(ctx, &cr)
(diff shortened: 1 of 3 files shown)


controllers/flp/flp_controller.go
```diff
@@ -269,8 +269,9 @@ func reconcileDataAccessRoles(ctx context.Context, r *reconcilers.Common, b *bui
 			}
 		}
 	}
-	// Install netobserv-metrics-reader role
-	return r.ReconcileClusterRole(ctx, &resources.PromReaderCR)
+	// Install netobserv-metrics-reader role; copy to avoid any undesired mutation
+	cr := resources.PromReaderCR
```
msherif1234, line 273: do u need deepCopy here ?

OlivierCazade (approved):

controllers/flp/flp_controller.go
jotak: it's not strictly needed today but it's right that it would be safer in case we add more mutations later...
or alternatively we can have `PromReaderCR` being a function, rather than a global var

jotak: @msherif1234 changed to use a function

Amoghrd:
LGTM

PR #827, opened 2024-10-28
Improve doc for ebpf filters

jotak:


apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -221,59 +221,60 @@ type EBPFFlowFilter struct {
 	// Examples: `10.10.10.0/24` or `100:100:100:100::/64`
 	CIDR string `json:"cidr,omitempty"`
 
-	// `action` defines the action to perform on the flows that match the filter.
+	// `action` defines the action to perform on the flows that match the filter. The available options are `Accept` and `Reject`.
 	// +kubebuilder:validation:Enum:="Accept";"Reject"
 	Action string `json:"action,omitempty"`
 
-	// `protocol` defines the protocol to filter flows by.
+	// `protocol` optionally defines a protocol to filter flows by. The available options are `TCP`, `UDP`, `ICMP`, `ICMPv6` and `SCTP`.
```
skrthomas, line 228:
```suggestion
	// `protocol` optionally defines a protocol to filter flows by. The available options are `TCP`, `UDP`, `ICMP`, `ICMPv6`, and `SCTP`.
```

apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -221,59 +221,60 @@ type EBPFFlowFilter struct {
...
 
-	// `direction` defines the direction to filter flows by.
+	// `direction` optionally defines a direction to filter flows by. The available options are `Ingress` and `Egress`.
 	// +kubebuilder:validation:Enum:="Ingress";"Egress"
 	// +optional
 	Direction string `json:"direction,omitempty"`
 
-	// `tcpFlags` defines the TCP flags to filter flows by.
+	// `tcpFlags` optionally defines TCP flags to filter flows by.
+	// In addition to the standard flags (RFC-9293), you can also filter by one of the three following combinations: `SYN-ACK`, `FIN-ACK` and `RST-ACK`.
```
skrthomas, line 239:
```suggestion
	// In addition to the standard flags (RFC-9293), you can also filter by one of the three following combinations: `SYN-ACK`, `FIN-ACK` and, `RST-ACK`.
```

apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -1334,7 +1335,7 @@ const (
 
 // `FlowCollectorExporter` defines an additional exporter to send enriched flows to.
 type FlowCollectorExporter struct {
-	// `type` selects the type of exporters. The available options are `Kafka` and `IPFIX`.
+	// `type` selects the type of exporters. The available options are `Kafka`, `IPFIX` and `OpenTelemetry`.
```
skrthomas, line 1338:
```suggestion
	// `type` selects the type of exporters. The available options are `Kafka`, `IPFIX` and, `OpenTelemetry`.
```

skrthomas (commented):
Peer review feedback. One additional comment was on lines 2541-2542 "We recommend you either enable it, or create your own network policy for Network Observability."

IBMSG advises avoiding claims or recommendations, instead, they say to describe more about why the action is beneficial to the extent that a user _should_ do it, in our case. You do have some description here: These network policies better isolate the Network Observability components to prevent undesired connections to them.

Maybe add instead: To increase the security of connections, enable this option or create your own network policy.

jotak:
thanks @skrthomas , comments addressed

apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -221,59 +221,60 @@ type EBPFFlowFilter struct {
 	// Examples: `10.10.10.0/24` or `100:100:100:100::/64`
 	CIDR string `json:"cidr,omitempty"`
 
-	// `action` defines the action to perform on the flows that match the filter.
+	// `action` defines the action to perform on the flows that match the filter. The available options are `Accept` and `Reject`.
```
msherif1234, line 224: agent default to `Accept` I don't think we need to default to Accept in the API but should be say we have a default ?

jotak: I thought it was actually required.. Well maybe I'm for setting an explicit default to Accept then, so it's clearly set in API and not handed over to agent's own default?

jotak: done

docs/flowcollector-flows-netobserv-io-v1beta2.adoc
```diff
@@ -377,7 +377,7 @@ Type::
 
 | `action`
 | `string`
-| `action` defines the action to perform on the flows that match the filter. The available options are `Accept` and `Reject`.
+| `action` defines the action to perform on the flows that match the filter. The available options are `Accept` (default) and `Reject`.
```
skrthomas, line 380: `action` defines the action to perform on the flows that match the filter. The available options are `Accept`, which is the default, and `Reject`.

skrthomas: can we add the default info in parentheses? Sorry for the nit but IBMSG doesn't like parentheses 👼

jotak:
merging since downstream was merged

PR #896, opened 2024-11-25
NETOBSERV-1913 skip <br> when bundle

jpinsonneau:
## Dependencies

n/a

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -402,6 +402,7 @@ bundle: bundle-prepare ## Generate final bundle files.
 	rm -r bundle/metadata
 	cp ./config/csv/bases/netobserv-operator.clusterserviceversion.yaml tmp-csv
 	hack/crd2csvSpecDesc.sh v1beta2
+	find ./config/crd -type f | xargs sed -i  's/<br>//g'
 	$(SED) -e 's/^/    /' config/descriptions/upstream.md > tmp-desc
 	$(KUSTOMIZE) build $(BUNDLE_CONFIG) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -402,6 +402,7 @@ bundle: bundle-prepare ## Generate final bundle files.
 	rm -r bundle/metadata
 	cp ./config/csv/bases/netobserv-operator.clusterserviceversion.yaml tmp-csv
 	hack/crd2csvSpecDesc.sh v1beta2
+	find ./config/crd -type f | xargs sed -i  's/<br>//g'
```
jpinsonneau, line 405: @jotak I feel that would be good enough as it's running only when `bundle` target is called.
Let me know if there is any concern here

jpinsonneau: oh ok! In that case we need to have two crds generated if we want to go that way :upside_down_face: 
- one for the doc
- one for the bundle

We could have two targets in the Makefile or reuse CLEAN_BUILD param but it means the downstream pipeline will have to generate these unless we keep two version of CRDs upstream :thinking: 

WDYT ?

jotak: Like I said in jira is it really worth it to change the build process for that, since we want to move away from the OLM form anyway, and have our own form? I don't think we should spend much time here ....

jpinsonneau: Even if you move away from the creation form, the YAML editor still display those descriptions.
Plus, most of the time, the editing is done outside of forms. That's why I wanted to see if we can find a quick solution here

jpinsonneau:
Closing this as it's not solving the initial issue as is

PR #912, opened 2024-11-26
NETOBSERV-1981: Fix Exporters form view

Amoghrd:
## Description
Fix form view for exporters section

## Dependencies
n/a

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [ ] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labeled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [ ] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [x] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #942, opened 2024-12-09
OLM forms, fix display name generation for mac

jotak:
We need to check if it fixes csv gen on mac.
Initial issue was mentioned here: https://github.com/netobserv/network-observability-operator/pull/912#discussion_r1868183836


jotak:
looks like it still requires bash 4 or more https://stackoverflow.com/questions/12487424/uppercase-first-character-in-a-variable-with-bash

PR #995, opened 2025-01-09
Readme: Warn infobox about community releases

jotak:


jotak:
I'll merge this ... even if we start a move to helm, this message still stands for OLM

PR #1022, opened 2025-01-17
Refresh archicture doc

jotak:
This doc is very old and needs a refresh!

I'd move to mermaid charts, like what we have in the agent, which makes it easier to update.


docs/Architecture.md
```diff
@@ -1,49 +1,93 @@
...
+    C[Console plugin] <-->|fetches| L
+    C <-->|fetches| P
+    O[Operator] -->|manages| A
+    O -->|manages| F
+    O -->|manages| C
+```
+
+## CLI
+
+When using the CLI, the operator is not involved, which means you can use it without installing NetObserv as a whole. It uses a special mode of the eBPF agents that embeds FLP.
```
jotak, line 81: cc @jpinsonneau does it look good to you? cf https://github.com/netobserv/network-observability-operator/blob/e80a75f18fa0f4c28c1819311415d735038206dd/docs/Architecture.md#cli

jpinsonneau: Yes sounds good. Note that the collector is not involved for the metrics generation. That could be mentionned too

docs/Architecture.md
```diff
@@ -1,49 +1,93 @@
...
+
+_See also: [architecture in the downstream documentation](https://docs.openshift.com/container-platform/latest/observability/network_observability/understanding-network-observability-operator.html#network-observability-architecture_nw-network-observability-operator)_
+
+NetObserv is a collection of components that can sometimes run independently, or as a whole.
+
+The components are:
+
+- An [eBPF agent](https://github.com/netobserv/netobserv-ebpf-agent), that generates network flows from captured packets.
+  - It is attached to any/all of the network interfaces in the host, and listens for packets (ingress+egress) with [eBPF](https://ebpf.io/).
+  - Packets are aggregated into logical flows (similar to NetFlows), periodically exported to a collector, generally FLP.
```
jpinsonneau, line 11: Should we mention the packet capture here ?
That may come in the operator too in the future

docs/Architecture.md
```diff
@@ -1,49 +1,93 @@
...
+When using the CLI, the operator is not involved, which means you can use it without installing NetObserv as a whole. It uses a special mode of the eBPF agents that embeds FLP.
+
+<!-- You can use https://mermaid.live/ to test it -->
+
+```mermaid
+flowchart TD
+    subgraph "for each node"
+        A[eBPF Agent w/ embedded FLP]
+    end
+    A -->|generates flows| C[Collector]
```
jpinsonneau, line 90:
```suggestion
    A -->|generates flows or packets| C[Collector]
```

docs/Architecture.md
jpinsonneau:
```suggestion
When using the CLI, the operator is not involved, which means you can use it without installing NetObserv as a whole. It uses a special mode of the eBPF agents that embeds FLP.

Flows and Packets captures will deploy both agents and collector.
Metrics captures will only deploy agents.
```

jotak: Maybe that's something I missed .. how does metrics work when using the kubectl plugin ? The readme mentions only a make target for capturing metrics. If I run `oc netobserv flows ...` can I get metrics as well?

jotak: nvm I've checked the PR, it's explained: `$ ./build/oc-netobserv metrics --enable_pktdrop="true" --enable_dns="true" --enable_rtt="true"`

next question: README says that's only for openshift, but what if I have prometheus without openshift? I guess I can still have the metrics, it's just the charts that I won't get, correct?

jotak: I added details in last commit https://github.com/netobserv/network-observability-operator/pull/1022/commits/c51d0d754fa9df89f9a9abeeaa3df7a3189279d8

jpinsonneau: Yes if you have a ServiceMonitor you will have the metrics
But the configmap in *openshift* namespace will be ignored

PR #1132, opened 2025-02-11
Label konflux PRs as dependency

jotak:
This should help for release notes categorization


jotak:
once it's merged, it'll start working only on new PRs, so I'll monitor newly created konflux PRs

PR #1165, opened 2025-02-18
Use rhel-els

jotak:


jotak:
looks like ubi9 is FIPS compliant. Closing.

PR #1202, opened 2025-02-25
fix catalog related images

jotak:


catalog/rc.yaml
```diff
@@ -994,14 +994,14 @@ properties:
       name: Red Hat
       url: https://www.redhat.com
 relatedImages:
```
jotak, line 996: these images are injected at build time, we'd better not set anything here, as that could hide a missed injection

PR #1210, opened 2025-03-06
Add QE team approvers on nudging PRs

jotak:


memodi:
thanks @jotak. Do we need similar update for other component repos too?

OlivierCazade (approved):

jotak:
> thanks @jotak. Do we need similar update for other component repos too?

I don't think, as nudging only happens in the operator repo

PR #1257, opened 2025-03-17
NETOBSERV-2174: check if LokiStack exists

jotak:
## Description

If configured LokiStack doesn't exist, it will be reported as a FlowCollector status condition.

Capture d’écran du 2025-03-17 14-28-37

## Dependencies

PR is based on #1255 and  #1258

## Checklist


pkg/manager/status/status_manager.go
```diff
@@ -173,6 +178,52 @@ func checkValidation(ctx context.Context, fc *flowslatest.FlowCollector) metav1.
...
+		}
+	}
+	if fc.Spec.Loki.Mode != flowslatest.LokiModeLokiStack {
+		return metav1.Condition{
+			Type:    LokiIssue,
+			Reason:  "Unused",
+			Status:  metav1.ConditionUnknown,
+			Message: "Loki is not configured in LokiStack mode",
+		}
+	}
```
jpinsonneau, line 197: How does that reports in the conditions ? :thinking: 

```
LokiIssue - Unused - Unknown - Loki is disabled
LokiIssue - Unused - Unknown - Loki is not configured in LokiStack
``` 

Maybe we could rename `LokiIssue` to `Loki` to report all the states as these are not issues.

pkg/manager/status/status_manager.go
```diff
@@ -173,6 +178,52 @@ func checkValidation(ctx context.Context, fc *flowslatest.FlowCollector) metav1.
...
+			Reason:  "Error",
+			Status:  metav1.ConditionTrue,
+			Message: fmt.Sprintf("Error while fetching configured LokiStack: %s", err.Error()),
+		}
+	}
+	return metav1.Condition{
+		Type:   LokiIssue,
+		Reason: "NoIssue",
+		Status: metav1.ConditionFalse,
+	}
```
jpinsonneau, line 224: Why not going one step further here checking LokiStack status and mode ?

pkg/manager/status/status_manager.go
jotak: I don't think because what matters here is that it's a condition that needs to be either TRUE or FALSE (or UNKNOWN). Just "Loki" being true or false doesn't make sense, whereas having LokiIssue true or false does make sense.
When the condition is true, users see in the Console "LokiIssue" so they know there's something to check. If it was just "Loki", that would be unclear. When the condition is false, LokiIssue isn't displayed in the condensed status view, so that's fine

jotak: See my previous PR https://github.com/netobserv/network-observability-operator/pull/1255 that was precisely about making conditions being "negative" so that they render better in the console

jpinsonneau: Ok make sense !

pkg/manager/status/status_manager.go
jotak: sounds good!
=> https://issues.redhat.com/browse/NETOBSERV-2189

PR #1263, opened 2025-03-19
Downstream dockerfiles: manage BUILDVERSION, and some cleanup

jotak:
- BUILDVERSION previsouly did not work. Add it as a global argument for all dockerfiles
- Remove COMMIT arg - it doesn't work, and actual upstream commit is anyway correctly labelled anyway as "vcs-ref" (not upstream-vcs-ref)
- Hence, remove upstream-vcs-ref
- Add 14d expiry on on-pushed images
- Bump versions in update-builds.sh


.tekton/network-observability-operator-bundle-push.yaml
```diff
@@ -30,6 +30,10 @@ spec:
     value: '{{revision}}'
   - name: output-image
     value: quay.io/redhat-user-workloads/ocp-network-observab-tenant/netobserv-operator/network-observability-operator-bundle:{{revision}}
+  - name: image-expires-after
+    value: 14d
```
jotak, line 34: @OlivierCazade I'm adding that because we generate a lot of builds and it surely is going to fill the quay repo :-/ 
idk if 14d is ok - I guess the point is that the last build has to be still there when it's time to release. If necessary maybe we can increase to 30d.

OlivierCazade (approved):
LGTM

PR #1342, opened 2025-04-01
Rename "sampling rate" => "sampling ratio"

jotak:
"Sampling rate" could be misleading, it could be understood as the number of flows/packets per second that are sampled. "Ratio" is the correct term to use here, as it refers to the proportion of packets or flows to sample.


OlivierCazade (approved):
LGTM

PR #1345, opened 2025-04-02
NETOBSERV-1942 - Always create console plugin

jpinsonneau:
## Description

Create the console plugin when FlowCollector doesn't exists to expose the new forms.

See https://github.com/netobserv/network-observability-console-plugin/pull/763 for the forms

## Dependencies

n/a

## Checklist


controllers/flowcollector_controller.go
```diff
@@ -74,6 +77,14 @@ func Start(ctx context.Context, mgr *manager.Manager) error {
 	}
 	r.watcher = watchers.NewWatcher(ctrl)
 
+	reconcileAfter := func() {
+		// force reconcile at startup
+		time.Sleep(30 * time.Second)
+		_, err := ctrl.Reconcile(ctx, reconcile.Request{})
+		log.Error(err, "Error while doing initial reconcile")
+	}
+	go reconcileAfter()
```
jpinsonneau, line 86: :point_up: I wonder if there is an out of box mechanism to trigger the loop after the cache loaded. That's why I'm using a sleep here and this will may work in all situations.

https://redhat-internal.slack.com/archives/C02939DP5L5/p1743518264445429

controllers/flowcollector_controller.go
```diff
@@ -74,9 +78,30 @@ func Start(ctx context.Context, mgr *manager.Manager) error {
...
+		time.Sleep(5 * time.Second)
+		_, err = r.Reconcile(ctx, reconcile.Request{})
+		if err != nil {
+			log.Error(err, "Error while doing initial reconcile", "attempt", attempt)
+		} else {
+			break
+		}
+	}
+	return err
+}
```
jpinsonneau, line 103: ☝️ I wonder if there is an out of box mechanism to trigger the loop after the cache loaded. That's why I'm using a sleep here and this will may work in all situations.

https://redhat-internal.slack.com/archives/C02939DP5L5/p1743518264445429

OlivierCazade: If I remember correctly, when a reconcile loop is failing, you can also return a time value to reschedule the reconciliation.

jpinsonneau: Yeah I gave a try with that without success. 

I'm refactoring the code again to move the static content to another controller wich will be cleaner I guess. I will give another try with the reschedule time on the new controller :+1:

controllers/static/static_controller.go
```diff
@@ -0,0 +1,99 @@
...
+		// delay the reconcile calls to let some time to the cache to load
+		time.Sleep(5 * time.Second)
+		_, err := r.Reconcile(ctx, ctrl.Request{})
+		if err != nil {
+			log.Error(err, "Error while doing initial reconcile", "attempt", attempt)
+		} else {
+			return
+		}
+	}
+}
```
jpinsonneau, line 61: @OlivierCazade I modified the PR to be in a dedicated controller.

As you can see, I force the `Reconcile` call during the `Start` function so OLM can't interpret the result containing Requeue / RequeueAfter here.

Since we don't create a dedicated CR for static plugin, I don't think we can rely on these here.

WDYT ?

controllers/static/static_controller.go
```diff
@@ -0,0 +1,99 @@
...
+// It manages the controller status at a high level. Business logic is delegated into `reconcile`.
+func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	l := log.Log.WithName("staticPlugin") // clear context (too noisy)
+	ctx = log.IntoContext(ctx, l)
+
+	r.status.SetUnknown()
+	defer r.status.Commit(ctx, r.Client)
+
+	// always reconcile static console plugin
+	scp := helper.UnmanagedClient(r.Client)
```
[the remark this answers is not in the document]
jotak: what about writing the operator deployment as the owner reference? Wouldn't that work automatically, ie. kube would delete it when the operator deployment is deleted as well? The good thinks would be that it would work regardless of the installation context ie. with or without a CSV/subscription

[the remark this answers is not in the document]
jotak: I think this ownerRef would work:
```
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: false
    controller: false
    kind: Deployment
    name: netobserv-controller-manager
    uid: <uid>
```

.. we just need to grab the actual uid..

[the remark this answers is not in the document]
jpinsonneau: I can give a try :ok_hand:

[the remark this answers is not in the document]
jpinsonneau: That's partially working as it can't manage cluster scoped resources such as `ClusterRole` :thinking: 

It's working fine with namespaced resources :partying_face: 
see https://github.com/netobserv/network-observability-operator/pull/1345/commits/fc521bc9be7cec4e832399a3bafbb585151dd4e3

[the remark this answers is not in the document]
jpinsonneau: Is there a magic trick for the tests ? 
https://github.com/netobserv/network-observability-operator/actions/runs/14755226631/job/41421690647?pr=1345

The controller deployment doesn't exists there

memodi:
@jpinsonneau there seems to be an issue where plugin does become ready, despite pod is in Running state and no issues with Loki whatsoever:

```
time="2025-05-05T20:00:34Z" level=error msg="cannot unmarshal, response was: Ingester not ready: waiting for 15s after being ready\n" error="invalid character 'I' looking for beginning of value" module=handler
time="2025-05-05T20:00:34Z" level=error msg="cannot unmarshal, response was: Ingester not ready: waiting for 15s after being ready\n" error="invalid character 'I' looking for beginning of value" module=handler
```

image

the static plugin pod had this error:
```
$ oc -n openshift-netobserv-operator logs pod/netobserv-plugin-static-7c97f478bf-t7gj8
time="2025-05-05T20:10:16Z" level=info msg="Starting netobserv-console-plugin [build version: main-2c3b33b, build date: 2025-05-05 10:42] at log level info" module=main
time="2025-05-05T20:10:16Z" level=fatal msg="invalid config" error="neither Loki nor Prometheus is configured; at least one of them should have a URL defined" module=main
```
I had monolithic loki configured in my flowcollector - anything I am missing for static console plugin config?

controllers/static/static_controller.go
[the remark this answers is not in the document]
jotak: > That's partially working as it can't manage cluster scoped resources such as ClusterRoleBinding

But, is there any of these cluster-scope resources needed by the static plugin? I don't think any `ClusterRoleBinding` is required here, right? How about the `ConsolePlugin` resource? It is cluster-scope, and needed here, no?

jpinsonneau:
> @jpinsonneau there seems to be an issue where plugin does become ready, despite pod is in Running state and no issues with Loki whatsoever:
> 
> ```
> time="2025-05-05T20:00:34Z" level=error msg="cannot unmarshal, response was: Ingester not ready: waiting for 15s after being ready\n" error="invalid character 'I' looking for beginning of value" module=handler
> time="2025-05-05T20:00:34Z" level=error msg="cannot unmarshal, response was: Ingester not ready: waiting for 15s after being ready\n" error="invalid character 'I' looking for beginning of value" module=handler
> ```
> 
> image
> 
> the static plugin pod had this error:
> 
> ```
> $ oc -n openshift-netobserv-operator logs pod/netobserv-plugin-static-7c97f478bf-t7gj8
> time="2025-05-05T20:10:16Z" level=info msg="Starting netobserv-console-plugin [build version: main-2c3b33b, build date: 2025-05-05 10:42] at log level info" module=main
> time="2025-05-05T20:10:16Z" level=fatal msg="invalid config" error="neither Loki nor Prometheus is configured; at least one of them should have a URL defined" module=main
> ```
> 
> I had monolithic loki configured in my flowcollector - anything I am missing for static console plugin config?

@memodi are you using the proper plugin image mentionned above ?
https://github.com/netobserv/network-observability-console-plugin/pull/763

controllers/static/static_controller.go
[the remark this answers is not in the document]
jotak: > Is there a magic trick for the tests ?

Hmm you could cheat and create a dummy deployment. Which would also be a way to test that removing that deployment does remove the other resources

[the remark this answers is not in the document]
jpinsonneau: > > That's partially working as it can't manage cluster scoped resources such as ClusterRoleBinding
> 
> But, is there any of these cluster-scope resources needed by the static plugin? I don't think any `ClusterRoleBinding` is required here, right? How about the `ConsolePlugin` resource? It is cluster-scope, and needed here, no?

The `/api` endpoint is checking for user auth which requires the `ClusterRoleBinding`.
I kept it to expose `status` and `config` even in the static plugin. I can remove it if we feel it's useless but we may face this back in the future if we want to rely on backend resources.

> > Is there a magic trick for the tests ?
> 
> Hmm you could cheat and create a dummy deployment. Which would also be a way to test that removing that deployment does remove the other resources

The dummy deployment would need an image to point :thinking: 
Plus the static reconciler needs a namespace as argument before creating the flow collector. I'm not sure how to do that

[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/1345/commits/0daa95e544ebe19aebbc7ef1d74ef4db654c4893

memodi:
> @memodi are you using the proper plugin image mentionned above ?
> [netobserv/network-observability-console-plugin#763](https://github.com/netobserv/network-observability-console-plugin/pull/763)

I was not, I didn't realize that's dependency. I'll try with it today. Thanks!

jpinsonneau:
@memodi FYI I did some changes to address https://github.com/netobserv/network-observability-operator/pull/1345#discussion_r2074986325 comment

Just tested and the behavior remains the same :wink: 
It avoid using an unecessary service account and role

controllers/static/static_controller.go
```diff
@@ -0,0 +1,103 @@
...
+		For(&flowslatest.FlowCollector{}, reconcilers.IgnoreStatusChange).
+		Named("staticPlugin").
+		Complete(&r)
+}
+
+func (r *Reconciler) InitReconcile(ctx context.Context) {
+	log := log.FromContext(ctx)
+	log.Info("Initializing resources...")
+
+	for attempt := range initReconcileAttempts {
```
jotak, line 52: btw, what happens if all 5 attempts fail? The static plugin wouldn't deploy, but the rest would work normally? Or does it may the controller CLBO or so ?

controllers/static/static_controller.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Reconcile is the controller entry point for reconciling current state with desired state.
+// It manages the controller status at a high level. Business logic is delegated into `reconcile`.
+func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	l := log.Log.WithName("staticPlugin") // clear context (too noisy)
+	ctx = log.IntoContext(ctx, l)
+
+	r.status.SetUnknown()
+	defer r.status.Commit(ctx, r.Client)
+
```
jotak, line 72: After setting "Unknown", I think this controller should return if openshift isn't detected, right?
We could check it in Kind

jotak (commented):
Just a comment when not running on openshift, other than that lgtm

controllers/static/static_controller.go
jpinsonneau: If the attempts fails, the static plugin will not be there at controller startup.
As soon as a reconcile loop is triggered, it will appears (ie creating a FlowCollector for example)

controllers/static/static_controller.go
jpinsonneau: That's done in the static reconciler using `HasConsolePlugin` function:
https://github.com/netobserv/network-observability-operator/blob/1080b69c915d647c11b75da7d39976b537171dbe/controllers/consoleplugin/consoleplugin_static_reconciler.go#L40-L77

If console plugin is not available, nothing is deployed and the status goes ready

jpinsonneau: The static controller could deploy something else than the console plugin in future so I think it's better to keeps things separated here. WDYT ?

jotak: ah ok, yes sounds good, thanks!

memodi:
@jpinsonneau - could you rebase this PR please? I tried locally using `make` commands but running into other issues.

jpinsonneau:
Rebased without changes

memodi:
@jpinsonneau Here's quick feedback from my initial review of the form view:

```
- ebpf flow filtering
	- what does option 1 and 2 mean for port fields? [1]
- Deployment model
- Mode:
	- add a note which Loki mode is recommended for production use cases. 
- Prometheus Mode:
	- add a node what "Auto" mode mean? Or hide it if its not relevant 
- pipeline stage:
	- flow filtering.
	- some advanced configuration like "secondaryNetworks" could be exposed.
```
[1]

Screenshot 2025-07-15 at 4 50 19 PM

at the end it generated empty yaml [2] , I was trying with 4.20 OCP version

Screenshot 2025-07-15 at 4 58 34 PM

I'll add more feedback as I test more.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -1258,6 +1259,10 @@ spec:
                   value: quay.io/netobserv/network-observability-console-plugin:v1.9.1-community
                 - name: RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT
                   value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.8.2-community
+                - name: NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
```
memodi, line 1268: @jpinsonneau - can we append this env var instead of adding it in middle so that index # for the existing vars doesn't change? We have several scripts and CI steps where we patch  csv which happens based on index of env var.

memodi:
Adding more feedback:

1. A way to skip through all the config options would be nice for users who are looking to just configure with default config.
2. Is there a way to have a hyperlink of a diagram presented on [k8s/cluster/flows.netobserv.io~v1beta2~FlowCollector/status] of each component for the "status" field shown on [flows.netobserv.io~v1beta2~FlowCollector] page
3. the view port of the diagram is very small on my 16" laptop screen
Screenshot 2025-07-16 at 10 12 46 PM

4. When I used kafka mode, the FLP component never went green even though it was working fine.
5. nit: the lines between FLP --> Loki/Prom. keeps flickering.

https://github.com/user-attachments/assets/71255fe7-853f-41d8-a4d6-f848c57ef53c

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: Sure, I cant put it in the end of the list if it help :smile_cat:

jpinsonneau:
> @jpinsonneau Here's quick feedback from my initial review of the form view:
> 
> ```
> - ebpf flow filtering
> 	- what does option 1 and 2 mean for port fields? [1]
> - Deployment model
> - Mode:
> 	- add a note which Loki mode is recommended for production use cases. 
> - Prometheus Mode:
> 	- add a node what "Auto" mode mean? Or hide it if its not relevant 
> - pipeline stage:
> 	- flow filtering.
> 	- some advanced configuration like "secondaryNetworks" could be exposed.
> ```

Sure I can add some notes and recommendations on top of the current descriptions. These are coming from the API so maybe we can also improve it.

About the options 1 / 2 and the advanced config, we need to decide what to expose and what to skip as these needs extra tweaking to make it work. 
For example, options 1 & 2 are about using a single port or an array or range:
https://github.com/netobserv/network-observability-operator/blob/main/internal/controller/ebpf/agent_controller.go#L560-L572

If we decide to not expose these, I will simply hide these fields.

jpinsonneau:
> 1. A way to skip through all the config options would be nice for users who are looking to just configure with default config.

You can directly click on the left menu to navigate to the last step. Isn't it enough ?

> 2. Is there a way to have a hyperlink of a diagram presented on [k8s/cluster/flows.netobserv.io~v1beta2~FlowCollector/status] of each component for the "status" field shown on [flows.netobserv.io~v1beta2~FlowCollector] page

Sure, good idea !

> 3. the view port of the diagram is very small on my 16" laptop screen

Indeed. Would you prefer to have a scrollbar on the status list or a scrollbar on the entire page ?

> 4. When I used kafka mode, the FLP component never went green even though it was working fine.

Indeed I should rely on the `WaitingFLPTransformer` for kafka:
https://github.com/netobserv/network-observability-console-plugin/pull/763/files
I'll do the change

> 5. nit: the lines between FLP --> Loki/Prom. keeps flickering.

I can see if I can skip some refreshs here to avoid that. Thanks for your feedback !

memodi:
> You can directly click on the left menu to navigate to the last step. Isn't it enough ?

ah, it wasn't clear me, that I could jump that way.

> Would you prefer to have a scrollbar on the status list or a scrollbar on the entire page ?

I think scrollbar on status list make sense to me.

> About the options 1 / 2 and the advanced config, we need to decide what to expose and what to skip as these needs extra tweaking to make it work.
For example, options 1 & 2 are about using a single port or an array or range:
https://github.com/netobserv/network-observability-operator/blob/main/internal/controller/ebpf/agent_controller.go#L560-L572
If we decide to not expose these, I will simply hide these fields.

I think these are much advanced config which can be handled directly in yaml.

of all the advancedConfig, IMHO secondaryNetworks may probably be more utilized than others, may be it make a case to move it out of advanced config in API for it.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/1345/commits/9bb32d16c2e546459617e4a9147a6f5d29efbbf8

jpinsonneau:
@memodi coming back on the notes, that's what we have today:
image
image

I would suggest to update the CRD first and then in the forms if we are not satisfied with those texts

memodi:
> I would suggest to update the CRD first and then in the forms if we are not satisfied with those texts

sure, np, though it would be nice to have a way to override API docs text

jotak (approved):
/lgtm

jotak:
another round of testing will be done post-merge; merging now!

PR #1346, opened 2025-04-02
Create static console plugin in OLM bundle

jotak:


jotak:
Closing because this approach doesn't work for installing custom resources (ConsolePlugin); Unless OLM implements that, it leaves us with two other options:
- #1345 install console plugin by the operator, outside of the FlowCollector reconcile loop
- or the hybrid approach: keep OLM-owned deployment and service like in this PR, and the ConsolePlugin installed by the operator.

Issue #1064, opened 2025-01-28
Helm: pipeline integration


jotak:
Follow-up on [adding a helm chart](https://github.com/netobserv/network-observability-operator/issues/1000) (#1000)

Currently, the helm chart is only static files. We should make them more integrated:

- With make targets that update the container image tags
- Check how to integrate that into ArtifactHub
- If possible, automate publishing on ArtifactHub via our Github action "release" workflow

Helion55:
Hello @jotak Is there anything you want me to work on regarding this issue?

jotak:
Hi @Helion55 ,
thanks for proposing your help again!
We're currently looking into publishing the 1.8.0 community version, I'm going to push the related images (the ones mentioned [here](https://github.com/netobserv/network-observability-operator/pull/1113/files#diff-a63ea33c81187c63a5163e4042419e5cce470581f485a1c99914c655f5570435R36-R45))
In #1113 I'm adding some automation for updating the helm charts on new releases, mostly [here](https://github.com/netobserv/network-observability-operator/pull/1113/files#diff-ee216438387c6afe859c18074f760a6619b3a7b76c17dac4d7acf569753ac93aR1) : scripting some of the manual changes that we did previously in your PR

Now, the part that I'm less familiar with, is ArtifactHub. If I'm correct, this is the go-to solution for anyone who want to publish helm charts, right? Maybe you're more familiar with it than me. Basically, I'm reading this:
- https://artifacthub.io/docs/topics/repositories/helm-charts/
- https://helm.sh/docs/topics/chart_repository/

So either we need to have a server that exposes the charts + an index.yaml, or we can do it via an OCI registry. We already serve our images via quay.io which I guess can be used as the OCI registry, but I don't know at this point what's better for us between the server or the OCI registry - if you happen to have some experience already in that field, I'm happy to get your advice! Else we can figure out together.

We also have github pages set up ( https://github.com/netobserv/netobserv.github.io / https://netobserv.io/ ) so that can probably be used for the non-OCI alternative.

Helion55:
Thanks @jotak .Got the idea, starting to work on the ArtifactHub publishing template. I think It required a much more dynamic chart than it is right now. Share your thoughts on that.

[...]

PR #1433, opened 2025-04-22
enhancement : Helm Charts upgraded 

Helion55:
## Description
Helm charts upgraded with,
1. Chart dependencies containing Prometheus and Loki chart which can be modified by user with ```--set``` flag during installation,
2. A separate namespace for every k8s objects of this deployment,
3. Parameterized image-name and image-version values in deployment.yaml file.
This will help to update those values from automated workflow pipeline. 

## Issue

#1064

## Proposing Further activities
1. Using a workflow to update those image values on every upgradtion.
2. Writting a description and additional instruction about this helm release in NOTES.txt
3. Also adding cert-manager as dependency chart.


jotak:
Thanks @Helion55 ! I'll test it!
One remark is, I'm not sure about hard-coding the namespace. It's just an assumption but I think users often want to organize their namespaces as they want...

jotak:
Hi @Helion55 
I'm getting errors with the namespace "netobserv" not found. I suspect there must be issues with the ordering of installations, e.g. if the namespace isn't created yet when other resources are installed? TBH I think that would be another reason why it's better to let helm deal with namespaces rather than forcing it.

jotak:
>     2. Writting a description and additional instruction about this helm release in NOTES.txt

Why not using the existing README.md in `./helm` ?

jotak:
Why do we need loki & prometheus tgz embedded there? I don't think we're supposed to host them, they should come from their respective repositories

jotak:
I'm also seeing that the loki default install includes promtail, which we don't need here, we should see if that's something we can remove by config

Helion55:
> > ```
> > 2. Writting a description and additional instruction about this helm release in NOTES.txt
> > ```
> 
> Why not using the existing README.md in `./helm`

Yes, no issue; we can refer to that. If anything more meaningful needs to be displayed, suggest it to me, and I will add that. Currently, I am just showing the manifest to apply and the port-forward steps, just to make things handy.

Helion55:
> Why do we need loki & prometheus tgz embedded there? I don't think we're supposed to host them, they should come from their respective repositories
> 
> [edit] hmmm after trying without, helm complains they are missing, so I guess it's needed indeed?

No, it's not mandatory to embed there. We can run,
``` helm dependency build ```
to install them can then execute ``` helm install ```  command.

Helion55:
> I'm also seeing that the loki default install includes promtail, which we don't need here, we should see if that's something we can remove by config

Alright, finding something for it.

jotak:
hey, sorry I rebased and pushed before noticing this was your `main` branch, I don't want to mess thing up for you, so I've rolled back my rebase.

I'm going to take a stab at these changes. One thing that will ease everything is that we don't need the conversion webhook at this point, I'll remove it  from the CRD, which will also remove the need for the namespace here.

jotak:
Thanks @Helion55, your contributions are mentioned in the release [1.9.1-community](https://github.com/netobserv/network-observability-operator/releases/tag/1.9.1-community). I'm closing this PR which was replaced with https://github.com/netobserv/network-observability-operator/pull/1712

PR #1483, opened 2025-05-06
CRD doc: update feature statuses

jotak:
UDN and filters are GA

### DIFF
--- a/apis/flowcollector/v1beta1/flowcollector_types.go
+++ b/apis/flowcollector/v1beta1/flowcollector_types.go
@@ -151,11 +151,11 @@ type FlowCollectorIPFIX struct {
 // Agent feature, can be one of:<br>
 // - `PacketDrop`, to track packet drops.<br>
 // - `DNSTracking`, to track specific information on DNS traffic.<br>
-// - `FlowRTT`, to track TCP latency [Unsupported (*)].<br>
-// - `NetworkEvents`, to track Network events.<br>
-// - `PacketTranslation`, to enrich flows with packets translation information. <br>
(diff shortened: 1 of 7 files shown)


OlivierCazade (approved):
LGTM

apis/flowcollector/v1beta1/flowcollector_types.go
```diff
@@ -151,11 +151,11 @@ type FlowCollectorIPFIX struct {
...
 // - `DNSTracking`, to track specific information on DNS traffic.<br>
-// - `FlowRTT`, to track TCP latency [Unsupported (*)].<br>
-// - `NetworkEvents`, to track Network events.<br>
-// - `PacketTranslation`, to enrich flows with packets translation information. <br>
-// - `EbpfManager`, to enable using EBPF Manager to manage netobserv ebpf programs [Developer Preview].<br>
-// - `UDNMapping`, to enable interfaces mapping to udn [Developer Preview]. <br>
+// - `FlowRTT`, to track TCP latency.<br>
+// - `NetworkEvents`, to track network events [Technology Preview].<br>
+// - `PacketTranslation`, to enrich flows with packets translation information, such as Service NAT.<br>
+// - `EbpfManager`, to enable using eBPF Manager to manage NetObserv eBPF programs. [Unsupported (*)].<br>
```
Amoghrd, line 157: This is also Technology Preview right?

jotak: iirc we're indeed calling it TP internally but for downstream doc it's not mentioned as such to avoid some additional hassle. @skrthomas do you confirm?
Network Events feature is different because it _is_ also a TP considered from core ocp point of view.

PR #1484, opened 2025-05-06
NETOBSERV-2101: remove CAP_SYS_RESOURCE, allow override

jotak:
## Dependencies

n/a

## Checklist


apis/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -1247,6 +1248,10 @@ type AdvancedAgentConfig struct {
 	// scheduling controls how the pods are scheduled on nodes.
 	// +optional
 	Scheduling *SchedulingConfig `json:"scheduling,omitempty"`
+
+	// Linux capapbilities override, when not running as privileged. Default capabilities are BPF, PERFMON and NET_ADMIN.
+	// +optional
+	CapOverride []string `json:"capOverride,omitempty"`
```
jpinsonneau, line 1254: Why not calling that capabilities and set default values ?

jotak: I prefer to not set default (openAPI) values as that doesn't work well with future changes of the defaults (when we change the defaults, people who upgrade keep the previous default values).
What we do sometimes is to use a nil/pointer field and write defaults in code, instead of via openAPI; which is pretty much what I'm doing here.

jpinsonneau: ok let's see if someone complains about it or not :smile_cat:

jpinsonneau (approved):

memodi:
/label qe-approved

verified on older kernel 4.12 and 4.14 OCP versions, all sanity tests passes.

PR #1517, opened 2025-05-15
NETOBSERV-2184 improve cache & reconcile events

jpinsonneau:
## Description

- remove unecessary info from cache
  - status conditions / container info
  - managed fields
  - annotations
- reduce reconcile calls

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -63,7 +63,20 @@ func NewManager(
 	log := log.FromContext(ctx)
 	log.Info("Creating manager")
 
-	narrowCache := narrowcache.NewConfig(kcfg, narrowcache.ConfigMaps, narrowcache.Secrets)
+	narrowCache := narrowcache.NewConfig(kcfg,
+		narrowcache.ConfigMaps,
+		// narrowcache.ConsolePlugin,
+		narrowcache.Daemonsets,
+		narrowcache.Deployments,
+		narrowcache.HorizontalPodAutoscalers,
+		narrowcache.Namespaces,
+		narrowcache.NetworkPolicies,
+		narrowcache.Pods,
+		narrowcache.Secrets,
+		// narrowcache.SecurityContextConstraints,
+		narrowcache.Services,
+		narrowcache.ServiceAccounts,
+	)
 	opts.Client = client.Options{Cache: narrowCache.ControllerRuntimeClientCacheOptions()}
 
 	internalManager, err := ctrl.NewManager(kcfg, *opts)
--- a/pkg/narrowcache/client.go
+++ b/pkg/narrowcache/client.go
@@ -6,6 +6,14 @@ import (
 	"reflect"
 	"sync"
 
+	osv1 "github.com/openshift/api/console/v1"
+	securityv1 "github.com/openshift/api/security/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	ascv2 "k8s.io/api/autoscaling/v2"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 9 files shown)


pkg/narrowcache/client.go
```diff
@@ -103,19 +111,82 @@ func (c *Client) getAndCreateWatchIfNeeded(ctx context.Context, info GVKInfo, gv
 	return fetched.(client.Object), objKey, nil
 }
 
-// "Terrible hack" cc directxman12 / sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go
 func copyInto(obj runtime.Object, out client.Object) error {
-	// Copy the value of the item in the cache to the returned value
-	// TODO(directxman12): this is a terrible hack, pls fix (we should have deepcopyinto)
+	// cleanup unecessary fields
```
jotak, line 115: I'm not sure about cleaning up annotations. We use annotations, such as for storing config digests, aren't we going to loose that?

jotak: Also this function is called when we retrieve an object from the cache.
Isn't the intent to do this cleanup when the objects are _set_ in cache? ie. in `setToCache` ?

pkg/manager/manager.go
```diff
@@ -63,7 +63,20 @@ func NewManager(
...
-	narrowCache := narrowcache.NewConfig(kcfg, narrowcache.ConfigMaps, narrowcache.Secrets)
+	narrowCache := narrowcache.NewConfig(kcfg,
+		narrowcache.ConfigMaps,
+		// narrowcache.ConsolePlugin,
+		narrowcache.Daemonsets,
+		narrowcache.Deployments,
+		narrowcache.HorizontalPodAutoscalers,
+		narrowcache.Namespaces,
+		narrowcache.NetworkPolicies,
+		narrowcache.Pods,
```
jotak, line 74: do we get/watch any Pod at some point?

pkg/narrowcache/client.go
jpinsonneau: Can you please point me where we use these ? 

I can add some exceptions for the ones needed. It's on purepose to remove unused content before we consume the object data to ensure we are not using it and be able to compare items from API or cache seamlessly.

Removing the data when we cache it may result in inconsistent behavior depending where the info comes from.

jotak: We use annotations a lot, e.g. [here](https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/consoleplugin_objects.go#L221) or [there](https://github.com/netobserv/network-observability-operator/blob/main/controllers/consoleplugin/consoleplugin_objects.go#L276)

But I get why it seems to work, for Deployments/DaemonSets we use the PodTemplate annotations, not the ones of the resource itself. And for Services, [we use deepDerivative](https://github.com/netobserv/network-observability-operator/blob/main/pkg/helper/comparators.go#L138) ie. we consider they change if a previous annotation is changed/removed (but not if new annotations are added) - with your change I think it comes down to not watching for annotation changes (since the "old" annotations will always be empty.

tbh perhaps that works but I'd still remove the annotation erasure, because it can easily hit us later, e.g. if we have anything to do with annotations. Moreover, on narrowcache I really don't think we're talking about a lot of saved space, since this cache only accumulate resources that we actually manage, we're talking about just a couple of Deployments/DaemonSets/Services, I don't think we need to transform them much. ManagedFields is certainly the most relevant thing to cleanup here, I would say.

jotak: An example of impact: if I edit `netobserv-plugin` service to change the cert annotation `service.beta.openshift.io/serving-cert-secret-name`, it won't reconcile now

jpinsonneau:
Moving back to draft as it requires some changes after testing on a bigger cluster

pkg/manager/manager.go
jpinsonneau: no, it's actually the `PodTemplate` I saw in pprof which comes from Deployments so this is never called. Gonna remove it.

Thanks

pkg/narrowcache/client.go
jpinsonneau: ok thanks for the details. Gonna restore it

jpinsonneau:
@jotak any idea on how to make the linter happy ?
https://github.com/netobserv/network-observability-operator/actions/runs/15115971335/job/42486600090?pr=1517

Dockerfile.downstream
```diff
@@ -1,7 +1,7 @@
 ARG BUILDVERSION
 
 # Build the manager binary
-FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.23 as builder
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.24 as builder
```
jotak, line 4: I don't think that exists at the moment (?)

go.mod
```diff
@@ -1,13 +1,14 @@
 module github.com/netobserv/network-observability-operator
 
-go 1.23.0
+go 1.24.0
```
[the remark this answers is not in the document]
jotak: Yes I know I delayed the k8s 0.33 bump for that reason too (dependabot and konflux keep opening PRs about that)

pkg/narrowcache/client.go
```diff
@@ -52,10 +59,7 @@ func (c *Client) Get(ctx context.Context, key client.ObjectKey, out client.Objec
 		if err != nil {
 			return err
 		}
-		err = copyInto(obj, out)
-		if err != nil {
-			return err
-		}
+		copier.Copy(out, obj)
```
jotak, line 62: I wanted to check how this `copier` performed versus the old hack, and it seems, pretty badly :(

Creating those benchmarks:

```go
func setup() corev1.ConfigMap {
	m := corev1.ConfigMap{
		ObjectMeta: v1.ObjectMeta{
			Labels: map[string]string{},
		},
		Data: map[string]string{},
	}
	for i := 0; i < 20; i++ {
		m.Labels[fmt.Sprintf("key-%d", i)] = fmt.Sprintf("value-%d", i)
		m.Data[fmt.Sprintf("key-%d", i)] = fmt.Sprintf("value-%d", i)
	}
	return m
}

func Benchmark_Copier(b *testing.B) {
	in := setup()
	for i := 0; i < b.N; i++ {
		out := corev1.ConfigMap{}
		_ = copier.Copy(&out, &in)
	}
}

func Benchmark_OldHack(b *testing.B) {
	in := setup()
	for i := 0; i < b.N; i++ {
		out := corev1.ConfigMap{}
		_ = copyInto(&in, &out)
	}
}
```

Then, running them:
```
Benchmark_Copier-8    	   23798	     50697 ns/op	   19888 B/op	     529 allocs/op
Benchmark_OldHack-8   	 9953776	       130.8 ns/op	     288 B/op	       1 allocs/op
```
... which makes the hack almost 400 times faster!
Which makes sense I guess as copier does a deep copy whereas the hack does some sort of reassignment trick if I understand correctly

jotak: fwiw the hack is still there in controller-runtime https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/internal/cache_reader.go#L93

jpinsonneau: yes the hack will ever be faster but I suspect it involves a memory leak in the end

jotak:
> @jotak any idea on how to make the linter happy ? https://github.com/netobserv/network-observability-operator/actions/runs/15115971335/job/42486600090?pr=1517

I must say I don't understand the error; but it seems related to bumping to go 1.24, as I don't get it with 1.23
So if we stick to 1.23, that should pass ... until next time

pkg/narrowcache/client.go
```diff
@@ -138,10 +126,54 @@ func (c *Client) updateCache(ctx context.Context, key string, watcher watch.Inte
 }
 
 func (c *Client) setToCache(key string, obj runtime.Object) error {
+	// cleanup unecessary fields
```
jotak, line 129: could you move this part into the `GVKInfo` struct e.g. as a new `Cleanup` callback?
Because the `narrowcache` client is designed in a way to allow callers to extend what types are managed, by adding their own `GVKInfo` if needed, and this big switch kind of break the pattern.

Also, I think emptying `SetManagedFields` could be called regardless of the type, as it's in `ObjectMeta`.

pkg/narrowcache/client.go
jotak: just ran some pprof alongside with a comparison between copier vs old hack, and couldn't find a memory leak. If no evidence of a leak I think we should keep the hack as it's more efficient. I'm currently running a NDH test with your PR ( https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/65121/rehearse-65121-periodic-ci-openshift-eng-ocp-qe-perfscale-ci-netobserv-perf-tests-netobserv-aws-4.19-nightly-x86-node-density-heavy-25nodes/1925115161513824256 )

What we can do is, after that, rollback this single change (with copier) and run again a NDH

jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/1517/commits/4fcdf8e2a33912578f5d78d9776785acefcf83df

pkg/narrowcache/client.go
jpinsonneau: https://github.com/netobserv/network-observability-operator/pull/1517/commits/57a1cab153a55899ccff3314cff4c89516f02610

I kept the types to have the autocompletion in vscode for future addition

Dockerfile.downstream
jpinsonneau: how do you check that ? :thinking:

jpinsonneau:
> Just realized we don't track the operator metrics... and unfortunately the per-pod metrics that we track are flawed when there's pods restart (which seemed to happen e.g. [here](https://docs.google.com/spreadsheets/d/1tUHNIGOeis5P6sDL1mBtA8MqpfnJDimANQ16F6aTs1U/edit?gid=0#gid=0) as it shows several operator pods) I've opened a PR to add the operator metrics to our tests: [openshift-eng/ocp-qe-perfscale-ci#742](https://github.com/openshift-eng/ocp-qe-perfscale-ci/pull/742)
> 
> In the meantime, I've compared with another run that I've found without pod restart: [Previous run](https://docs.google.com/spreadsheets/d/1hr__X6dcyYNwxlxJLe5Bmi4WLY0-pQfb10wWV_vRUuo/edit?gid=0#gid=0): RSS = 115979810.1 [New run](https://docs.google.com/spreadsheets/d/1UiNrZOAHad7UNf9GDHjBZFerW7r3IoRxgxcqn_VCy7I/edit?gid=0#gid=0) (this PR): RSS = 105546001.1
> 
> So that's a -9% RSS

@jotak Is it worth running a new test since https://github.com/openshift-eng/ocp-qe-perfscale-ci/pull/742 is now merged ?

Dockerfile.downstream
jotak: Sorry I missed that comment!
I'm just doing:

```
podman pull --authfile=/path/to/pull-secret.json brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.24
```

btw, it's now available :-)

Amoghrd:
@jotak @jpinsonneau For testing this will regression tests suffice? 
We can look at the perf impact post-merge on everyday runs. 
Is there anything extra that needs to be done?

jotak:
@Amoghrd yes it's just regression + perf, which can be checked on the periodic jobs

Amoghrd:
Here's the regression run [ginkgo-test#294762](https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/ocp-common/job/ginkgo-test/294762/console) All tests passed on rerun except the alerts test. Its failing in [test_flowcollector.go#L201](https://github.com/openshift/openshift-tests-private/blob/release-4.20/test/extended/netobserv/test_flowcollector.go#L201) wherein the flowcollector does not come up ready when set with incorrect Loki URL. I was checking the cluster when the test was running and see all pods are running(FLP, ConsolePlugin, eBPF) but still the flowcollector status still shows waiting for FLPMonolith and does not come Ready at all.
Test is passing on latest downstream operator version

jotak:
@Amoghrd good catch, that's exactly the kind of thing I was fearing :-)
patch's on its way

jotak:
@jpinsonneau I'm also wondering if that could break your recent/planned changes on the custom status page. I don't remember, do we watch non-owned component status already, like Loki or Kafka? If so, we should relax the watch-update predicate...

jpinsonneau:
> @jpinsonneau I'm also wondering if that could break your recent/planned changes on the custom status page. I don't remember, do we watch non-owned component status already, like Loki or Kafka? If so, we should relax the watch-update predicate...

The status view indeed show our components based on FlowCollector status `field` + Loki errors also based on the same field.

Amoghrd:
Will rerun regression on new image

Amoghrd:
Regression reun passed [ginkgo-test#294797](https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/ocp-common/job/ginkgo-test/294797/console) (with known flakes). 
/label qe-approved

PR #1632, opened 2025-06-12
NETOBSERV-2182 & NETOBSERV-2183 PoC : Filters refactoring

jpinsonneau:
## Description

Update columns configs for https://github.com/netobserv/network-observability-console-plugin/pull/877

## Dependencies

n/a

## Checklist

### DIFF
--- a/controllers/consoleplugin/config/static-frontend-config.yaml
+++ b/controllers/consoleplugin/config/static-frontend-config.yaml
@@ -322,46 +322,55 @@ columns:
   - id: K8S_Name
     name: Names
     calculated: '[SrcK8S_Name,DstK8S_Name]'
+    filter: name
     default: false
     width: 15
   - id: K8S_Type
     name: Kinds
     calculated: '[SrcK8S_Type,DstK8S_Type]'
+    filter: kind
     default: false
     width: 10
   - id: K8S_OwnerName
     name: Owners
     calculated: '[SrcK8S_OwnerName,DstK8S_OwnerName]'
+    filter: owner_name
     default: false
     width: 15
   - id: K8S_OwnerType
     name: Owner Kinds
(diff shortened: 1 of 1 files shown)


jpinsonneau:
Rebased without changes

internal/controller/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -673,6 +685,22 @@ filters:
     name: Cluster
     component: autocomplete
     hint: Specify a cluster ID or name.
+  - id: namespace
+    name: Namespace
+    component: autocomplete
+    autoCompleteAddsQuotes: true
+    category: targeteable
```
jotak: what exactly does it mean to be targeteable in this context?
Also, searching the web, it seems we should remove the "e": "targetable".

jotak: I guess I get what it means, it's a "common" field that can also be specialized into "src/dst" variants, right ?

jotak: Thinking out loud - not for this PRs, but I think we could eventually get rid of all the Src/Dst variants in this config file, keep just the targetable, and "expand" it after loading from the console side ? That would shrink this file size and also avoid some repeated content.

jpinsonneau: > I guess I get what it means, it's a "common" field that can also be specialized into "src/dst" variants, right ?

yeah it can be source, dest, or both

jotak: just to not forget my earlier comment: can you remove the "e"? :-)

jotak: actually perhaps rename category as "endpoint" so it matches the terminology that we agreed on in the UI

PR #1712, opened 2025-07-03
[Helm/upstream] helm chart improvements

jotak:
Based on https://github.com/netobserv/network-observability-operator/pull/1433, with some changes:

- Keep dynamic `{{ .Release.Namespace }}`
- Remove conversion webhook
- Integrate in current build workflow (`make helm-update`)
- Configure loki & prom subcharts
- Some renaming

Next I'll try to add the dependency to cert-manager


jotak:
fyi @Helion55 : I kept your first 3 commits on #1433 but not the others because they were heading to a wrong direction imho, with the hardcoded namespace. I've found a workaround by removing the conversion webhook, as it's not really needed.

Helion55:
Thanks @jotak, if we can proceed without conversion webhook, then its fine. I was proposing the hardcoded namespace only because of this. Any suggestions for me to work on further?

Helion55:
Hello @jotak any plans for merging this commit : https://github.com/netobserv/network-observability-operator/pull/1433/commits/3b711f9e39010c8d82d11ebb093a85f8f69e03cb ?

jotak:
Hi @Helion55,
On https://github.com/netobserv/network-observability-operator/commit/3b711f9e39010c8d82d11ebb093a85f8f69e03cb , I didn't keep it because this isn't the way we're versioning (it would change the version based on the branch name), and this would add commits from github CI on behalf of myself, it's not something we usually do. We don't create versions and releases automatically, it's a more controlled process that also involves quality testing.

Currently, the version in the helm charts are updated when running `make helm-update`, see [here](https://github.com/netobserv/network-observability-operator/pull/1712/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R539-R545). There is also a CI check that ensures there's no discrepancy between the helm chart version and the version for OLM delivery: [here](https://github.com/netobserv/network-observability-operator/blob/main/.github/workflows/pull_request.yml#L64-L65). I don't think more is needed at the moment.

Helion55:
Okay, got it. Yes that is very better way of doing version updataion. Thanks @jotak.

helm/templates/deployment.yaml
```diff
@@ -35,17 +35,17 @@ spec:
...
             - name: RELATED_IMAGE_FLOWLOGS_PIPELINE
-              value: quay.io/netobserv/flowlogs-pipeline:v1.9.0-community
+              value: '{{ .Values.flowlogsPipeline.image }}:{{ .Values.flowlogsPipeline.version }}'
             - name: RELATED_IMAGE_CONSOLE_PLUGIN
-              value: '{{ if .Values.standaloneConsole.enable }}quay.io/netobserv/network-observability-standalone-frontend:v1.9.0-community{{ else }}quay.io/netobserv/network-observability-console-plugin:v1.9.0-community{{ end }}'
+              value: '{{ if .Values.standaloneConsole.enable }}{{ .Values.standaloneConsole.image }}:{{ .Values.standaloneConsole.version }}{{ else }}{{ .Values.consolePlugin.image }}:{{ .Values.consolePlugin.version }}{{ end }}'
             - name: RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT
-              value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.8.2-community
+              value: '{{ if .Values.standaloneConsole.enable }}{{ .Values.standaloneConsole.image }}:{{ .Values.standaloneConsole.version }}{{ else }}{{ .Values.consolePlugin.image }}:{{ .Values.consolePlugin.version }}{{ end }}'
```
jpinsonneau: Is it on purepose to use the same image / version for the compat console plugin and the current one ?

If so, maybe we could simply make `RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT` env optionnal ?

jotak: yeah, being a bit lazy here, the few people (if any) who'd need compat could just switch manually to the correct image; let me see to make it optional then

jotak: done [bfe8bb7](https://github.com/netobserv/network-observability-operator/pull/1712/commits/bfe8bb71c7a4368163c07d7f27f489b4dd5e90dd)

jpinsonneau (approved):

PR #1731, opened 2025-07-11
NETOBSERV-2294: Adding egress support to network policy

OlivierCazade:
## Description

Adding egress support to network policy

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/networkpolicy/np_objects.go
+++ b/internal/controller/networkpolicy/np_objects.go
@@ -45,21 +45,48 @@ func buildMainNetworkPolicy(desired *flowslatest.FlowCollector, mgr *manager.Man
 						// Setting empty namespace selector will authorize every pod from the same namespace
 						PodSelector: &metav1.LabelSelector{},
 					},
-					// Allow traffic from the eBPF agents
-					peerInNamespace(privNs),
+				},
+			}},
+			Egress: []networkingv1.NetworkPolicyEgressRule{{
+				To: []networkingv1.NetworkPolicyPeer{
+					{
+						// Setting empty namespace selector will authorize every pod from the same namespace
+						PodSelector: &metav1.LabelSelector{},
+					},
 				},
 			}},
 			PolicyTypes: []networkingv1.PolicyType{
 				networkingv1.PolicyTypeIngress,
+				networkingv1.PolicyTypeEgress,
 			},
 		},
 	}
+	// Allow traffic from the eBPF agents
+	np.Spec.Ingress = append(np.Spec.Ingress, networkingv1.NetworkPolicyIngressRule{
+		From: []networkingv1.NetworkPolicyPeer{
+			peerInNamespace(privNs),
+		},
+	})
+	np.Spec.Egress = append(np.Spec.Egress, networkingv1.NetworkPolicyEgressRule{
+		To: []networkingv1.NetworkPolicyPeer{
+			peerInNamespace(privNs),
+		},
+	})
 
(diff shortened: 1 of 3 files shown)


internal/controller/networkpolicy/np_objects.go
```diff
@@ -95,6 +133,53 @@ func buildMainNetworkPolicy(desired *flowslatest.FlowCollector, mgr *manager.Man
...
+			To: []networkingv1.NetworkPolicyPeer{
+				peerInNamespace(constants.DNSNamespace),
+			},
+		})
+		// Allow egress for FLP when some exporters are configured
+		if len(desired.Spec.Exporters) > 0 {
+			np.Spec.Egress = append(np.Spec.Egress, networkingv1.NetworkPolicyEgressRule{
+				To: []networkingv1.NetworkPolicyPeer{
+					{
+						NamespaceSelector: &metav1.LabelSelector{MatchLabels: map[string]string{
```
jotak: I'm not sure to understand this rule, it says traffic allowed to namespace selector "app": "flowlogs-pipeline" ? Which namespace is that? Is it in order to allow users to configure namespaces where they deployed their custom exporter?

OlivierCazade: Thanks for catching this, I did some cleaning after testing and incorectly changed this...

The goal is to allow every egress for FLP only if an exporter is configured.

OlivierCazade:
About following up task, we have the [NETOBSERV-2293](https://issues.redhat.com//browse/NETOBSERV-2293) that is following.

The plan for the "netobserv-operator-namespace" is to have a static np in the bundle targeting only the operator.

internal/controller/networkpolicy/np_objects.go
jotak: ah, I see ... else, there's also the AdditionalNamespace that can be used for that purpose, no?

OlivierCazade: Some exporters may target something outside of the Openshift cluster. We have two possibilities here :
- allow every egress traffic for FLP if an exporter is configured
- do nothing and document how to configure an aditional NP when also configuring an exporter

I was going for the first option but I am also fine with the second

jotak: If there's something outside of the cluster, they can also create a new netpol just for that, on top of the one that we create. It's cumulative. So, that would work and IMO is better because it's not opening the doors too wide

jotak (approved):
/lgtm

Amoghrd:
@jotak @OlivierCazade Tested and it works as expected. See new egress policy deployed. 
One comment: Adding to the additionalNamespaces list(for ex `openshift-ingress`) only updates the networkPolicy for netobserv NS but is not reflected in netobserv-privileged NS. @OlivierCazade mentioned it was case previously when ingress policy was added in previous release but I just wanted to get a confirmation if that still holds true

jotak:
> One comment: Adding to the additionalNamespaces list(for ex `openshift-ingress`) only updates the networkPolicy for netobserv NS but is not reflected in netobserv-privileged NS. @OlivierCazade mentioned it was case previously when ingress policy was added in previous release but I just wanted to get a confirmation if that still holds true

Yes, the rationale was that additional namespaces would in general be useful when configuring some specific exporter, which require communication with FLP and not the agents.
If users need more policy fine tuning they still can roll out their own netpols

Amoghrd:
Sounds good, thanks for confirming.
/label qe-approved

PR #1744, opened 2025-07-17
NETOBSERV-2333: switch to tc hook on ocp4.15

jotak:
## Description

ocp 4.15 uses rhel 9.2, same as 4.14, so must be switched to TC hook as well.

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/ebpf/agent_controller.go
+++ b/internal/controller/ebpf/agent_controller.go
@@ -761,7 +761,7 @@ func getEnvConfig(coll *flowslatest.FlowCollector, cinfo *cluster.Info) []corev1
 	})
 
 	defaultAttach := "tcx"
-	if old, _ := cinfo.IsOpenShiftVersionLessThan("4.15.0"); old {
+	if old, _ := cinfo.IsOpenShiftVersionLessThan("4.16.0"); old {
 		defaultAttach = "tc"
 	}
 


jpinsonneau (approved):

internal/controller/ebpf/agent_controller.go
```diff
@@ -761,7 +761,7 @@ func getEnvConfig(coll *flowslatest.FlowCollector, cinfo *cluster.Info) []corev1
 	})
 
 	defaultAttach := "tcx"
-	if old, _ := cinfo.IsOpenShiftVersionLessThan("4.15.0"); old {
+	if old, _ := cinfo.IsOpenShiftVersionLessThan("4.16.0"); old {
```
memodi, line 764: I am curious how did we use to fallback on `tc` hook prior to https://github.com/netobserv/network-observability-operator/pull/1670 for older versions?  

I am not sure if there's a way to check a RHEL kernel version instead which would make it more explicit.

jotak: The previous "fallback to tc" logic was removed (*) because it introduced discrepancies when the tcx attachment failed not because it wasn't supported, but for any other reason, it ended up having a mix of tc and tcx hooks used which doesn't make sense. One of the leaks that we had was related to using mixed attachment methods.
Either the host supports TCX, and we should use it, or it does not, and we don't try to use it: it's simpler and safer (as long as correctly configured).

(*) => It's actually still possible to use the mixed mode with `TC_ATTACH_MODE: any`

On checking kernel version: that wouldn't be accurate because RHEL sometimes backports feature on the linux kernel, so what works on rhel may not work on other distro.

jotak: (we could also try to be smart and run a one-time check at startup to see if TCX is supported; my fear is just that in case it fails for any reason we're no further ahead ; having things hard-written in config makes everything more deterministic.)

memodi: thanks, I agree some explicit check is better.

PR #1746, opened 2025-07-17
Remove now unused 'replaces' csv field, and re-use BUILDVERSION when …

jotak:
…building downstream bundle


jotak:
I need to merge this before preparing the next versions

PR #1785, opened 2025-07-24
NETOBSERV-1742: use same metrics set with and without loki

jotak:
DNM: this is only for performance testing at this point


jotak:
I've tested on my cluster with load injected between 250 pods, sampling 1, all features enabled.
The metrics cardinality increases from roughly 15K to 30K
Capture d’écran du 2025-09-04 16-44-01

So I'm in favor to abandon this change

PR #1789, opened 2025-07-25
NETOBSERV-2225 - Deploy static plugin at operator startup - fix

jpinsonneau:
## Description

Fix namespace reconciliation when create by the user.
Since https://github.com/netobserv/network-observability-operator/pull/1345 `openshift.io/cluster-monitoring: "true"` is missing when the user created the namespace by himself.

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/monitoring/monitoring_controller.go
+++ b/internal/controller/monitoring/monitoring_controller.go
@@ -105,7 +105,7 @@ func (r *Reconciler) reconcile(ctx context.Context, clh *helper.Client, desired 
 		if err != nil {
 			return err
 		}
-	} else if helper.IsManaged(nsExist) && !helper.IsSubSet(nsExist.ObjectMeta.Labels, desiredNs.ObjectMeta.Labels) {
+	} else if !helper.SkipOwnership(nsExist) && !helper.IsSubSet(nsExist.ObjectMeta.Labels, desiredNs.ObjectMeta.Labels) {
 		err = r.Update(ctx, desiredNs)
 		if err != nil {
 			return err
--- a/internal/pkg/helper/flowcollector.go
+++ b/internal/pkg/helper/flowcollector.go
@@ -218,6 +218,13 @@ func IsManaged(obj client.Object) bool {
 	return labels[netobservManagedLabel] == "true"
 }
 
+// special case where ownership is ignored if netobserv-managed label is explicitly set to false
+// this is used to allow users to create namespaces with custom labels and annotations prior to the reconciliation
+func SkipOwnership(obj client.Object) bool {
+	labels := obj.GetLabels()
+	return labels != nil && labels[netobservManagedLabel] == "false"
+}
+
 func IsOwned(obj client.Object) bool {
 	// ownership is forced if netobserv-managed label is explicitly set to true
 	if IsManaged(obj) {


jotak (approved):
thanks!
/lgtm

memodi:
this does not work, when I created namespace with label `openshift.io/cluster-monitoring: "true"`


cat <<EOF | oc apply -f -
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    openshift.io/cluster-monitoring: "true"
  name: openshift-netobserv-operator
EOF

after installing flowcollector `openshift.io/cluster-monitoring: "true"` is removed:

```
$ oc get ns/netobserv -o yaml
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/sa.scc.mcs: s0:c25,c20
    openshift.io/sa.scc.supplemental-groups: 1000640000/10000
    openshift.io/sa.scc.uid-range: 1000640000/10000
    security.openshift.io/MinimallySufficientPodSecurityStandard: privileged
  creationTimestamp: "2025-07-25T19:47:53Z"
  labels:
    kubernetes.io/metadata.name: netobserv
    netobserv-managed: "true"
    pod-security.kubernetes.io/audit: privileged
    pod-security.kubernetes.io/audit-version: latest
    pod-security.kubernetes.io/warn: privileged
    pod-security.kubernetes.io/warn-version: latest
  name: netobserv
  resourceVersion: "34102"
  uid: f6987279-9a04-40a4-afff-55e05f04a27b
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

/needs-changes

jotak:
@memodi I guess you didn't set the downstream deployment env set to true, did you?
I can reproduce what you say if I don't set it, but not if I set it. This works as expected when I try.

jpinsonneau:
@memodi @jotak what about merging previous labels and annotations in any case ?

https://github.com/netobserv/network-observability-operator/pull/1789/commits/a6d2a2418a96123455f3114453ffa629925a082f

That's a new behavior but I feel it better match the needs here

memodi:
> @memodi I guess you didn't set the downstream deployment env set to true, did you?

@jotak you're right, my script needed to be updated for DOWNSTREAM_VAR index change.

memodi:
@jpinsonneau - sorry, I missed with previous image due to [this reason](https://github.com/netobserv/network-observability-operator/pull/1789#issuecomment-3130002362), I am able to see correct behavior even with [previous image](https://github.com/netobserv/network-observability-operator/pull/1789#issuecomment-3120121620) . 

Sorry for the false alarm, I guess we could revert [a6d2a24](https://github.com/netobserv/network-observability-operator/pull/1789/commits/a6d2a2418a96123455f3114453ffa629925a082f) ?

jpinsonneau:
> @jpinsonneau - sorry, I missed with previous image due to [this reason](https://github.com/netobserv/network-observability-operator/pull/1789#issuecomment-3130002362), I am able to see correct behavior even with [previous image](https://github.com/netobserv/network-observability-operator/pull/1789#issuecomment-3120121620) .
> 
> Sorry for the false alarm, I guess we could revert [a6d2a24](https://github.com/netobserv/network-observability-operator/pull/1789/commits/a6d2a2418a96123455f3114453ffa629925a082f) ?

I'll let @jotak decide on that once since it's a new behavior which make sense to me.

jotak:
This behaviour (keeping labels) makes sense to me, just trying to catch any potential side-effect...
For instance if there is an old label/annotation that we **want** to remove, it's not going to be removed. But, does that happen? I just see the corner case of having the openshift-monitoring annotation already set up, and we switch the downstream-deployment flag to false, which should make the annotation disappear, but it won't work here I guess. I say it's a corner case because normally we don't play with the downstream-deployment flag after the fact, it's either set or unset from the start, especially we hardly never want to _unset_ it... wdyt?

jotak:
but honestly I though we were already keeping previous labels/annotations ...

jotak:
or we should perhaps keep a hard-coded list of labels/annotations explicitly managed by us that we don't want to preserve from old revisions?

jpinsonneau:
> but honestly I though we were already keeping previous labels/annotations ...

We kind of do it as we don't update when not necessary but currently we override everything if it needs changes.

> or we should perhaps keep a hard-coded list of labels/annotations explicitly managed by us that we don't want to preserve from old revisions?

I don't want overengineering here. I'd prefer to keep the previous approach if you feel this is incomplete as is

jotak:
ok then I'd say let's revert to keep previous behaviour, and we can think about a more complete fix if someone complains at some point

jpinsonneau:
Removed https://github.com/netobserv/network-observability-operator/commit/a6d2a2418a96123455f3114453ffa629925a082f commit

memodi:
/label qe-approved

thanks, let's merge this asap so perf runs can get back on track.

PR #1790, opened 2025-07-25
NETOBSERV-2326: new alerting API in FlowCollector

jotak:
## Description

- new API in FlowCollector spec.processor.metrics.alertGroups
  - validation hook: validate alerts wrt enabled features, enabled metrics, etc.
  - in order to leverage FlowCollector helper functions from the validation hooks, I had to move many functions from internal/pkg/helper to api/flowcollector (without it, there would be cyclic dependencies)
  - some code is also moved from internal/pkg/metrics (default include lists...) to api/flowcollector, so that validation webhook can leverage them
- move existing PrometheusRule creation funcs to internal/pkg/metrics
- create conversion functions to convert the new API into PrometheusRules resources
- initial alertGroup: TooManyDrops, that works with PacketDrops feature. 3 alerts are enabled by default: From+To Namespace, From Node and To Node.

In order to have alerts working well by default, some default metrics are added: namespace_ingress_packets_total, node_drop_packets_total, node_ingress_packets_total

## Dependencies

- https://github.com/netobserv/network-observability-console-plugin/pull/983 (not required for having the alerts, but it provides the new health page that is built on top of that)

## Checklist

### DIFF
--- a/internal/pkg/test/envtest.go
+++ b/internal/pkg/test/envtest.go
@@ -3,6 +3,7 @@ package test
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -50,7 +51,13 @@ const (
 	Interval = 1 * time.Second
 )
 
-func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, basePath string) (context.Context, client.Client, *envtest.Environment, context.CancelFunc) {
+type SuiteContext struct {
+	testEnv    *envtest.Environment
+	cancel     context.CancelFunc
+	kubeConfig string
+}
+
+func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, basePath string) (context.Context, client.Client, *SuiteContext) {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 	ctx, cancel := context.WithCancel(context.TODO())
 
@@ -82,6 +89,9 @@ func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, baseP
 	cfg, err := testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
+
+	kubeConfig, err := writeKubeConfig(testEnv)
+	Expect(err).NotTo(HaveOccurred())
 
 	err = flowsv1beta2.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
@@ -134,7 +144,7 @@ func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, baseP
 	Expect(err).NotTo(HaveOccurred())
 
 	k8sManager, err := manager.NewManager(
-		context.Background(),
+		ctx,
 		cfg,
 		&manager.Config{
 			EBPFAgentImage:        "registry-proxy.engineering.redhat.com/rh-osbs/network-observability-ebpf-agent@sha256:6481481ba23375107233f8d0a4f839436e34e50c2ec550ead0a16c361ae6654e",
@@ -164,13 +174,36 @@ func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, baseP
 		Expect(err).ToNot(HaveOccurred(), "failed to run manager")
 	}()
 
-	return ctx, k8sClient, testEnv, cancel
+	return ctx, k8sClient, &SuiteContext{
+		testEnv:    testEnv,
+		cancel:     cancel,
+		kubeConfig: kubeConfig,
+	}
 }
 
-func TeardownEnvTest(testEnv *envtest.Environment, cancel context.CancelFunc) {
-	cancel()
+func writeKubeConfig(testEnv *envtest.Environment) (string, error) {
+	f, err := os.CreateTemp("", "testenv-kubeconfig-")
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	if _, err := f.Write(testEnv.KubeConfig); err != nil {
+		return f.Name(), err
+	}
+
+	logf.Log.Info("To debug with kubectl, run:")
+	logf.Log.Info("export KUBECONFIG=" + f.Name())
+	return f.Name(), nil
+}
+
+func TeardownEnvTest(suiteContext *SuiteContext) {
+	if suiteContext.kubeConfig != "" {
+		defer os.Remove(suiteContext.kubeConfig)
+	}
 	By("tearing down the test environment")
-	err := testEnv.Stop()
+	suiteContext.cancel()
+	err := suiteContext.testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
 }
 
--- a/internal/controller/flp/suite_test.go
+++ b/internal/controller/flp/suite_test.go
@@ -8,22 +8,16 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
 
 	"github.com/netobserv/network-observability-operator/internal/pkg/manager"
 	"github.com/netobserv/network-observability-operator/internal/pkg/test"
 )
 
-// These tests use Ginkgo (BDD-style Go testing framework). Refer to
-// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
-
-var namespacesToPrepare = []string{"main-namespace", "other-namespace"}
-
 var (
-	ctx       context.Context
-	k8sClient client.Client
-	testEnv   *envtest.Environment
-	cancel    context.CancelFunc
+	namespacesToPrepare = []string{"main-namespace", "other-namespace"}
+	ctx                 context.Context
+	k8sClient           client.Client
+	suiteContext        *test.SuiteContext
 )
 
 func TestAPIs(t *testing.T) {
@@ -41,9 +35,9 @@ var _ = Describe("FLP Controller", Ordered, Serial, func() {
 })
 
 var _ = BeforeSuite(func() {
-	ctx, k8sClient, testEnv, cancel = test.PrepareEnvTest([]manager.Registerer{Start}, namespacesToPrepare, "..")
+	ctx, k8sClient, suiteContext = test.PrepareEnvTest([]manager.Registerer{Start}, namespacesToPrepare, "..")
 })
 
 var _ = AfterSuite(func() {
-	test.TeardownEnvTest(testEnv, cancel)
+	test.TeardownEnvTest(suiteContext)
 })
--- a/internal/controller/monitoring/suite_test.go
+++ b/internal/controller/monitoring/suite_test.go
@@ -8,22 +8,16 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
 
 	"github.com/netobserv/network-observability-operator/internal/pkg/manager"
 	"github.com/netobserv/network-observability-operator/internal/pkg/test"
 )
 
-// These tests use Ginkgo (BDD-style Go testing framework). Refer to
-// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
-
-var namespacesToPrepare = []string{"openshift-config-managed", "main-namespace"}
-
 var (
-	ctx       context.Context
-	k8sClient client.Client
-	testEnv   *envtest.Environment
-	cancel    context.CancelFunc
+	namespacesToPrepare = []string{"openshift-config-managed", "main-namespace"}
+	ctx                 context.Context
+	k8sClient           client.Client
+	suiteContext        *test.SuiteContext
 )
 
 func TestAPIs(t *testing.T) {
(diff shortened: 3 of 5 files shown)


api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -591,12 +579,19 @@ type FLPMetrics struct {
 	// +optional
 	IncludeList *[]FLPMetric `json:"includeList,omitempty"`
 
-	// `disableAlerts` is a list of alerts that should be disabled.
+	// `disableAlerts` is a list of alert groups that should be disabled from the default set of alerts.
 	// Possible values are:<br>
 	// `NetObservNoFlows`, which is triggered when no flows are being observed for a certain period.<br>
 	// `NetObservLokiError`, which is triggered when flows are being dropped due to Loki errors.<br>
+	// TODO: add full list of default alerts
```
jpinsonneau: Is it something you plan to do in that PR ?

api/flowcollector/v1beta2/flowcollector_validation_webhook.go
```diff
@@ -274,6 +274,7 @@ func validatePortString(s string) (uint16, error) {
 	return uint16(p), nil
 }
 
+// nolint:cyclop
```
jpinsonneau: I guess we could divide this function in two or three subfunctions for:
- groups
- threashold
- warnings

config/samples/flows_v1beta2_flowcollector.yaml
```diff
@@ -94,11 +94,33 @@ spec:
...
+      #       warning: "10"
+      #       info: "5"
+      #     grouping: PerNode
+      #     groupingDirection: ByDestination
+      #   - thresholds:
+      #       critical: "15"
+      #       warning: "10"
+      #       info: "5"
+      #     grouping: PerNamespace
+      #     groupingDirection: BySource
```
jpinsonneau: I guess we should provide some defaults here

@stleerh maybe you have some ideas ?

jpinsonneau: I didn't got at first look that those are customs alerts :upside_down_face: 
Maybe we should rename `alertGroups` to make it explicit

api/flowcollector/v1beta2/flowcollector_validation_webhook.go
jotak: I can break it down in more chunks but I don't find that it helps readability, quite the opposite.. this function validates different things in a quite straightforward way, I find it easier to read when reading things sequentially rather than jumping in different places to follow the logic... idk that's kind of subjective I guess

jotak: in fact I can refactor the whole thing here to remove some of the boilerplate, let me try that...

jotak: done in aeb58716dcda357b0181395f32f4760b92a65393

api/flowcollector/v1beta2/flowcollector_types.go
jotak: done in aeb58716dcda357b0181395f32f4760b92a65393

api/flowcollector/v1beta2/flowcollector_validation_webhook.go
jpinsonneau: awesome ! thanks @jotak !

config/samples/flows_v1beta2_flowcollector.yaml
jotak: I don't know, it's "semi-custom alerts": it's tweaking some parameters of the alerts that netobserv can generate, but at the end of the day it's still quite opinionated (netobserv controls the underlying promql). Users who would like truly custom alerts can create their own from scratch.

Also I called it "alertGroups" because inside that, there's again a list of alerts. I don't want to have `alerts.alerts.something` in the spec path.

jotak: I can see this, as an alternative:
```
      alerts:
      - template: TooManyDrops
        groupings:
        - group: global
          thresholds:
            critical: "15"
            warning: "10"
            info: "5"
        - group: PerNode
          direction: ByDestination
          thresholds:
            critical: "15"
            warning: "10"
            info: "5"
```

would that sound better?

memodi: IMO alternate seems better than the original one, especially using `template` as name instead of simply generic `name` .
Few other suggestions:
-  set default thresholds for each template and let user override.
-  have `groupBy` instead of `group` defaults to global if not specified.
-  Having `direction` in this config feels like mixing alerting and flows concept  together. Does this field in combination with group: PerNode,  translate to `DstK8S_Name = Node` condition in Promql?

also about:
> We need to take a decision:
>
> Either we had those metrics by default
> Or we do not enable the drops alerts by default
> Or we only enable those metrics by default when the drops feature is enabled

I vote for # 3 to enable those metrics and alerts by default when drop features is enabled.

jpinsonneau: I'm aligned with @memodi on this :rocket: 

Everything could be optionnal appart from the `template` field to list alerts you want to use. The rest is about overriding defaults.

jotak: this is already like you are suggesting, there are default alerts provided, that users can override, and global is already the default grouping :-)
The sample yaml here is an override.
On fields naming, I'm also thinking about changing "groupings" for "variants" (these are variations of the same template).

> Having direction in this config feels like mixing alerting and flows concept together. Does this field in combination with group: PerNode, translate to DstK8S_Name = Node condition in Promql?

Yes, metrics/alerts are still based on flows and the concept of direction doesn't disappear. Yes also to your second point, or to be very accurate, direction=ByDestination,group=PerNode translates into `by (DstK8S_HostName)` in promql.

jotak: Actually maybe I misunderstood your remark, `direction` here is indeed not the same as the Flow direction that we have in flows. Flow direction is whether the flow was observed from source or from destination. It's not what it is here. So maybe `direction` is a poor naming choice here, indeed, because it's getting overloaded...

internal/pkg/metrics/alerts/alerts_test.go
```diff
@@ -0,0 +1,208 @@
...
+			},
+		},
+	}
+	rules := BuildRules(context.Background(), &fc)
+	assert.Equal(t, []string{
+		"TooManyKernelDrops_WNamespace",
+		"TooManyKernelDrops_INamespace",
+		"TooManyKernelDrops_WNode",
+		"TooManyKernelDrops_INode",
+		"TooManyDeviceDrops_WNode",
```
memodi: what does W and I mean in W/INamespace/Node ?

jotak: C/W/I = severity critical/warning/info
Workload/Namespace/Node is the grouping

That's because each alert template can be declined in variants with different severity and grouping. At the end of the day, prometheus needs a unique alert name. I try to make it not too visible to the user, although if they navigate into the alerts page they will see these names.

jotak:
(rebased, and renamed alerts)

internal/pkg/metrics/alerts/alerts_test.go
memodi: thanks. I wish the Alert naming could have been more dynamic to denote the different level of grouping. May be we could have suffixes like: `PacketDropsByKernel_PerNSWarn` `PacketDropsByKernel_PerNodeCritical` .  Not trying to nit-pick much on naming, but trying to have name as descriptive and simple as possible.

Also, Do you think an info level is needed?  I am thinking, could we use the `Pending` state of the "Warning/Critical" Alert to show as "info" on the health page? Trying to think if we can simplify further.

jotak: I'm doing the suggested changes on names

However for simplification, I don't think we should get away from the model defined by monitoring. I think that could bring more confusion if our definition of "info" differs from the usual definition used in other places ?

memodi:
Hi @jotak - any reason for Src/Dst alerts split? I wasn't sure if that was something planned or something you noticed caused you to split them, just curious.

While testing DNSAlerts, I noticed on Global page it shows one of the alerts as empty dst namespace:

image

When click on "Navigate to alert details", I get "No Alert found"

jotak:
@memodi I preferred to split them because it drastically reduce the complexity of the promql (you can see a diff [here](https://github.com/netobserv/network-observability-operator/pull/1790/commits/7fe774e8a2c0cbabc94eb3c59f05a79e15c25599#diff-1d1388fb006c52d96a50595adaf0e71453ce66f40db292f7e45ad0be5f80296bR260-R268))

but that said, it's not 100% expected that there is an alert with empty namespace in the global page ... Or I should say, I haven't really thought about where those kind of alerts should go

> When click on "Navigate to alert details", I get "No Alert found"

or isn't it because the alert has disappeared in the meantime? I've seen it happening quite a few times

memodi:
> or isn't it because the alert has disappeared in the meantime? I've seen it happening quite a few times

I don't think so, I've almost instantly clicked on them and it gives me 404.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -1070,7 +1072,7 @@ spec:
                 - name: RELATED_IMAGE_CONSOLE_PLUGIN
                   value: quay.io/netobserv/network-observability-console-plugin:v1.9.2-community
                 - name: RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT
-                  value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.8.2-community
+                  value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.9.1-community
```
jpinsonneau: Why is there `network-observability-console-plugin:v1.9.2-community` alongside `network-observability-console-plugin-pf4:v1.9.1-community` here ? :thinking:

internal/pkg/metrics/alerts/alerts.go
```diff
@@ -0,0 +1,198 @@
...
+		legend = " [node={{ $labels.instance }}]"
+	case flowslatest.GroupByNamespace:
+		return nil, fmt.Errorf("PacketDropsByDevice alert does not support grouping per namespace")
+	case flowslatest.GroupByWorkload:
+		return nil, fmt.Errorf("PacketDropsByDevice alert does not support grouping per workload")
+	}
+
+	promql := percentagePromQL(
+		fmt.Sprintf("sum(rate(node_network_receive_drop_total[2m]))%s + sum(rate(node_network_transmit_drop_total[2m]))%s", byLabels, byLabels),
+		fmt.Sprintf("sum(rate(node_network_receive_packets_total[2m]))%s + sum(rate(node_network_transmit_packets_total[2m]))%s", byLabels, byLabels),
```
jpinsonneau, line 84: Should we expose the range (2m) here ? I guess it can impact a lot on the thresholds you are setting in some cases.

internal/pkg/test/envtest.go
```diff
@@ -170,8 +170,8 @@ func PrepareEnvTest(controllers []manager.Registerer, namespaces []string, baseP
 func TeardownEnvTest(testEnv *envtest.Environment, cancel context.CancelFunc) {
 	cancel()
 	By("tearing down the test environment")
-	err := testEnv.Stop()
-	Expect(err).NotTo(HaveOccurred())
+	_ = testEnv.Stop()
+	// Expect(err).NotTo(HaveOccurred())
```
jpinsonneau: Could you please explain why skipping this ?

jpinsonneau (approved):
Code looks good, just small comments

internal/pkg/test/envtest.go
jotak: ah .. it's related to a flaky test, see also my failed fix attempt here https://github.com/netobserv/network-observability-operator/pull/1950  
basically that line sometimes fail since a couple of months, with the kube env never stopping (even if I increase the timeout), I haven't found the root cause so far and was quite annoyed ... but I'll remove the change, and will try harder .. later :-)

jpinsonneau: sure no worries. Just add a comment in the code pointing your attempt

internal/pkg/metrics/alerts/alerts.go
jotak: I didn't on purpose, with the idea to start simple and add things if we feel the need. I assume that this whole feature is for users who don't go too much into fine-tuning, because otherwise they can define alerts from scratch, as well.
So, the less configuration option there is, the easier it for users.
But there will be more testing done in the next release for GA, and maybe we can revisit that if we feel it's better to have

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jotak: yeah, some wrong things happening with the `make update-bundle` I think, I'll remove that

internal/pkg/test/envtest.go
jotak: reverted

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: yeah we should fix that in parallel

internal/pkg/metrics/alerts/alerts.go
jpinsonneau: I'm fine with that if we are feeling it's enough for now as we don't cover much cases yet

internal/pkg/test/envtest.go
jotak: Finally, I hope, problem solved: this was this single line here https://github.com/netobserv/network-observability-operator/pull/1790/commits/fd53c92706c4800f1274c4da639541f3c5896ff9#diff-68dd4181a92af8126232a8a9c284fa21cee59937a1253b9a16e2670df699a117R147

memodi:
@jotak - I am testing for multi-tenancy, looks like we'll need to add `monitoring.coreos.com` apiGroup to our metrics-reader clusterrole for non-admins to be able to fetch alerts. 

image

jotak:
@memodi good catch, let's add a story to the GA epic

=> https://issues.redhat.com/browse/NETOBSERV-2409

memodi:
I think I am good here, except about the alerts without namespaces for which I filed an issue which we can address as part of GA work: https://issues.redhat.com/browse/NETOBSERV-2411

/label qe-approved

PR #1801, opened 2025-07-28
NETOBSERV-2360 Loki dev mode

jpinsonneau:
## Description

Implement a new loki mode `Dev` that deploy the zero click loki.

Also took the opportunity to update the config to rely on latest release using tsdb.

TODOs:
- [x] move that new mode inside monolithic config
- [x] expose loki uri in CSV
  - that will allow us to change it downstream to `registry.redhat.io/openshift-logging/logging-loki-rhel9:v3.4.3` without effort
  - `config.file` is exposed as container ARG to support downstream image

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/loki/loki_reconciler.go
+++ b/internal/controller/loki/loki_reconciler.go
@@ -67,8 +67,19 @@ func (r *LReconciler) Reconcile(ctx context.Context, desired *flowslatest.FlowCo
 			return err
 		}
 	} else {
-		// delete any existing owned object
-		r.Managed.TryDeleteAll(ctx)
+		// delete any existing owned object (only if they were created by netobserv)
+		if err := r.DeleteIfOwned(ctx, r.configMap); err != nil {
+			return err
+		}
+		if err := r.DeleteIfOwned(ctx, r.deployment); err != nil {
+			return err
+		}
+		if err := r.DeleteIfOwned(ctx, r.pvc); err != nil {
+			return err
+		}
+		if err := r.DeleteIfOwned(ctx, r.service); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/internal/pkg/helper/client_helper.go
+++ b/internal/pkg/helper/client_helper.go
@@ -119,6 +119,29 @@ func (c *Client) UpdateIfOwned(ctx context.Context, old, obj client.Object) erro
 	return c.UpdateOwned(ctx, old, obj)
 }
 
+// DeleteIfOwned is an helper function that deletes an object only if it's currently owned and managed by the operator
+func (c *Client) DeleteIfOwned(ctx context.Context, obj client.Object) error {
+	log := log.FromContext(ctx)
+	kind := reflect.TypeOf(obj).String()
+
+	if obj == nil {
+		return nil
+	}
+
+	if !IsOwned(obj) {
+		log.Info("SKIP "+kind+" deletion since not owned", "Namespace", obj.GetNamespace(), "Name", obj.GetName())
+		return nil
+	}
+
+	log.Info("DELETING "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
+	err := c.Delete(ctx, obj)
+	if err != nil {
+		log.Error(err, "Failed to delete "+kind, "Namespace", obj.GetNamespace(), "Name", obj.GetName())
+		return err
+	}
+	return nil
+}
+
 func getFlowCollector(ctx context.Context, c client.Client) (*flowslatest.FlowCollector, error) {
 	log := log.FromContext(ctx)
 	desired := &flowslatest.FlowCollector{}


jpinsonneau:
Updated bundle

jpinsonneau:
Rebased without changes

memodi:
e2e-operator test is failing here with FLP not being able to come ready, does the test needs to be updated or something that needs be updated in PR?

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -862,6 +862,11 @@ type LokiMicroservicesParams struct {
 
 // LokiMonolithParams is the configuration for monolithic Loki (https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/#monolithic-mode)
 type LokiMonolithParams struct {
+	// Set `zeroClick` to `true` to automatically create Loki deployment, service and storage.
+	// This is useful for development pureposes. Do not use it in production.
+	//+kubebuilder:default:=true
```
jotak: I think it should be false by default
Also, I think we can use `bool` here (no pointer) as I don't think we'll ever want to change the default over time for existing users. (we use pointers mostly for settings where we want to keep the ability to change defaults later upon upgrades - or is there another reason for having a pointer?)

jotak: Maybe also rename "zeroClick" to something like "installDemoLoki" ? (something more explicit about what it does)

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -1243,6 +1240,8 @@ spec:
     name: console-plugin
   - image: quay.io/netobserv/network-observability-console-plugin:v1.10.0-community-pf4
     name: console-plugin-compat
+  - image: grafana/loki:3.5.0
```
jotak, line 1260: I'm still pretty sure this is going to fail konflux validation of the related images, as this isn't an approved registry

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -862,6 +862,11 @@ type LokiMicroservicesParams struct {
 
 // LokiMonolithParams is the configuration for monolithic Loki (https://grafana.com/docs/loki/latest/fundamentals/architecture/deployment-modes/#monolithic-mode)
 type LokiMonolithParams struct {
+	// Set `zeroClick` to `true` to automatically create Loki deployment, service and storage.
+	// This is useful for development pureposes. Do not use it in production.
```
jotak: nit / typo:

```suggestion
	// This is useful for development and demo purposes. Do not use it in production.
```

jotak:
IMO there should be a systematic warning triggered by the validation webhook, that this isn't supported. Or maybe, at least when there's the DOWNSTREAM_DEPLOYMENT flag (it's ok for upstream)

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: The corresponding image downstream should be https://catalog.redhat.com/en/software/containers/openshift-logging/logging-loki-rhel9/64479926a759a6e51ef79b12

I need to test it to see how it goes :thinking: 

Maybe we should make this configurable so we don't force a specific container image and default to the logging one if it works fine ?

jpinsonneau:
- Rebased
- Renamed zeroClick to demoLoki
- Disabled by default

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: By the way, all of our upstream content points to quay.io so I don't see the issue here.
If you don't want to point grafana/loki image, I can copy it under quay.io/netobserv but that should be the exact same image upstream.
Konflux should replace that image downstream (by the logging one or another we build ourself).

jpinsonneau: I confirm it works with image `registry.redhat.io/openshift-logging/logging-loki-rhel9:v3.5.7` :ok_hand:

jotak: > By the way, all of our upstream content points to quay.io so I don't see the issue here.
> If you don't want to point grafana/loki image, I can copy it under quay.io/netobserv but that should be the exact same image upstream.

Sorry I didn't mean that, for upstream it's ok to have the grafana image, my concern really is about how that would work downstream. See next point:

> Konflux should replace that image downstream (by the logging one or another we build ourself).

It's actually something done here: https://github.com/netobserv/network-observability-operator/blob/main/hack/patch_csv.py#L82-L90 So we need to update that file for Loki.

It's nice that `logging-loki-rhel9` works as a drop-in replacement, so we can reference it there, but then we need to see how that will work with konflux & renovate: if renovate is able to update it, we can use a pinned digest; if renovate cannot do it then I guess we should use a stream reference (3.5 currently) but that means having to regularly check for updates manually. (And I'm not sure if konflux will accept the non-digest version). Last option could be to disable the feature for downstream builds.

jotak: (I'm asking renovate folks in slack)

jotak (changes requested):
There's still the problem that I pointed in my previous comment: https://github.com/netobserv/network-observability-operator/pull/1801#discussion_r2634386020

Internal slack thread: https://redhat-internal.slack.com/archives/C06KTFLDS69/p1766137791942929 - there's a suggestion proposed to make it work downstream with renovate.

For upstream, I would just use the grafana image, and avoid maintaining our own.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: That should do it

https://github.com/netobserv/network-observability-operator/pull/1801/commits/0c5b8ce5cc8b2ef472313ac69701abaea4bd103e

jpinsonneau:
Addressed feedback and rebased

Amoghrd:
@jpinsonneau The flowcollector status is not coming up ready with errors
Screenshot 2026-01-12 at 1 34 13 PM
I have set DOWNSTREAM_DEPLOYMENT to true so the warning can be ignored

jotak (approved):

jpinsonneau:
> @jpinsonneau The flowcollector status is not coming up ready with errors Screenshot 2026-01-12 at 1 34 13 PM I have set DOWNSTREAM_DEPLOYMENT to true so the warning can be ignored

Good catch @Amoghrd ! PVC specs where modified and trying to reconcile. I fixed that by comparing only what's needed.
https://github.com/netobserv/network-observability-operator/pull/1801/commits/0a58b4414416abcd0fb36e6a2214f79ffa88fb40

Amoghrd:
@jpinsonneau 2 errors/nits:
- Default Flowcollector deployment is broken too with error
Screenshot 2026-01-13 at 1 12 13 PM

- Demo loki deployment section header should be caps like other fields
Screenshot 2026-01-13 at 1 14 03 PM

memodi:
any reason why metrics is not populated causing e2e-operator test to fail?

```
Metric netobserv_ingest_flows_processed not found after 180 secs!!!
```

Amoghrd:
My bad seems like the default flowcollector error for Prom alertManagerURL not being present is present on latest downstream version too. So might not be related to this PR

jpinsonneau:
> * Demo loki deployment section header should be caps like other fields
> 
> Screenshot 2026-01-13 at 1 14 03 PM

We'll need to decide either to expose it properly or hide it in the plugin. That's another PR to open in console plugin repo

> any reason why metrics is not populated causing e2e-operator test to fail?
> 
> ```
> Metric netobserv_ingest_flows_processed not found after 180 secs!!!
> ```

Is that only happening on that PR ? :thinking: I don't see why it would happen here.

I can rebase when main branch will be more stable no worries :wink:

memodi:
> Is that only happening on that PR ? 🤔 I don't see why it would happen here.

it looks like its consistently failing on this PR at that error. I looked at other open PRs and it passing there:
https://github.com/netobserv/network-observability-operator/pull/2327
https://github.com/netobserv/network-observability-operator/pull/2329

jpinsonneau:
> > Is that only happening on that PR ? 🤔 I don't see why it would happen here.
> 
> it looks like its consistently failing on this PR at that error. I looked at other open PRs and it passing there: #2327 #2329

Rebased without changes to see if it fixes the issue

memodi:
> Rebased without changes to see if it fixes the issue
 thanks, looks like it's still failing, FLP may not be becoming ready:

```
====> Waiting for flowlogs-pipeline daemonset to be created
NAME                READY   UP-TO-DATE   AVAILABLE   AGE
flowlogs-pipeline   0/3     3            0           14s
```

is it possible the `samples/flows_v1beta2_flowcollector.yaml` needs to be updated for changes in PR? That is primarily test is using https://github.com/openshift/release/blob/master/ci-operator/config/netobserv/network-observability-operator/netobserv-network-observability-operator-main.yaml#L95C88-L95C128

jpinsonneau:
> > Rebased without changes to see if it fixes the issue
> > thanks, looks like it's still failing, FLP may not be becoming ready:
> 
> ```
> ====> Waiting for flowlogs-pipeline daemonset to be created
> NAME                READY   UP-TO-DATE   AVAILABLE   AGE
> flowlogs-pipeline   0/3     3            0           14s
> ```
> 
> is it possible the `samples/flows_v1beta2_flowcollector.yaml` needs to be updated for changes in PR? That is primarily test is using https://github.com/openshift/release/blob/master/ci-operator/config/netobserv/network-observability-operator/netobserv-network-observability-operator-main.yaml#L95C88-L95C128

The loki dev mode is disabled in the sample so that right.

However, I found that the controller deletes a PVC !
```
2026-01-14T16:15:28.605Z	INFO	legacy.loki	DELETING *v1.PersistentVolumeClaim	{"Namespace": "netobserv", "Name": "loki-store"}
```

Which means the CI loki naming match the loki dev one. I was not expecting that since we only update resources having the `netobserv-managed: true` label. However, that never been applied to deletions (to pass the update scenario where this label was introduced).

I'm applying that check only for loki reconciler to avoid extra issues but we may think about extending it to all reconcilers in future.

https://github.com/netobserv/network-observability-operator/pull/1801/commits/7f0154f62ccb307edd9fb94a1d56f768ee860983

jpinsonneau:
@memodi I don't see the PVC being removed in the CI logs anymore and there is no error in FLP.

Is there a way to get FlowCollector status from the artifacts ?

memodi:
> @memodi I don't see the PVC being removed in the CI logs anymore and there is no error in FLP.

@jpinsonneau unfortunately, no. I tried to look into [must-gather](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/netobserv_network-observability-operator/1801/pull-ci-netobserv-network-observability-operator-main-e2e-operator/2011731992332210176/artifacts/e2e-operator/gather-must-gather/artifacts/) from the run and FLP pod logs, I couldn't find any smoking guns.

memodi:
@jpinsonneau it's because you have new env var in the bundle `RELATED_IMAGE_DEMO_LOKI` and in CI we're patching DOWNSTREAM_DEPLOYMENT based on the the index https://github.com/openshift/release/blob/master/ci-operator/config/netobserv/network-observability-operator/netobserv-network-observability-operator-main.yaml#L91C170-L91C173 , I'll update the CI to patch based on env var name instead.

memodi:
updating tests here https://github.com/openshift/release/pull/73590

jpinsonneau:
> @jpinsonneau it's because you have new env var in the bundle `RELATED_IMAGE_DEMO_LOKI` and in CI we're patching DOWNSTREAM_DEPLOYMENT based on the the index https://github.com/openshift/release/blob/master/ci-operator/config/netobserv/network-observability-operator/netobserv-network-observability-operator-main.yaml#L91C170-L91C173 , I'll update the CI to patch based on env var name instead.

Oh ok thanks for finding that. But why the controller logs seems fine then ? I would have expected errors there :thinking: 
I'm probably missing something !

memodi:
> Oh ok thanks for finding that. But why the controller logs seems fine then ? I would have expected errors there 🤔
> I'm probably missing something !

no, because it patched the var at index 4, it actually ended up patching RELATED_IMAGE_DEMO_LOKI env var:
```
    - name: RELATED_IMAGE_DEMO_LOKI
      value: "true"
    - name: DOWNSTREAM_DEPLOYMENT
      value: "false"
```

jpinsonneau:
Tests are now passing :partying_face:

Amoghrd:
@jpinsonneau Could you rebase the PR to get the alertManager fix?

jpinsonneau:
> @jpinsonneau Could you rebase the PR to get the alertManager fix?

Rebased without changes

jpinsonneau:
Thanks @Amoghrd !

PR #1846, opened 2025-08-07
NETOBSERV-2293 : Set default value for network policy to true

OlivierCazade:
## Description

Set the default value for network policy to true.

This story first intended to also add a network policy for the operator namespace but this is going to be done in a later task.

## Dependencies

n/a

## Checklist

  * [X] If so, make sure the JIRA epic is labeled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.

### DIFF
--- a/internal/pkg/helper/flowcollector.go
+++ b/internal/pkg/helper/flowcollector.go
@@ -14,6 +14,34 @@ const (
 	netobservManagedLabel = "netobserv-managed"
 )
 
+func GetSampling(spec *flowslatest.FlowCollectorSpec) int {
+	if spec.Agent.EBPF.Sampling == nil {
+		return 50
+	}
+	return int(*spec.Agent.EBPF.Sampling)
+}
+
+func UseKafka(spec *flowslatest.FlowCollectorSpec) bool {
+	return spec.DeploymentModel == flowslatest.DeploymentModelKafka
+}
+
+func DeployNetworkPolicy(spec *flowslatest.FlowCollectorSpec) bool {
+	return spec.NetworkPolicy.Enable != nil && *spec.NetworkPolicy.Enable
+}
+
+func HasKafkaExporter(spec *flowslatest.FlowCollectorSpec) bool {
+	for _, ex := range spec.Exporters {
+		if ex.Type == flowslatest.KafkaExporter {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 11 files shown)


internal/pkg/helper/flowcollector.go
```diff
@@ -27,6 +27,10 @@ func UseKafka(spec *flowslatest.FlowCollectorSpec) bool {
 	return spec.DeploymentModel == flowslatest.DeploymentModelKafka
 }
 
+func DeployNetworkPolicy(spec *flowslatest.FlowCollectorSpec) bool {
+	return spec.NetworkPolicy.Enable != nil && *spec.NetworkPolicy.Enable
```
OlivierCazade, line 29: Previous default value was nil, by still not deploying when the value is nil, we prevent breaking netobserv during update by deploying the NP in not ready namespace.

It is also possible to enforce NP deployment by changing the logic here.

OlivierCazade:
/hold

We need to merge this PR first to also have the NP enabled by default in the console form :
https://github.com/netobserv/network-observability-console-plugin/pull/956

memodi:
@OlivierCazade - I'll be picking up from Amogh here, if we run as default, this would mean we'd need to update all our automated tests with Loki to have LokiStack deployed in different Namespace correct?

memodi:
also, could you rebase?

jotak:
/lgtm
As discussed with the team, we must now recommend deploying loki in its own namespace

OlivierCazade:
I also changed the Network Policy additional namespace to also apply to the privileged namespace, this way we can use this field to allow communications when kafka is deployed in a different namespace.

I can also add a dedicated field but IMO it is better to reuse this field.

memodi:
/label qe-approved

thanks @OlivierCazade 

just the upgrade test is pending now.

memodi:
/hold
(for upgrade testing on pre-merge downstream bundle)

memodi:
/unhold

couldn't complete upgrade testing on pre-merge konflux bundle since upgrade path is not defined in it, only 1.10 entry exist on that bundle:
```
      version: 1.10.0
    entries:
    - name: network-observability-operator.v1.10.0
      version: 1.10.0
    name: operator-sdk-run-bundle
  defaultChannel: operator-sdk-run-bundle
```

will complete upgrade test post-merge as part of this task.

PR #1876, opened 2025-08-12
[upstream] Fix deployment on older openshift / pf4

jotak:


jotak:
FYI I hit this issue when deploying netobserv on ocp 4.14 from the makefile: it looks for an invalid quay image. There's nothing like "netobserv/network-observability-console-plugin-pf4" on quay, pf4 repo is more of a downstream thing, however we do have "main-pf4" tags.

OlivierCazade (approved):
LGTM

PR #1973, opened 2025-09-11
NETOBSERV-584: Support for writing logs to Loki (distributor) using gRPC

leandroberetta:
## Description

Support for writing logs to Loki (distributor) using gRPC.

Ref: https://issues.redhat.com/browse/NETOBSERV-584

## Dependencies

https://github.com/netobserv/loki-client-go/pull/3
https://github.com/netobserv/flowlogs-pipeline/pull/1086
https://github.com/netobserv/network-observability-console-plugin/pull/1021

## Checklist


leandroberetta:
This PR adds support for writing logs to Loki using gRPC. In terms of performance, results are pretty similar to the existing http implementation. 

I was investigating this and the current implementation uses http but messages are protocol buffers compressed (snappy), so it's already using less memory I guess. 

In other hands, gRPC establishes a connection and has mechanisms to reconnect, keep alive. That take more resources (according to what I read).

Probably we can do a better performance testing until we can decide.

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -883,6 +883,29 @@ const (
...
+	MaxSendMsgSize int `json:"maxSendMsgSize,omitempty"`
+
+	//+kubebuilder:default:="30s"
+	// `keepAlive` is the gRPC keep-alive interval.
+	KeepAlive *metav1.Duration `json:"keepAlive,omitempty"`
+
+	//+kubebuilder:default:="5s"
+	// `keepAliveTimeout` is the gRPC keep-alive timeout.
+	KeepAliveTimeout *metav1.Duration `json:"keepAliveTimeout,omitempty"`
+}
```
jpinsonneau: I'm not sure we need those exposed in the API.

On top of that, what's the difference between `maxSendMsgSize` and the existing `writeBatchSize` ?
Maybe we can reuse the same field here and put the other options as `Env map[string]string` in the `advanced` section

Here is an example for eBPF agent consuming envs: https://github.com/netobserv/network-observability-operator/blob/main/internal/controller/ebpf/agent_controller.go#L33-L75

jotak: Yes I think we can probably start with not exposing anything at all (not even the `clientType`) and have it under a env-based feature gate. Probably users want us to make the choice for them, whether to use it or not.
If after more tests we see that there are no clear winner, just pros and cons, then maybe we will expose it?

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -1072,7 +1085,7 @@ spec:
                 - name: RELATED_IMAGE_CONSOLE_PLUGIN
                   value: quay.io/netobserv/network-observability-console-plugin:v1.9.2-community
                 - name: RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT
-                  value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.8.2-community
+                  value: quay.io/netobserv/network-observability-console-plugin:v1.9.2-community
```
jotak: the change here, and below line 1233, should be reverted. Not the first time we notice this kind of change, that certainly came when you ran `make update-bundle`, something looks broken there, I need to investigate ...

config/manager/manager.yaml
```diff
@@ -39,7 +39,7 @@ spec:
           - name: RELATED_IMAGE_CONSOLE_PLUGIN
             value: quay.io/netobserv/network-observability-console-plugin:v1.9.2-community
           - name: RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT
-            value: quay.io/netobserv/network-observability-console-plugin-pf4:v1.8.2-community
+            value: quay.io/netobserv/network-observability-console-plugin:v1.9.2-community
```
jotak: same here, it must be reverted (and I must investigate why that happens)

api/flowcollector/v1beta2/flowcollector_types.go
leandroberetta: This suggestions are addressed. 

To enable this feature, an environent variable in the processor's advanced config can be set as follows:

```
    env:
      LOKI_USE_GRPC_CLIENT_PROTOCOL: 'true'
```

leandroberetta:
Coming back to this issue, suggestions are addressed. 

Also, I was checking (Again just in case) if for grpc we can avoid talking directly to the distributor component, because it's true that I guess we shouldn't be doing that. 

I investigate the following API and the mentions to gRPC (https://github.com/observatorium/api/blob/main/authorization/grpc.go) but I see that gRPC is supported for traces not for logging. For logging it uses the http cient.

I think we can leave this code for experimental use but given the results are similar (still being similar after some testing), I'm not sure. 

We have flp changes (and the loki-client-go lib too), probably we can merge those and wait to see if in the future there is the ability in LokiStack to write using gRPC using the gateway?.

What do you think?. Probably would be a good idea to do a better performance testing from the QE tests?.

jotak:
@leandroberetta do you confirm we can close this PR?

leandroberetta:
As commented here: https://issues.redhat.com/browse/NETOBSERV-584, we are closing this PR as it won't be used for now. FLP changes were merged for the future.

PR #1415, opened 2025-04-18
NETOBSERV-2054: Adding version label to upstream bundle

OlivierCazade:
## Description

Adding version label to upstream bundle

## Dependencies

n/a

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [ ] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [ ] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labeled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [ ] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [ ] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #2017, opened 2025-09-23
Fixing bundle.Dockerfile overwrite

OlivierCazade:
## Description

Operator-sdk now overwrite the bundle.Dockerfile, this PR disable it.

Metadata and manifest directory are not impacted since they are deleted before generating the bundle.

## Dependencies

n/a

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -411,7 +411,7 @@ bundle: bundle-prepare ## Generate final bundle files.
 		| $(SED) -e 's~:container-image:~$(IMAGE)~' \
 		| $(SED) -e "/':full-description:'/r tmp-desc" \
 		| $(SED) -e "s/':full-description:'/|\-/" \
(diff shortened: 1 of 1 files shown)


jotak:
/lgtm

The downside is to lose automated metadata updates, but I guess that's ok because they don't change so often, they're partly overwritten anyway (for downstream builds), and, well, we can perhaps refine the approach later if we hit an issue.

jotak:
/hold
discussed with @OlivierCazade , we actually don't need that
it was initially to fix https://github.com/netobserv/network-observability-operator/pull/1415 but instead of editing the generated dockerfile, we can pass `--label` via the build command

jotak:
not needed anymore, cf previous comment and https://github.com/netobserv/network-observability-operator/pull/1415

PR #2036, opened 2025-09-25
Fix upstream deploy/bundle issues with PF4 image

jotak:
The Makefile was screwing up the bundle when running first `make deploy` then `make update-bundle` (we had to manually discard changes from `make deploy` before running  `make update-bundle`  to avoid that)

This change updates all `sed` commands when updating manager.yaml so that it substitute next lines based on RELATED_IMAGE_X.

Note that we actually don't provide a plugin compat image in the community releases, so the tag `v1.9.2-community-pf4` is essentially a placeholder here. But we do support creating upstream OLM bundle/catalog for testing purpose, and this must work as well with compat images.


jotak:
Cherry-picking to avoid this kind of issue on release-1.10 branch too - no code change here
/cherry-pick release-1.10

PR #2039, opened 2025-09-25
Remove uneeded pip deps

jotak:


jotak:
@OlivierCazade : sounds like there are 4 unneeded python dependencies for the bundle build
I haven't found an easy way to clean/regenerate the requirement file, other than check manually one by one removing dependencies.. would be good to automate that; do you remember how you created it in the first place?

jotak:
alternatively we could also throw python away and use something else .....

PR #2044, opened 2025-09-26
Fixing np default deployment after upgrade

OlivierCazade:
## Description

The default value is now set at the section level.

## Dependencies

n/a

## Checklist

### DIFF
--- a/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
+++ b/config/crd/bases/flows.netobserv.io_flowcollectors.yaml
@@ -3909,7 +3909,7 @@ spec:
                       type: array
                     enable:
                       description: |-
-                        Set `enable` to `true` to deploy network policies on the namespaces used by NetObserv (main and privileged). It is disabled by default.
+                        Deploys network policies on the namespaces used by NetObserv (main and privileged).
                         These network policies better isolate the NetObserv components to prevent undesired connections to them.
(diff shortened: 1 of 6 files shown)


config/crd/bases/flows.netobserv.io_flowcollectors.yaml
```diff
@@ -3906,7 +3908,6 @@ spec:
                         type: string
                       type: array
                     enable:
-                      default: true
                       description: |-
                         Set `enable` to `true` to deploy network policies on the namespaces used by NetObserv (main and privileged). It is disabled by default.
```
Amoghrd:
```suggestion
                        Deploys network policies on the namespaces used by NetObserv (main and privileged). It is enabled by default.
```

OlivierCazade: Thanks!

OlivierCazade:
This fix is a tradeoff.

Now there should not be any case where the network policy get deployed during upgrade.

The default value for any new flowcollector will deploy the NP but if the NetworkPolicy is defined by the user, then the Enable field has to be manually set to true.

jotak:
@OlivierCazade  You need to update the CRD doc in source file (related to @Amoghrd 's commit suggestion) before regenerating the bundle

jotak:
I've just put back the text suggested by Amogh and that was reverted inadvertently

PR #2070, opened 2025-10-03
Update Alerts doc

jotak:
## Description

- Remove some details of the default alerts (thresholds & variants) , just describe the templates
- Add a section that describes metadata & the health annotation
- Move the embedded sample YAML into an actual sample


jotak:
(merging - no code change)

PR #2079, opened 2025-10-06
Add validation hook for network policies when Loki in same namespace

jotak:
## Description

We recommend to install Loki in a different namespace; if users still want to use the same ns, they need to disable the network policy and possibly create their own.

This webhook deals with these restrictions by rejecting invalid config.

## Dependencies

n/a

## Checklist

### DIFF
--- a/api/flowcollector/v1beta2/flowcollector_validation_webhook.go
+++ b/api/flowcollector/v1beta2/flowcollector_validation_webhook.go
@@ -62,6 +62,7 @@ func (r *FlowCollector) Validate(_ context.Context, fc *FlowCollector) (admissio
 	v := validator{fc: &fc.Spec}
 	v.validateAgent()
 	v.validateFLP()
+	v.validateNetPol()
 	v.warnLogLevels()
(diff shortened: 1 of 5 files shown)


Amoghrd:
@jotak Webhook validation works as expected. But since we are recommending Loki to be installed in a different NS, this page needs to be updated while deploying flowcollector. Basically make Loki namespace field mandatory and remove line mentioning assumed to be in same namespace
Screenshot 2025-10-06 at 11 23 49 AM

jotak:
> @jotak Webhook validation works as expected. But since we are recommending Loki to be installed in a different NS, this page needs to be updated while deploying flowcollector. Basically make Loki namespace field mandatory and remove line mentioning assumed to be in same namespace 

that makes sense but it's a slightly bigger change (if we take it for 1.10, I try to minimize the size of the change here)

Amoghrd:
That makes sense. I will create a follow-up bug for the same then
/label qe-approved

jotak:
thanks @Amoghrd !
/approve
/cherry-pick release-1.10

PR #2088, opened 2025-10-10
[backport 1.10] NETOBSERV-2439: New "part-of" label + limit netpol to netobserv pods

jotak:
Introduce the "part-of" label to designate by labels all resources part of the netobserv operator.

This aims to simplify user workflow if they need to do something related to all netobserv resources; such as configuring a network policy that only target netobserv pods, for instance.

Update the default policy to target only pods with this label.

Backport of #2078  and #2086


memodi:
@jotak - couple of things:

1. the upgrade scenario will need to be tested post-merge since on pre-merge bundle upgrade path doesn't exist.

2. when I try to create flowcollector on 1.10 pre-merge build., I hit validator check that Loki needs to be in separate NS:
```
$ oc apply -f /tmp/flowcollector.yaml
Error from server (Forbidden): error when creating "/tmp/flowcollector.yaml": admission webhook "flowcollectorconversionwebhook.netobserv.io" denied the request: cannot deploy the NetObserv network policy with Loki installed in the same namespace: it is recommended to install Loki in a separate namespace; alternatively, you can disable the NetObserv network policy in spec.networkPolicy, and create your own that covers Loki connectivity
```
is this check still valid? since we're now applying policy to only `part-of` netobserv workloads.

jotak:
@memodi strange that you get that on the pre-merge build: I removed that check, see https://github.com/netobserv/network-observability-operator/pull/2088/files#diff-9d1c644ef788ee18f068f964d3920be8ae211a795b60f9fa69dba25a1a786c23

jotak:
@memodi I've merged it so you can test the upgrade.
I'll open a new PR if there's anything wrong

jotak:
@memodi I just double-checked and I confirm not seeing the validation error

PR #2112, opened 2025-10-27
NETOBSERV-2365: Add the ability to create recording rules instead of alerts for the Network Health feature

leandroberetta:
## Description

Add the ability to create recording rules instead of alerts for the Network Health feature.

Instructions to test on console PR.

## Dependencies

https://github.com/netobserv/network-observability-console-plugin/pull/1163

## Checklist

### DIFF
--- a/internal/controller/consoleplugin/config/config.go
+++ b/internal/controller/consoleplugin/config/config.go
@@ -125,6 +125,7 @@ type HealthRuleVariantMetadata struct {
 	GroupBy            string            `yaml:"groupBy,omitempty" json:"groupBy,omitempty"`
 	LowVolumeThreshold string            `yaml:"lowVolumeThreshold,omitempty" json:"lowVolumeThreshold,omitempty"`
 	Thresholds         ThresholdMetadata `yaml:"thresholds" json:"thresholds"`
+	UpperBound         string            `yaml:"upperBound,omitempty" json:"upperBound,omitempty"`
 }
 
 type ThresholdMetadata struct {
--- a/internal/controller/consoleplugin/consoleplugin_objects.go
+++ b/internal/controller/consoleplugin/consoleplugin_objects.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"math"
 	"path/filepath"
 	"slices"
 	"strconv"
@@ -516,6 +517,29 @@ func (b *builder) getHealthRulesMetadata() []cfg.HealthRuleMetadata {
 
 		var variants []cfg.HealthRuleVariantMetadata
 		for _, variant := range healthRule.Variants {
+			// Calculate upperBound for trending alerts: max(threshold × 5, 100)
+			// For trending rules (LatencyHighTrend, ExternalEgressHighTrend, ExternalIngressHighTrend),
+			// we need an upper bound for score calculation
+			upperBound := ""
+			isTrending := healthRule.Template == flowslatest.HealthRuleLatencyHighTrend ||
+				healthRule.Template == flowslatest.HealthRuleExternalEgressHighTrend ||
+				healthRule.Template == flowslatest.HealthRuleExternalIngressHighTrend
+			if isTrending {
+				// Use the highest defined threshold (critical > warning > info)
+				thresholdStr := variant.Thresholds.Critical
+				if thresholdStr == "" {
+					thresholdStr = variant.Thresholds.Warning
+				}
+				if thresholdStr == "" {
+					thresholdStr = variant.Thresholds.Info
+				}
+				if thresholdStr != "" {
+					if val, err := strconv.ParseFloat(thresholdStr, 64); err == nil {
+						upperBound = strconv.Itoa(int(math.Max(val*5, 100)))
+					}
+				}
+			}
+
 			variants = append(variants, cfg.HealthRuleVariantMetadata{
 				GroupBy:            string(variant.GroupBy),
 				LowVolumeThreshold: variant.LowVolumeThreshold,
@@ -524,6 +548,7 @@ func (b *builder) getHealthRulesMetadata() []cfg.HealthRuleMetadata {
 					Warning:  variant.Thresholds.Warning,
 					Critical: variant.Thresholds.Critical,
 				},
+				UpperBound: upperBound,
 			})
 		}
 


api/flowcollector/v1beta2/flowcollector_alert_types.go
```diff
@@ -10,64 +10,75 @@ import (
...
 	// Possible values are: `PacketDropsByKernel`, `PacketDropsByDevice`, `IPsecErrors`, `NetpolDenied`,
 	// `LatencyHighTrend`, `DNSErrors`, `ExternalEgressHighTrend`, `ExternalIngressHighTrend`, `CrossAZ`.
-	// More information on alerts: https://github.com/netobserv/network-observability-operator/blob/main/docs/Alerts.md
+	// More information on health rules: https://github.com/netobserv/network-observability-operator/blob/main/docs/Alerts.md
 	// +kubebuilder:validation:Enum:="PacketDropsByKernel";"PacketDropsByDevice";"IPsecErrors";"NetpolDenied";"LatencyHighTrend";"DNSErrors";"ExternalEgressHighTrend";"ExternalIngressHighTrend";"CrossAZ"
 	// +required
-	Template AlertTemplate `json:"template,omitempty"`
+	Template HealthRuleTemplate `json:"template,omitempty"`
+
+	// Mode defines whether this health rule should be generated as an alert or a recording rule.
```
jotak, line 55: Perhaps add more information about what is the difference. Such as: "Recording rules violations are visible in the Network Health dashboard without generating any Prometheus alert."

api/flowcollector/v1beta2/flowcollector_alert_types.go
```diff
@@ -10,64 +10,75 @@ import (
...
 	// It is provided as an absolute rate (bytes per second or packets per second, depending on the context).
 	// When provided, it must be parsable as a float.
 	LowVolumeThreshold string `json:"lowVolumeThreshold,omitempty"`
 
-	// Thresholds of the alert per severity.
+	// Thresholds of the health rule per severity.
 	// They are expressed as a percentage of errors above which the alert is triggered. They must be parsable as floats.
-	// +required
-	Thresholds AlertThresholds `json:"thresholds,omitempty"`
+	// Required for alert mode, optional for recording mode.
```
jotak: Maybe something I don't get here: how can it be optional for recording mode, if we still want them to appear in the health dashboard, associated with a severity?

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -580,18 +580,19 @@ type FLPMetrics struct {
...
 
-	// `disableAlerts` is a list of alert groups that should be disabled from the default set of alerts.
+	// `disableHealthRules` is a list of health rule templates that should be disabled from the default set.
 	// Possible values are: `NetObservNoFlows`, `NetObservLokiError`, `PacketDropsByKernel`, `PacketDropsByDevice`, `IPsecErrors`, `NetpolDenied`,
 	// `LatencyHighTrend`, `DNSErrors`, `ExternalEgressHighTrend`, `ExternalIngressHighTrend`, `CrossAZ`.
-	// More information on alerts: https://github.com/netobserv/network-observability-operator/blob/main/docs/Alerts.md
+	// More information on health rules: https://github.com/netobserv/network-observability-operator/blob/main/docs/Alerts.md
 	// +optional
-	DisableAlerts []AlertTemplate `json:"disableAlerts"`
+	DisableHealthRules []HealthRuleTemplate `json:"disableHealthRules"`
```
jotak: I think for disabling, we should keep the existing API and continue to only affect alerts. Disabling is mostly to help users when they have too much alerting noise, and it's not so needed for recording rules
Also because we should not rename this field because it already existed before the TP feature, that would be a breaking change.

api/flowcollector/v1beta2/flowcollector_alert_types.go
```diff
@@ -10,64 +10,75 @@ import (
...
-	AlertLatencyHighTrend         AlertTemplate = "LatencyHighTrend"
-	AlertDNSErrors                AlertTemplate = "DNSErrors"
-	AlertExternalEgressHighTrend  AlertTemplate = "ExternalEgressHighTrend"
-	AlertExternalIngressHighTrend AlertTemplate = "ExternalIngressHighTrend"
-	AlertCrossAZ                  AlertTemplate = "CrossAZ"
-	GroupByNode                   AlertGroupBy  = "Node"
-	GroupByNamespace              AlertGroupBy  = "Namespace"
-	GroupByWorkload               AlertGroupBy  = "Workload"
+	HealthRuleNoFlows                  HealthRuleTemplate = "NetObservNoFlows"
+	HealthRuleLokiError                HealthRuleTemplate = "NetObservLokiError"
```
jotak: "NoFlows" and "LokiError" are different from the others, I think we should keep referring to them as alerts only; unlike the others, they are not health items for the dashboard, they're only for alerting.
We should probably add a validation check that they are not used in "recording" mode

internal/pkg/metrics/alerts/builder.go
```diff
@@ -110,27 +142,99 @@ func convertToRules(template flowslatest.AlertTemplate, alert *flowslatest.Alert
...
+	case flowslatest.HealthRuleLokiError, flowslatest.HealthRuleNoFlows:
 		// error
 	}
-	return nil, fmt.Errorf("unknown alert template: %s", rb.template)
+	return nil, fmt.Errorf("unknown health rule template: %s", rb.template)
 }
 
 func (rb *ruleBuilder) additionalDescription() string {
-	return fmt.Sprintf("You can turn off this alert by adding '%s' to spec.processor.metrics.disableAlerts in FlowCollector, or reconfigure it via spec.processor.metrics.alerts.", rb.template)
+	return fmt.Sprintf("You can turn off this health rule by adding '%s' to spec.processor.metrics.disableHealthRules in FlowCollector, or reconfigure it via spec.processor.metrics.healthRules.", rb.template)
```
jotak: (just fyi: we will remove this added description, as it will be moved to the runbooks ; I was about to say something about this message wrt disabled alerts, but we don't care)

internal/pkg/metrics/alerts/builder.go
```diff
@@ -110,27 +142,99 @@ func convertToRules(template flowslatest.AlertTemplate, alert *flowslatest.Alert
...
+			i++
+		}
+	}
+
+	return strings.ToLower(result.String())
+}
+
+// buildRecordingRuleName builds recording rule name following the convention:
+// netobserv:health:<template>:<groupby>:<side>:rate2m
+func (rb *ruleBuilder) buildRecordingRuleName() string {
```
jotak, line 207: I think I would prefer if the recording rule name was passed to `createRule` directly from each alert directly (in alerts.go) ; maybe it's less smart but it's more explicit and make it easier if there's a specific name or something that we want to change for any reason

Actually surely this function can be kept but it's the `acronyms` and `toSnakeCase` that IMO can be replaced with an explicit string per alert?

internal/pkg/metrics/alerts/builder.go
```diff
@@ -140,9 +244,22 @@ func (rb *ruleBuilder) createRule(promQL, summary, description string) (*monitor
...
+	if rb.healthRule.GroupBy != "" {
+		gr = "Per" + string(rb.side) + string(rb.healthRule.GroupBy)
 	}
+
+	// Generate recording rule
+	if rb.mode == flowslatest.ModeRecording {
+		recordName := rb.buildRecordingRuleName()
+		return &monitoringv1.Rule{
+			Record: recordName,
+			// Note: Recording rules cannot have annotations in Prometheus
```
jotak, line 248: really? So we cannot pass any metadata?

api/flowcollector/v1beta2/flowcollector_alert_types.go
leandroberetta: I originally thought the recording rule to be just a value, without the notion of a severity, but now that you mention, we can display the value with severities. You're right. Thanks for the feedback.

internal/pkg/metrics/alerts/builder.go
leandroberetta: Only labels: https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/

That would be an option although is not as good as annotations for these kind of information.

api/flowcollector/v1beta2/flowcollector_alert_types.go
```diff
@@ -10,64 +10,88 @@ import (
...
+	HealthRuleDNSNxDomain              HealthRuleTemplate = "DNSNxDomain"
+	HealthRuleExternalEgressHighTrend  HealthRuleTemplate = "ExternalEgressHighTrend"
+	HealthRuleExternalIngressHighTrend HealthRuleTemplate = "ExternalIngressHighTrend"
+
+	GroupByNode      HealthRuleGroupBy = "Node"
+	GroupByNamespace HealthRuleGroupBy = "Namespace"
+	GroupByWorkload  HealthRuleGroupBy = "Workload"
+
+	ModeAlert     HealthRuleMode = "alert"
+	ModeRecording HealthRuleMode = "recording"
```
jotak: hey, I only noticed that when reviewing the runbooks: per openshift API guidelines, we should use CapitalCamelCase for enum values, so it should be `Alert` and `Recording`.

Another question that I'm wondering, is if the name "Recording" is meaningful enough. Technically speaking it is derived from prometheus "recording rules", but for an uninformed user, does it make sense?
Maybe something like "MetricOnly" or "HealthOnly" makes more sense? Or, do you have any other idea? I fear I'm not the best for naming things.

leandroberetta: Thanks for the heads up on the CapitalCamelCase, I will changed it. 

In terms of naming, I'm not the best either but I guess you're right. We can be pretty explicit:

Alert or RecordingRule?

leandroberetta: Also, Alert or Record could be, but I know require some kind of knowledge about the feature.

internal/pkg/metrics/alerts/builder.go
```diff
@@ -120,43 +122,134 @@ func convertToRules(template flowslatest.AlertTemplate, alert *flowslatest.Alert
...
+// toSnakeCase converts a camelCase or PascalCase string to snake_case
+// Handles known acronyms specially (e.g., "DNSErrors" -> "dns_errors", "CrossAZ" -> "cross_az")
+func toSnakeCase(s string) string {
+	var result strings.Builder
+	i := 0
+
+	for i < len(s) {
+		// Check if current position matches any acronym
+		matched := false
+		for acronym, replacement := range acronyms {
```
jpinsonneau: A map is not ordered in golang. That means the range over acronyms may not go over `IPsec` before `IP` :smile_cat: 

You should consider an array like
```
var acronyms = []struct {
	original    string
	replacement string
}{
	{"IPsec", "ipsec"},
	{"HTTPS", "https"},
	{"HTTP", "http"},
	{"DNS", "dns"},
	{"TCP", "tcp"},
	{"UDP", "udp"},
	{"IP", "ip"},
	{"AZ", "az"},
}
```

jpinsonneau: Seems like there are possible functions for that
https://stackoverflow.com/questions/56616196/how-to-convert-camel-case-string-to-snake-case

But we'll need a special case for IPSec:
```
var (
	// Regex patterns for converting camelCase/PascalCase to snake_case
	// First pattern: matches any character followed by uppercase+lowercase (e.g., "eErrors" in "DNSErrors")
	matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
	// Second pattern: matches lowercase/digit followed by uppercase (e.g., "sE" in "DNSErrors")
	matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
)

// toSnakeCase converts a camelCase or PascalCase string to snake_case
// Uses regex-based approach that automatically handles acronyms and word boundaries
// Examples: "DNSErrors" -> "dns_errors", "IPsecErrors" -> "ipsec_errors", "CrossAZ" -> "cross_az"
func toSnakeCase(s string) string {
	if len(s) == 0 {
		return ""
	}
	// Normalize special compound words before processing
	// "IPsec" should be treated as "IPSec" so regex handles it correctly
	s = strings.ReplaceAll(s, "IPsec", "IPSec")

	snake := matchFirstCap.ReplaceAllString(s, "${1}_${2}")
	snake = matchAllCap.ReplaceAllString(snake, "${1}_${2}")
	result := strings.ToLower(snake)

	// Fix the normalized compound word back to its original form
	result = strings.ReplaceAll(result, "ip_sec", "ipsec")
	return result
}
```

internal/pkg/metrics/alerts/builder.go
jotak: I will maybe follow-up later regarding this comment

leandroberetta: Sorry I thought I addressed this one. I will take it into account for a follow-up PR.

PR #2113, opened 2025-10-28
Add flowmetrics sample for workload flow matrix

jotak:
Workload flows matrix for TCP/SYN traffic, including destination port, with labels remapped for focusing on incoming or outgoing traffic


jotak:
merging, this is just sample files

PR #2142, opened 2025-11-13
NETOBSERV-2402: Adding lokistack status to console plugin configmap

OlivierCazade:
## Description

Add Lokistack status to the plugin configmap.

The Lokistack is also now watched to update any status change.

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/flowcollector_controller.go
+++ b/internal/controller/flowcollector_controller.go
@@ -12,12 +12,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	flowslatest "github.com/netobserv/network-observability-operator/api/flowcollector/v1beta2"
 	"github.com/netobserv/network-observability-operator/internal/controller/consoleplugin"
+	"github.com/netobserv/network-observability-operator/internal/controller/constants"
 	"github.com/netobserv/network-observability-operator/internal/controller/ebpf"
 	"github.com/netobserv/network-observability-operator/internal/controller/loki"
 	"github.com/netobserv/network-observability-operator/internal/controller/reconcilers"
@@ -35,9 +39,11 @@ const (
 // FlowCollectorReconciler reconciles a FlowCollector object
 type FlowCollectorReconciler struct {
 	client.Client
-	mgr     *manager.Manager
-	status  status.Instance
-	watcher *watchers.Watcher
+	mgr                *manager.Manager
+	status             status.Instance
+	watcher            *watchers.Watcher
+	ctrl               controller.Controller
+	lokiWatcherStarted bool
 }
 
 func Start(ctx context.Context, mgr *manager.Manager) (manager.PostCreateHook, error) {
@@ -66,8 +72,15 @@ func Start(ctx context.Context, mgr *manager.Manager) (manager.PostCreateHook, e
 		builder.Owns(&osv1.ConsolePlugin{}, reconcilers.UpdateOrDeleteOnlyPred)
 	}
 
-	if mgr.ClusterInfo.HasLokiStack() {
-		builder.Watches(&lokiv1.LokiStack{}, &handler.EnqueueRequestForObject{})
+	if mgr.ClusterInfo.HasLokiStack(ctx) {
+		builder.Watches(
+			&lokiv1.LokiStack{},
+			handler.EnqueueRequestsFromMapFunc(func(_ context.Context, _ client.Object) []ctrl.Request {
+				// When a LokiStack changes, trigger reconcile of the FlowCollector
+				return []ctrl.Request{{NamespacedName: constants.FlowCollectorName}}
+			}),
+		)
+		r.lokiWatcherStarted = true
 		log.Info("LokiStack CRD detected")
 	}
 
@@ -75,6 +88,7 @@ func Start(ctx context.Context, mgr *manager.Manager) (manager.PostCreateHook, e
 	if err != nil {
 		return nil, err
 	}
+	r.ctrl = ctrl
 	r.watcher = watchers.NewWatcher(ctrl)
 
 	return nil, nil
@@ -102,6 +116,14 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	// Dynamically start LokiStack watcher if the API became available
+	if desired.Spec.Loki.Mode == flowslatest.LokiModeLokiStack {
+		if err := r.ensureLokiStackWatcher(ctx); err != nil {
+			l.Error(err, "Failed to start LokiStack watcher")
+			// Don't fail reconciliation, just log the error
+		}
+	}
+
 	// At the moment, status workflow is to start as ready then degrade if necessary
 	// Later (when legacy controller is broken down into individual controllers), status should start as unknown and only on success finishes as ready
 	r.status.SetReady()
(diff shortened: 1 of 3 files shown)


jotak (approved):
/lgtm

internal/controller/flowcollector_controller.go
```diff
@@ -68,6 +70,11 @@ func Start(ctx context.Context, mgr *manager.Manager) error {
 		log.Info("CNO not detected: using ovnKubernetes config and reconciler")
 	}
 
+	if mgr.ClusterInfo.HasLokiStack() {
+		builder.Watches(&lokiv1.LokiStack{}, &handler.EnqueueRequestForObject{})
```
jotak: ideally, we would enqueue only when our configured lokistack is affected, not all lokistacks
but I guess that's fine, as we don't expect hundreds of lokistacks out there :-)

memodi: from Claude review:

```
  Problem: This will create reconcile requests for LokiStack objects, not FlowCollectors. When a LokiStack named "logging-loki" changes, it will try to reconcile a FlowCollector named "logging-loki", which likely doesn't exist.

  Fix Required:
  builder.Watches(&lokiv1.LokiStack{}, handler.EnqueueRequestsFromMapFunc(
      func(ctx context.Context, obj client.Object) []reconcile.Request {
          lokiStack := obj.(*lokiv1.LokiStack)
          var flowCollectors flowslatest.FlowCollectorList
          if err := mgr.GetClient().List(ctx, &flowCollectors); err != nil {
              log.FromContext(ctx).Error(err, "Failed to list FlowCollectors")
              return []reconcile.Request{}
          }

          var requests []reconcile.Request
          for _, fc := range flowCollectors.Items {
              if fc.Spec.Loki.Mode == flowslatest.LokiModeLokiStack &&
                 fc.Spec.Loki.LokiStack.Name == lokiStack.Name {
                  ns := fc.Spec.Loki.LokiStack.Namespace
                  if ns == "" {
                      ns = fc.Namespace
                  }
                  if ns == lokiStack.Namespace {
                      requests = append(requests, reconcile.Request{
                          NamespacedName: types.NamespacedName{
                              Name:      fc.Name,
                              Namespace: fc.Namespace,
                          },
                      })
                  }
              }
          }
          return requests
      },
  ))

```

internal/controller/consoleplugin/consoleplugin_reconciler.go
```diff
@@ -179,8 +180,22 @@ func (r *CPReconciler) reconcilePlugin(ctx context.Context, builder *builder, de
...
+	lokiStack := &lokiv1.LokiStack{}
+	if desired.Loki.Mode == flowslatest.LokiModeLokiStack {
+		if r.ClusterInfo.HasLokiStack() {
+			ns := desired.Loki.LokiStack.Namespace
+			if ns == "" {
+				ns = desired.Namespace
+			}
+			if err := r.Client.Get(ctx, types.NamespacedName{Name: desired.Loki.LokiStack.Name, Namespace: ns}, lokiStack); err != nil {
+				lokiStack = nil
+				log.FromContext(ctx).Info("Could not get the LokiStack resource.")
```
memodi: from Claude review:
```
  Problems:
  - Logs at Info level instead of Warning
  - Actual error is discarded
  - Can't distinguish between "not found" and "permission denied"
  - Should surface in FlowCollector status

  Fix:
  if err := r.Client.Get(ctx, types.NamespacedName{Name: desired.Loki.LokiStack.Name, Namespace: ns}, lokiStack); err != nil {
      lokiStack = nil
      if apierrors.IsNotFound(err) {
          log.FromContext(ctx).Info("LokiStack resource not found, status will not be available",
              "name", desired.Loki.LokiStack.Name,
              "namespace", ns)
      } else {
          log.FromContext(ctx).Error(err, "Failed to get LokiStack resource",
              "name", desired.Loki.LokiStack.Name,
              "namespace", ns)
      }
      // TODO: Consider surfacing this in FlowCollector status
  }
```

internal/controller/flowcollector_controller.go
OlivierCazade: I did not think about making a k8s query inside the handler to filter Lokistacks.

The approach I thought was to wait for the flowcollector to be created and start a dedicated controller with a static flowcollector name. This was adding a lot of complexity and I was not sure if it was worth it.

This looks like a more simple solution, to the price of a k8s querry in the handler function.

@jotak what do you think ?

jotak: Not sure we need to overcomplicate things here. It's basically my comment here: https://github.com/netobserv/network-observability-operator/pull/2142#discussion_r2545921343 ; having a couple of false-positive reconcile events is not so important; we're talking about lokistack objects, it's not expected to have many, and they don't change often.

btw I think claude answer is wrong the enqueue request is not for a flow-collector named after the loki stack, it's for any flow-collector? (`EnqueueRequestForObject{}` with empty params)

jotak: If we really want to narrow down to our configured lokistack, in other situation we just keep in controller state the last-time seen element that we want to check (we do that in a couple of places for flowcollector.spec.namespace iirc); we could do the same with the configured lokistack.
Like this: https://github.com/netobserv/network-observability-operator/blob/main/internal/controller/flp/flp_controller.go#L60

internal/controller/consoleplugin/consoleplugin_objects.go
```diff
@@ -493,9 +494,24 @@ func (b *builder) setFrontendConfig(fconf *cfg.FrontendConfig) error {
 	return nil
 }
 
+func getLokiStatus(lokiStack *lokiv1.LokiStack) string {
```
jotak: not tested, but looking at the code I think there's a problem here:
In `reconcileConfigMap`, the `lokiStack` object is created and can have the following values:
- a reference to the lokistack that was found
- nil if an error occurred
- a reference to an empty (0-value) struct `&lokiv1.LokiStack{}` when not in LokiStack mode

So in this last situation, the returned value would be `pending`, which seems incorrect?

IMO, what we could do:
- if there was an error when fetching LokiStack, set this error as status (e.g. the console plugin could display something like "LokiStack not found")
- if the lokistack shows a non-ready status (error/pending condition), set a message with that condition status
- if the loki stack is ready, set as ready
- if not in lokistack mode, set as an empty string

Also, I'm not sure it's useful to check for the presence of the LokiStack API: if it's configured in LokiStack mode BUT the API is not present, the config is wrong, so it's ok to just display the error message that would come up when trying to fetch LokiStack?
wdyt?

OlivierCazade: In theory, we should not enter the case of not being in LokiStack, but since we are passing LokiStack as a pointer I addded a not nil check. I modified the function to make this more understandable.

When not in LokiStack mode, the operator fallback to the previous mode, using the StatusURL.

If the lokistack shows a non-ready status, finding the right error might be tricky, each LokiStack component has its own status meaning we may display the wrong error. IMO this is simpler to display a pending status and let the user investigate, the LokiStack was provided by the user.

I removed the LokiStack API check.

internal/controller/consoleplugin/consoleplugin_reconciler.go
```diff
@@ -179,8 +180,28 @@ func (r *CPReconciler) reconcilePlugin(ctx context.Context, builder *builder, de
 	return nil
 }
 
-func (r *CPReconciler) reconcileConfigMap(ctx context.Context, builder *builder) (string, error) {
-	newCM, configDigest, err := builder.configMap(ctx)
+func (r *CPReconciler) reconcileConfigMap(ctx context.Context, builder *builder, desired *flowslatest.FlowCollectorSpec) (string, error) {
+	lokiStack := &lokiv1.LokiStack{}
+	if desired.Loki.Mode == flowslatest.LokiModeLokiStack {
+		ns := desired.Loki.LokiStack.Namespace
```
jotak: I just double-checked trying your PR, and confirm it provides a wrong status when not in LokiStack mode (it says pending) ; that's because you initialize it as non-nil.

This should work I think:

```suggestion
	var lokiStack *lokiv1.LokiStack
	if desired.Loki.Mode == flowslatest.LokiModeLokiStack {
		lokiStack = &lokiv1.LokiStack{}
		ns := desired.Loki.LokiStack.Namespace
```

OlivierCazade: Thanks!

memodi:
@OlivierCazade needs rebase here...

memodi:
@OlivierCazade - I am seeing an issue where the field in configmap is never updated from `pending` state, despite waiting for more than 10 mins since Loki reported `Ready`:


```
Every 2.0s: oc get cm/console-plugin-config -o yaml -n netobserv | grep -i status:                                                                                                      memodi-mac: Tue Jan 27 12:40:46 2026

      status: pending

-------


  status:
    components:
      compactor:
        Ready:
        - lokistack-compactor-0
      distributor:
        Ready:
        - lokistack-distributor-66b85bf7cb-dnrxl
      gateway:
        Ready:
        - lokistack-gateway-766bb497fd-f8vwb
        - lokistack-gateway-766bb497fd-t274x
      indexGateway:
        Ready:
        - lokistack-index-gateway-0
      ingester:
        Ready:
        - lokistack-ingester-0
      querier:
        Ready:
        - lokistack-querier-68fd69cfc5-m5mpj
      queryFrontend:
        Ready:
        - lokistack-query-frontend-69844dffd7-mf2wk
      ruler:
        Failed: []
        Pending: []
        Ready: []
        Running: []
    conditions:
    - lastTransitionTime: "2026-01-27T17:29:13Z"
      message: All components ready
      reason: ReadyComponents
      status: "True"
      type: Ready
```
Also, how do we plan to use `status` field ? 

could you PTAL?

OlivierCazade:
@memodi this should be fixed now, can you test again ?

memodi:
> @memodi this should be fixed now, can you test again ?

okay, can you confirm if this is a priority for 1.11?

memodi:
works much better now! thanks @OlivierCazade

PR #2145, opened 2025-11-14
NETOBSERV-2450: NETOBSERV-2460: webhook validation for sdn and direct mode

jotak:
## Description

**[NETOBSERV-2450](https://issues.redhat.com//browse/NETOBSERV-2450)**
- Refactor ClusterInfo to improve how readiness works: prevent getting uninitialized data
- Fetch Network resource via ClusterInfo to get CNI
- Ignore netpol creation with openshift-sdn
- For unknown CNI, allow netpol creation with a warning
- Changed default rules: it's now only with ovn that it's enabled by
  default

**[NETOBSERV-2460](https://issues.redhat.com//browse/NETOBSERV-2460)**
- warn when using direct mode with more than 15 nodes

## Dependencies

n/a

## Checklist

  * [x] If so, fill in "Release Note Text" in the JIRA.

### DIFF
--- a/api/flowcollector/v1beta2/flowcollector_types.go
+++ b/api/flowcollector/v1beta2/flowcollector_types.go
@@ -70,7 +70,7 @@ type FlowCollectorSpec struct {
 	ConsolePlugin FlowCollectorConsolePlugin `json:"consolePlugin,omitempty"`
 
 	// `deploymentModel` defines the desired type of deployment for flow processing. Possible values are:<br>
-	// - `Direct` (default) to make the flow processor listen directly from the agents using the host network, backed by a DaemonSet.<br>
+	// - `Direct` (default) to make the flow processor listen directly from the agents using the host network, backed by a DaemonSet. Only recommended on small clusters, below 15 nodes.<br>
 	// - `Service` to make the flow processor listen as a Kubernetes Service, backed by a scalable Deployment.<br>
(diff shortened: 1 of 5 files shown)


jpinsonneau (approved):

Amoghrd:
@jotak I see the netpol is not created for SDN networkType clusters so that bug is working fine. And also see warning for the same

But I dont see warning wrt using Direct on large clusters. I scaled up the cluster to 25 nodes and dont see the warning during/after creation of flowcollector

jotak:
@Amoghrd that's because we don't refresh all the time the cluster info (which nodes count is part of), it's only done once at startup when the operator starts. I think if you restart manually the operator pod, then you should see the warning.

In fact, that's not a new thing, the same happens with all the cluster info that we keep in memory (cluster ID for multi-cluster, available APIs (such as console plugin or servicemonitors), cluster versions etc.)
We can definitely fix this, but I'd prefer not doing it via this zstream release, because it makes the change a bit more risky (involving concurrent accesses), I prefer to keep that for 1.11

So in 1.10.1 we would have the limitation that this warning message won't be displayed after a cluster upscaling (or will still be displayed after downscaling). I think that's corner cases which we can accept.

Ticket created: [NETOBSERV-2494](https://issues.redhat.com//browse/NETOBSERV-2494)

PR #2153, opened 2025-11-18
NETOBSERV-2146: Add trackedKinds parameter to FLP config

leandroberetta:
## Description

Add trackedKinds parameter to FLP config

### How to test

With this change, Deployment and Gateway can owners, in the case of gateway API:

ReplicaSet -> Deployment -> Gateway

To test this, we can use Istio that uses the gateway API for ingress traffic into the cluster.

To install Service Mesh 3:

```
oc apply -f - <<'EOF'
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: servicemeshoperator3
  namespace: openshift-operators
spec:
  channel: stable
  name: servicemeshoperator3
  source: redhat-operators
  sourceNamespace: openshift-marketplace
  installPlanApproval: Automatic

---
apiVersion: v1
kind: Namespace
metadata:
  name: istio-cni

---
apiVersion: sailoperator.io/v1
kind: IstioCNI
metadata:
  name: default
spec:
  version: v1.24.6
  namespace: istio-cni

---
apiVersion: v1
kind: Namespace
metadata:
  name: istio-system

---
apiVersion: sailoperator.io/v1
kind: Istio
metadata:
  name: default
spec:
  version: v1.24.6
  namespace: istio-system
EOF
```

After these commands, we will get Istio working on the cluster. Now we need to install Bookinfo, the sample application and a gateway to get traffic into the mesh.

```
oc new-project bookinfo
oc label namespace bookinfo istio-discovery=enabled istio-injection=enabled
oc apply -f https://raw.githubusercontent.com/openshift-service-mesh/istio/release-1.24/samples/bookinfo/platform/kube/bookinfo.yaml -n bookinfo
oc apply -f https://raw.githubusercontent.com/openshift-service-mesh/istio/release-1.24/samples/bookinfo/gateway-api/bookinfo-gateway.yaml -n bookinfo
```

Finally, to test the UI, we can generate traffic using the [following script](https://gist.github.com/leandroberetta/529a0973128d82fc2acfc45eec6df52b#file-generate-bookinfo-traffic-sh):

```
./generate-bookinfo-traffic.sh
```

## Dependencies

https://github.com/netobserv/flowlogs-pipeline/pull/1125
https://github.com/netobserv/network-observability-console-plugin/pull/1119
https://github.com/netobserv/network-observability-console-plugin/pull/1165

## Checklist


internal/controller/flp/flp_pipeline_builder.go
```diff
@@ -194,6 +194,7 @@ func (b *PipelineBuilder) AddProcessorStages() error {
 		SubnetLabels: flpLabels,
 		KubeConfig: api.NetworkTransformKubeConfig{
 			SecondaryNetworks: secondaryNetworks,
+			TrackedKinds:      []string{"ReplicaSet", "Deployment", "Gateway"},
```
jotak, line 218: So we should remove "Gateway" from there, right? (according to https://github.com/netobserv/flowlogs-pipeline/pull/1125#discussion_r2535129402 )

leandroberetta: @jotak thanks for reviewing.

That is still require given the tracking logic, which wasn't require anymore was the informer (because with n+1 is ok, in this case, Deployments informer).

If we remove Gateway from that list, the ownership tracking will stop at Deployment.

The logic looks for valid parents, and Gateway is a valid one.

jotak: ah, I see, thanks for the explanation

jotak (approved):
/lgtm
/hold

(on hold, to remove the replace directive before merging)

leandroberetta:
> /lgtm /hold
> 
> (on hold, to remove the replace directive before merging)

removing the on hold as the replace directive is now removed.

PR #2157, opened 2025-11-19
NETOBSERV-2508: Network policy fix for HyperShift

leandroberetta:
## Description

In HyperShift deployments and other external control plane configurations, the `flowlogs-pipeline` pods were unable to connect to the Kubernetes API server, resulting in timeout errors:

  Cannot get dynamic config: Get "https://172.30.0.1:443/api/v1/namespaces/netobserv/configmaps/flowlogs-pipeline-config-dynamic": dial tcp 172.30.0.1:443: i/o timeout
  Failed to watch err="failed to list *v1.Pod: Get https://172.30.0.1:443/api/v1/pods?limit=500&resourceVersion=0\": dial tcp 172.30.0.1:443: i/o 
  timeout"

In HyperShift deployments, the control plane (API server, etcd, controllers) runs in a separate management cluster, while worker nodes run in a different cluster. This is evident from:

  1. **HyperShift label**: The `kubernetes` service endpoint has the label `hypershift.openshift.io/managed: "true"`
  3. **No local pods**: The `default` namespace contains no pods serving the API server endpoint

  The previous NetworkPolicy only allowed egress traffic to port 6443 when targeting specific in-cluster namespaces (`openshift-apiserver`, `openshift-kube-apiserver`) using `namespaceSelector` with
  `podSelector`. However, external control plane endpoints are not represented as pods within the worker cluster, so the namespace selector rules don't apply and traffic is blocked.

_NB: this issue is specific to main branch; release-1.10 is unaffected_

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/networkpolicy/apiserver_endpoint.go
+++ b/internal/controller/networkpolicy/apiserver_endpoint.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/netobserv/network-observability-operator/internal/pkg/cluster"
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -18,32 +19,35 @@ const (
 )
 
 // GetAPIServerEndpointIPs retrieves the API server endpoint IP addresses.
-// It first tries to use EndpointSlice API (v1), and falls back to Endpoints API if unavailable.
-func GetAPIServerEndpointIPs(ctx context.Context, cl client.Client) ([]string, error) {
+// It uses EndpointSlice API if available, otherwise falls back to Endpoints API.
+func GetAPIServerEndpointIPs(ctx context.Context, cl client.Client, clusterInfo *cluster.Info) ([]string, error) {
 	logger := log.FromContext(ctx)
 
-	// Try EndpointSlice first (discovery.k8s.io/v1, available since k8s 1.21)
-	ips, err := getEndpointIPsFromEndpointSlice(ctx, cl)
-	if err == nil && len(ips) > 0 {
+	var ips []string
+	var err error
+
+	// Use EndpointSlice if available (discovery.k8s.io/v1, available since k8s 1.21)
+	if clusterInfo.HasEndpointSlices() {
+		logger.V(1).Info("Using EndpointSlice API to get API server endpoint IPs")
+		ips, err = getEndpointIPsFromEndpointSlice(ctx, cl)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get API server endpoint IPs from EndpointSlice: %w", err)
+		}
 		logger.V(1).Info("Retrieved API server endpoint IPs from EndpointSlice", "ips", ips)
-		return ips, nil
-	}
-
-	if err != nil {
-		logger.V(1).Info("Failed to get EndpointSlice, falling back to Endpoints API", "error", err)
-	}
-
-	// Fallback to Endpoints API (core/v1, deprecated but widely available)
-	ips, err = getEndpointIPsFromEndpoints(ctx, cl)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get API server endpoint IPs: %w", err)
+	} else {
+		// Fallback to Endpoints API (core/v1, deprecated but widely available)
+		logger.V(1).Info("EndpointSlice API not available, using Endpoints API")
+		ips, err = getEndpointIPsFromEndpoints(ctx, cl)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get API server endpoint IPs from Endpoints: %w", err)
+		}
+		logger.V(1).Info("Retrieved API server endpoint IPs from Endpoints", "ips", ips)
 	}
 
 	if len(ips) == 0 {
 		return nil, fmt.Errorf("no API server endpoint IPs found")
 	}
 
-	logger.V(1).Info("Retrieved API server endpoint IPs from Endpoints", "ips", ips)
 	return ips, nil
 }
 
--- a/internal/controller/networkpolicy/np_controller.go
+++ b/internal/controller/networkpolicy/np_controller.go
@@ -80,7 +80,7 @@ func (r *Reconciler) reconcile(ctx context.Context, clh *helper.Client, desired 
 	// Get API server endpoint IPs for network policy
 	var apiServerIPs []string
 	if r.mgr.ClusterInfo.IsOpenShift() {
-		apiServerIPs, err = GetAPIServerEndpointIPs(ctx, r.Client)
+		apiServerIPs, err = GetAPIServerEndpointIPs(ctx, r.Client, r.mgr.ClusterInfo)
 		if err != nil {
 			l.Error(err, "Failed to get API server endpoint IPs")
 			return fmt.Errorf("cannot determine API server endpoint IPs: %w", err)
--- a/internal/pkg/cluster/cluster.go
+++ b/internal/pkg/cluster/cluster.go
@@ -68,12 +68,12 @@ func (c *Info) fetchAvailableAPIs(ctx context.Context) error {
 		return err
 	}
 	apisMap := map[string]bool{
-		consolePlugin: false,
-		cno:           false,
-		svcMonitor:    false,
-		promRule:      false,
-		ocpSecurity:   false,
-    endpointSlices: false,
+		consolePlugin:  false,
+		cno:            false,
+		svcMonitor:     false,
+		promRule:       false,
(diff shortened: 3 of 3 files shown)


jotak:
Hmm, this weakens considerably the egress policy
There should be another way, no? We're surely not the first having to deal with that

leandroberetta:
> Hmm, this weakens considerably the egress policy There should be another way, no? We're surely not the first having to deal with that

I will investigate this issue. I think that we may want to watch the endpoint IP and add it to the network policy, but that would be something more complex, I don't think that ip would change, maybe we can do it one time, although if it changes, won't work, so we may want to watch that endpoint.

jotak:
@leandroberetta yes, I asked the ovn folks, restricting on 172.20.0.1:6443 should work, and like you said we can know this address by watching the endpoint.
The only downside I see, is that the Endpoint API is deprecated so sooner or later it will break; but ok, we will fix it when that happens, I guess
EDIT: I see it's also defined as an EndpointSlice, so we can use the new API

leandroberetta:
> @leandroberetta yes, I asked the ovn folks, restricting on 172.20.0.1:6443 should work, and like you said we can know this address by watching the endpoint. The only downside I see, is that the Endpoint API is deprecated so sooner or later it will break; but ok, we will fix it when that happens, I guess EDIT: I see it's also defined as an EndpointSlice, so we can use the new API

Hi @jotak, good to know, yes, I implemented a version where first looks for EndpointSlice and fallback to Endpoint in case the new API is not available.

image

It's working fine for HyperShift, although I'm not sure how to properly try if the ip changes (I'm not sure is a common case or even possible). Same for IPv6 clusters (network policy needs CIDR so the /32 or /128 in the range needs to be provided accordingly).

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -253,7 +253,7 @@ metadata:
     categories: Monitoring, Networking, Observability
     console.openshift.io/plugins: '["netobserv-plugin"]'
     containerImage: quay.io/netobserv/network-observability-operator:1.10.0-community
-    createdAt: "2025-11-25T09:57:07Z"
+    createdAt: "2025-11-25T14:02:27Z"
```
jotak: just a FYI : there's no real requirement to keep updating the createdAt timestamp, even though it's automatically changed when we run `make update-bundle`.
Personally I often revert that line change from the commit, because it's something that can often create rebasing conflicts. So it makes rebasing (and backporting) easier to not have it.
cc @luisjira , as you ran into that situation yesterday

internal/controller/networkpolicy/apiserver_endpoint.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	logger.V(1).Info("Retrieved API server endpoint IPs from Endpoints", "ips", ips)
+	return ips, nil
+}
+
+// getEndpointIPsFromEndpointSlice retrieves endpoint IPs using the EndpointSlice API
+func getEndpointIPsFromEndpointSlice(ctx context.Context, cl client.Client) ([]string, error) {
+	// List all EndpointSlices for the kubernetes service
+	endpointSliceList := &discoveryv1.EndpointSliceList{}
```
jotak: why not fetching the slice directly (rather than via label matching) ? Is it that we're not sure of the name?

internal/controller/networkpolicy/np_controller.go
```diff
@@ -30,11 +45,33 @@ func Start(ctx context.Context, mgr *manager.Manager) error {
...
+		//nolint:staticcheck // SA1019: Endpoints is deprecated but used as fallback for k8s < 1.21
+		builder = builder.Watches(&corev1.Endpoints{}, handler.EnqueueRequestsFromMapFunc(enqueueFlowCollectorOnEndpointChange))
+	}
+
+	return builder.Complete(&r)
+}
+
+// isEndpointSliceAvailable checks if the EndpointSlice API (discovery.k8s.io/v1) is available
+// in the cluster. This API was introduced in Kubernetes 1.21.
+func isEndpointSliceAvailable(mgr *manager.Manager) bool {
```
jotak: we have a different pattern for discovering APIs, see in `cluster.go` function `fetchAvailableAPIs` : you can add slices to the API map there, instead

internal/controller/networkpolicy/np_controller.go
```diff
@@ -70,11 +107,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
+
 	cni, err := r.mgr.ClusterInfo.GetCNI()
 	if err != nil {
 		return err
 	}
-	npName, desiredNp := buildMainNetworkPolicy(desired, r.mgr, cni)
+
+	// Get API server endpoint IPs for network policy
+	var apiServerIPs []string
+	if r.mgr.ClusterInfo.IsOpenShift() {
```
jotak, line 82: I don't think it is specific to Openshift? Is it?

internal/controller/networkpolicy/np_controller.go
```diff
@@ -30,11 +45,33 @@ func Start(ctx context.Context, mgr *manager.Manager) error {
...
-	return ctrl.NewControllerManagedBy(mgr).
+
+	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&flowslatest.FlowCollector{}, reconcilers.IgnoreStatusChange).
 		Named("networkPolicy").
-		Owns(&networkingv1.NetworkPolicy{}, reconcilers.UpdateOrDeleteOnlyPred).
-		Complete(&r)
+		Owns(&networkingv1.NetworkPolicy{}, reconcilers.UpdateOrDeleteOnlyPred)
+
+	// Watch EndpointSlice if available (preferred, k8s >= 1.21), otherwise fallback to Endpoints
```
jotak: I'm wondering if it's worth it to have this watch here. I guess it's very unlikely that this IP changes? Because it comes at the price of watching yet-another-kind with pretty high cardinality, cluster-wide, so it has an impact on memory & bandwidth consumption.
(I'm not even sure this IP can change, or not without triggering other disruptive changes that would anyway trigger the reconcile loop - but I'm speculating here)

jotak: (note that if we prefer to have the watch anyway, we've implemented a more memory-friendly watch in the `narrowcache` package)

internal/controller/networkpolicy/np_controller.go
```diff
@@ -70,11 +107,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result
...
 		return err
 	}
-	npName, desiredNp := buildMainNetworkPolicy(desired, r.mgr, cni)
+
+	// Get API server endpoint IPs for network policy
+	var apiServerIPs []string
+	if r.mgr.ClusterInfo.IsOpenShift() {
+		apiServerIPs, err = GetAPIServerEndpointIPs(ctx, r.Client)
+		if err != nil {
+			l.Error(err, "Failed to get API server endpoint IPs, network policy will allow all IPs on port 6443")
```
jotak: I think I'm rather for not creating the egress rule at all, than creating a less secure one. Ie. I'm rather for having it breaking the connectivity, forcing the user to troubleshoot the issue, rather than opening all egress on 6443 somewhat silently

internal/controller/networkpolicy/np_objects.go
```diff
@@ -166,6 +185,40 @@ func buildMainNetworkPolicy(desired *flowslatest.FlowCollector, mgr *manager.Man
...
+				}
+			}
+			np.Spec.Egress = append(np.Spec.Egress, networkingv1.NetworkPolicyEgressRule{
+				To: peers,
+				Ports: []networkingv1.NetworkPolicyPort{{
+					Protocol: ptr.To(corev1.ProtocolTCP),
+					Port:     ptr.To(intstr.FromInt32(constants.K8sAPIServerPort)),
+				}},
+			})
+		} else {
```
jotak: Would remove the whole else block (see previous comment)

internal/controller/networkpolicy/np_controller.go
jotak: ok I see, this policy rule is only created for openshift; forget my comment :-)

internal/controller/networkpolicy/np_controller.go
jotak: PS I see you were wondering the same thing:
> It's working fine for HyperShift, although I'm not sure how to properly try if the ip changes (I'm not sure is a common case or even possible). Same for IPv6 clusters (network policy needs CIDR so the /32 or /128 in the range needs to be provided accordingly).

I'd say we can remove the watches

leandroberetta:
@jotak thanks for your review! All feedback is addressed now.

PR #2169, opened 2025-11-20
NETOBSERV-2484: improve ovn subnets detection

jotak:
## Description

- Check for internal/masquerade/transitswitch subnets
- Add API server

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/flp/detect_subnets.go
+++ b/internal/controller/flp/detect_subnets.go
@@ -94,7 +94,7 @@ func (r *Reconciler) getOpenShiftSubnets(ctx context.Context) ([]flowslatest.Sub
 	}
 	if len(extIPs) > 0 {
 		subnets = append(subnets, flowslatest.SubnetLabel{
-			Name:  "ExternalIP",
+			Name:  "EXT:ExternalIP",
 			CIDRs: extIPs,
 		})
 	}


internal/controller/flp/flp_controller.go
```diff
@@ -232,59 +233,94 @@ func reconcileMonitoringCerts(ctx context.Context, info *reconcilers.Common, tls
...
+	machines, err = readMachineFromConfig(cm)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(machines) > 0 {
-		subnets = append(subnets, machines...)
+	// Additional OVN subnets
+	networkOp := &operatorv1.Network{}
+	err = r.Get(ctx, types.NamespacedName{Name: "cluster"}, network)
```
OlivierCazade:
```suggestion
	err = r.Get(ctx, types.NamespacedName{Name: "cluster"}, networkOp)
```
If I understand correctly this is networkOp that you wanted to get here.

jotak: oops, good catch, but ... it worked when I tested!! 🧐
weird...

jotak: oh actually, it worked because it means I only tested the defaults 😅

OlivierCazade (approved):
LGTM!

internal/controller/flp/detect_subnets.go
```diff
@@ -0,0 +1,183 @@
...
+	if ovnk != nil {
+		if ovnk.V4InternalSubnet != "" {
+			internalSubnet = ovnk.V4InternalSubnet
+		}
+		if ovnk.IPv4 != nil && ovnk.IPv4.InternalTransitSwitchSubnet != "" {
+			transitSwitchSubnet = ovnk.IPv4.InternalTransitSwitchSubnet
+		}
+		if ovnk.GatewayConfig != nil && ovnk.GatewayConfig.IPv4.InternalMasqueradeSubnet != "" {
+			masqueradeSubnet = ovnk.GatewayConfig.IPv4.InternalMasqueradeSubnet
+		}
```
jpinsonneau, line 177: Can't we import https://raw.githubusercontent.com/openshift/cluster-network-operator/refs/heads/master/pkg/network/ovn_kubernetes.go here ?

So at some point if their defaults or CR changes, we'll be updated automatically.

jotak: Those constants aren't exported.
Also, if they change, we will have a bigger problem anyway, because that would be tied to the openshift version, so we would have to write specific code anyway. I just hope it won't change :-)

jotak: I'm merging it, as this is approved - let me know if it's not satisfying, we can still have follow-ups

jpinsonneau: The `GetInternalSubnets` function is public if you want to import that in the future :wink:

jotak: Yes but like I said it doesn't really help, as we need to support past ovn versions, we will always need some hardcoding if the IPs change

PR #2171, opened 2025-11-21
NETOBSERV-2418: NETOBSERV-2419: new FlowCollectorSlice API

jotak:
## Description

- First commit is just the API definition (CRD) and the related boilerplate
- Second commit is the actual implementation

TODO:

- [x] ~Some tests!~

## Dependencies

- Console: https://github.com/netobserv/network-observability-console-plugin/pull/1166

## Checklist

### DIFF
--- a/config/samples/flows_v1alpha1_flowcollectorslice.yaml
+++ b/config/samples/flows_v1alpha1_flowcollectorslice.yaml
@@ -3,7 +3,7 @@ kind: FlowCollectorSlice
 metadata:
   name: flowcollectorslice-sample
 spec:
-  sampling: 10
+  sampling: 1
   # subnetLabels:
   # - name: my-database
   #   cidrs:
--- a/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
@@ -11,64 +11,21 @@ metadata:
             "name": "flowcollectorslice-sample"
           },
           "spec": {
-            "sampling": 10
+            "sampling": 1
           }
         },
         {
           "apiVersion": "flows.netobserv.io/v1alpha1",
           "kind": "FlowMetric",
           "metadata": {
-            "labels": {
-              "app.kubernetes.io/created-by": "netobserv-operator",
-              "app.kubernetes.io/instance": "flowmetric-sample",
-              "app.kubernetes.io/managed-by": "kustomize",
-              "app.kubernetes.io/name": "flowmetric",
-              "app.kubernetes.io/part-of": "netobserv-operator"
-            },
             "name": "flowmetric-sample"
           },
           "spec": {
-            "charts": [
-              {
-                "dashboardName": "Main",
-                "queries": [
-                  {
-                    "legend": "",
-                    "promQL": "sum(rate($METRIC[2m]))"
-                  }
-                ],
-                "title": "External ingress traffic",
-                "type": "SingleStat",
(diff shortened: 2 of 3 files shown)


config/samples/flows_v1alpha1_flowcollectorslice.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: flows.netobserv.io/v1alpha1
+kind: FlowCollectorSlice
+metadata:
+  name: flowcollectorslice-sample
+spec:
+  sampling: 10
```
jpinsonneau: Can you please add an example with subnets here ? (commented out by default)

api/flowcollectorslice/v1alpha1/flowcollectorslice_types.go
```diff
@@ -0,0 +1,59 @@
...
+	// Label name, used to flag matching flows.
+	//+required
+	Name string `json:"name,omitempty"`
+}
+
+// FlowCollectorSliceStatus defines the observed state of FlowCollectorSlice
+type FlowCollectorSliceStatus struct {
+	// `conditions` represent the latest available observations of an object's state
+	Conditions []metav1.Condition `json:"conditions"`
+}
```
jpinsonneau, line 42: It would be interesting to fill the status here.

I'm not sure about reporting overall netobserv components readiness, even if that would be useful for project admins not having access to cluster FlowCollector.

However, at least an overall readiness status and error messages would be good to have.

config/samples/flows_v1alpha1_flowcollectorslice.yaml
jotak: done

api/flowcollectorslice/v1alpha1/flowcollectorslice_types.go
jotak: I've added some error here, related to the slice config, in particular related to the subnets and if there is overlaps with the admin config or with previously processed slices

Amoghrd:
Yup as Leandro suggested, while testing I see the issue wherein there is no warning when not using loki

jotak:
hey @Amoghrd @leandroberetta that's a good catch, indeed the task description says that:
> A warning must be displayed if per-tenant is enabled and loki is enabled but not in LokiStack mode (because data is not isolated between tenants in that case)

However, when I wrote that, the intent was to call this API something like `NetObservTenant`, and the warning aimed at removing potential confusion when customers use this API while not having actual multi-tenancy configured with Loki. But now that the API was renamed into `FlowCollectorSlice` without any mention of a tenant, I don't think this warning still makes sense
wdyt?

Amoghrd:
So in that case, flowCollectorSlice can be used with and without Loki?

jotak:
> So in that case, flowCollectorSlice can be used with and without Loki?

yes correct, it's independent from loki

leandroberetta:
> hey @Amoghrd @leandroberetta that's a good catch, indeed the task description says that:
> 
> > A warning must be displayed if per-tenant is enabled and loki is enabled but not in LokiStack mode (because data is not isolated between tenants in that case)
> 
> However, when I wrote that, the intent was to call this API something like `NetObservTenant`, and the warning aimed at removing potential confusion when customers use this API while not having actual multi-tenancy configured with Loki. But now that the API was renamed into `FlowCollectorSlice` without any mention of a tenant, I don't think this warning still makes sense wdyt?

I see, makes sense!

PR #2197, opened 2025-11-27
dockerignore

jpinsonneau:
## Description

Improve dockerignore to speed up image build
- reduce unecessary context when building images
- avoid copying sensitive content

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,60 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore build and test binaries.
-bin/
-testbin/
+
+# Build artifacts
+**/bin/
+**/testbin/
+manager
+*.o
+*.tar
+*.tgz
+
+# Test files and directories
+**/*_test.go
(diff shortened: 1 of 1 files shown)


jpinsonneau:
> .[dockerignore docs](https://docs.docker.com/build/concepts/context/#syntax) mention `**/*.go` to ignore all files ending with `.go` anywhere in the build context, including subdirectories.
> 
> I think it makes sense to add `**/` before the lines that don't have a fixed location in the root directory.

indeed that would be better ! Thanks for pointing that @luisjira !

jotak:
I don't understand why the konflux jobs aren't triggered...
I'd like to make sure the downstream bundle build isn't affected

jotak:
Oh, I know, you need to add `".dockerignore".pathChanged()` in the on-cell-expression of every tekton jobs

jotak:
@jpinsonneau you didn't add what I asked:

> Oh, I know, you need to add ".dockerignore".pathChanged() in the on-cell-expression of every tekton jobs

PR #2271, opened 2025-12-15
NETOBSERV-2545: Add quick filters for external traffic

jotak:
## Description

Now that external traffic detection starts getting quite accurate, adding a quick filter looks like a quick win:

image

## Dependencies

n/a

## Checklist

### DIFF
--- a/docs/FlowCollector.md
+++ b/docs/FlowCollector.md
@@ -12062,7 +12062,7 @@ When a subnet matches the source or destination IP of a flow, a corresponding fi
         <td>
           `customLabels` allows you to customize subnets and IPs labeling, such as to identify cluster external workloads or web services.
 External subnets must be labeled with the prefix `EXT:`, or not labeled at all, in order to work with default quick filters and some metrics examples provided.<br/>
-If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to configure manually labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
+If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to manually configure labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
 If `openShiftAutoDetect` is enabled, `customLabels` overrides the detected subnets when they overlap.<br/><br/>
         </td>
         <td>false</td>
--- a/api/flowcollector/v1beta2/flowcollector_types.go
+++ b/api/flowcollector/v1beta2/flowcollector_types.go
@@ -1456,7 +1456,7 @@ type SubnetLabels struct {
 
 	// `customLabels` allows you to customize subnets and IPs labeling, such as to identify cluster external workloads or web services.
 	// External subnets must be labeled with the prefix `EXT:`, or not labeled at all, in order to work with default quick filters and some metrics examples provided.<br/>
-	// If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to configure manually labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
+	// If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to manually configure labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
 	// If `openShiftAutoDetect` is enabled, `customLabels` overrides the detected subnets when they overlap.<br/>
 	//+optional
 	CustomLabels []SubnetLabel `json:"customLabels,omitempty"`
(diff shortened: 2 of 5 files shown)


stleerh:
I'm trying to understand how this works and what flows I will see. Let's just take the "External ingress" case.  The filter is:
`src_subnet_label: '"",EXT:'`

1. This assumes `openShiftAutoDetect: true`. It is the default, so it's likely the case, but this should be mentioned.
2. When does it set the label to "EXT:"?  I don't see this documented anywhere.
3. If I define a custom label for an external CIDR, will it still work?
4. If I generate traffic outside of the cluster to a service, it creates two flows, one at the node level (external to router) and another at the service level (router to service).  Is the first flow only considered "External ingress" or both?

jotak:
@stleerh 
> I'm trying to understand how this works and what flows I will see. Let's just take the "External ingress" case. The filter is: `src_subnet_label: '"",EXT:'`
> 
>     1. This assumes `openShiftAutoDetect: true`. It is the default, so it's likely the case, but this should be mentioned.

Right, good point. Also I haven't thought much about the non-openshift case, maybe I should make it openshift-only.

> 
>     2. When does it set the label to "EXT:"?  I don't see this documented anywhere.
+
>     3. If I define a custom label for an external CIDR, will it still work?

I'd recommend to take this PR in the light of my blog here: https://github.com/netobserv/netobserv.github.io/pull/29
Without custom labelling, these quick filters should work just by the "empty label" match, and "EXT" serves no purpose in that case.
But the idea with EXT is precisely for this to continue to work when users start creating custom labels. The goal is to make it a convention: if what you're labelling is for external traffic, then prefix it with EXT.
I don't think we have any way to enforce that rule, but we can push that as a convention.

You're right about the doc, I should add that in the subnet labels doc

>     4. If I generate traffic outside of the cluster to a service, it creates two flows, one at the node level (external to router) and another at the service level (router to service).  Is the first flow only considered "External ingress" or both?

yes, only the first

stleerh:
Let's document the expectations and requirements, and provide clarity on what "External ingress" is (e.g. only the flow part of a conversation where the source is external).  Perhaps, add a tooltip to point to an online documentation.

One way to remove the requirement that external CIDRs must begin with "EXT:" for this to work is to have a separate field for object type instead of overloading subnet label.  It will need one for source and destination.  The object type will have a fixed set of enum values like "pod", "service", "external", etc., similar to what `openShiftAutoDetect` does.

jotak:
@stleerh 
> Let's document the expectations and requirements, 

done

> and provide clarity on what "External ingress" is (e.g. only the flow part of a conversation where the source is external). Perhaps, add a tooltip to point to an online documentation.

I'm not convinced extra doc is needed here, I don't feel there's anything unexpected in how it's done. Traffic between openshift-ingress and internal workloads IS cluster traffic, I don't think it's ambiguous.

> 
> One way to remove the requirement that external CIDRs must begin with "EXT:" for this to work is to have a separate field for object type instead of overloading subnet label. It will need one for source and destination. The object type will have a fixed set of enum values like "pod", "service", "external", etc., similar to what `openShiftAutoDetect` does.

yeah, we can improve that in a follow-up; but let's try to not create any breaking change, if users start to leverage the EXT prefix already. We could just make it so that when the object type is "external", the prefix is automatically added, but the end result is the same with the label

docs/FlowCollector.md
```diff
@@ -12059,8 +12060,10 @@ When a subnet matches the source or destination IP of a flow, a corresponding fi
         <td><b><a href="#flowcollectorspecprocessorsubnetlabelscustomlabelsindex">customLabels</a></b></td>
         <td>[]object</td>
         <td>
-          `customLabels` allows to customize subnets and IPs labelling, such as to identify cluster-external workloads or web services.
-If you enable `openShiftAutoDetect`, `customLabels` can override the detected subnets in case they overlap.<br/>
+          `customLabels` allows you to customize subnets and IPs labeling, such as to identify cluster external workloads or web services.
+External subnets must be labeled with the prefix `EXT:`, or not labeled at all, in order to work with default quick filters and some metrics examples provided.<br/>
+If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to configure manually labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
```
Amoghrd:
```suggestion
If `openShiftAutoDetect` is disabled or you are not using OpenShift, it is recommended to manually configure labels for the cluster subnets, to distinguish internal traffic from external traffic.<br/>
```

docs/FlowCollector.md
```diff
@@ -12059,8 +12060,10 @@ When a subnet matches the source or destination IP of a flow, a corresponding fi
...
+If `openShiftAutoDetect` is enabled, `customLabels` overrides the detected subnets when they overlap.<br/><br/>
```
Amoghrd, line 12066:
```suggestion
If `openShiftAutoDetect` is enabled, `customLabels` overrides the detected subnets if there is overlap.<br/><br/>
```

docs/FlowCollector.md
jotak: done

docs/FlowCollector.md
jotak: is there something wrong with "when they overlap" ? I like it better than "if there is overlap", but not sure :-)

Amoghrd: yeah dont mind the `when they overlap` too

PR #2310, opened 2026-01-08
bump ebpf agent

jotak:


oliver-smakal:
I can confirm the PR fixes the issue, it looks good upon manual inspection and the [previously failing  testcase is passing as well](https://github.com/user-attachments/files/24499795/81410.log).

PR #2323, opened 2026-01-12
NETOBSERV-2516: Links for runbooks in alerts

leandroberetta:
## Description

Links for runbooks in alerts.

image

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/controller/consoleplugin/config/static-frontend-config.yaml
+++ b/internal/controller/consoleplugin/config/static-frontend-config.yaml
@@ -1170,7 +1170,7 @@ filters:
       Specify a single TCP state name like:
               - A _LINUX_TCP_STATES_H number like 1, 2, 3
               - A _LINUX_TCP_STATES_H TCP name like ESTABLISHED, SYN_SENT, SYN_RECV
-    docUrl: https://github.com/torvalds/linux/blob/main/include/net/tcp_states.h
+    docUrl: https://github.com/torvalds/linux/blob/master/include/net/tcp_states.h
   - id: pkt_drop_cause
     name: Packet drop latest cause
     component: autocomplete


internal/pkg/metrics/alerts/builder_test.go
```diff
@@ -209,3 +208,34 @@ func TestToSnakeCase(t *testing.T) {
...
+		{
+			template: "NetObservLokiError",
+			expected: "https://github.com/netobserv/runbooks/blob/main/alerts/network-observability-operator/NetObservLokiError.md",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.template, func(t *testing.T) {
+			url := buildRunbookURL(tt.template)
+			assert.Equal(t, tt.expected, url)
```
jpinsonneau, line 269: Should we also test that each URL exists ? :thinking: 

That may help to maintain those, especially when we add more.

Currently, those return a 404 :angel:

leandroberetta: That sounds more like an integration test but can be good for pushing  https://github.com/openshift/runbooks/pull/373.

internal/pkg/metrics/alerts/builder.go
```diff
@@ -262,6 +271,7 @@ func (rb *ruleBuilder) createRule(promQL, summary, description string) (*monitor
 			"description":                 description,
 			"summary":                     summary,
 			"netobserv_io_network_health": string(bAnnot),
+			"runbook_url":                 buildRunbookURL(string(rb.template)),
```
jotak, line 281: we should also add them in the "links" of the health annotation, so they will be picked up as well in our health page
plus, that will work for recording rules as well, if I'm correct

internal/pkg/metrics/alerts/builder_test.go
jotak: +1 to check for all available templates/runbooks

leandroberetta: I implemented the suggestion, we need to merge the runbooks PR because test will start failing now :)

internal/pkg/metrics/alerts/builder.go
leandroberetta: I added the link.

internal/pkg/metrics/alerts/builder_test.go
leandroberetta: tests are passing which means files now exist, which made sense because we merged the runbook PR.

internal/controller/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -1170,7 +1170,7 @@ filters:
       Specify a single TCP state name like:
               - A _LINUX_TCP_STATES_H number like 1, 2, 3
               - A _LINUX_TCP_STATES_H TCP name like ESTABLISHED, SYN_SENT, SYN_RECV
-    docUrl: https://github.com/torvalds/linux/blob/master/include/net/tcp_states.h
+    docUrl: https://github.com/torvalds/linux/blob/main/include/net/tcp_states.h
```
memodi: the main branch doesn't seem to exist,  the original link works for me https://github.com/torvalds/linux/blob/master/include/net/dropreason-core.h

memodi:
minor comment for the link update, looks good otherwise.

/label qe-approved

internal/controller/consoleplugin/config/static-frontend-config.yaml
jotak:
```suggestion
    docUrl: https://github.com/torvalds/linux/blob/master/include/net/tcp_states.h
```

jotak: +1, not sure why this change ?

internal/controller/consoleplugin/config/static-frontend-config.yaml
```diff
@@ -1180,7 +1180,7 @@ filters:
       Specify a single packet drop cause like:
               - A _LINUX_DROPREASON_CORE_H number like 2, 3, 4
               - A _LINUX_DROPREASON_CORE_H SKB_DROP_REASON name like NOT_SPECIFIED, NO_SOCKET, PKT_TOO_SMALL
-    docUrl: https://github.com/torvalds/linux/blob/master/include/net/dropreason-core.h
+    docUrl: https://github.com/torvalds/linux/blob/main/include/net/dropreason-core.h
```
jotak:
```suggestion
    docUrl: https://github.com/torvalds/linux/blob/master/include/net/dropreason-core.h
```

internal/controller/consoleplugin/config/static-frontend-config.yaml
leandroberetta: Sorry, this shouldn't be commited, I suspect of a search and replace I've done.

PR #2348, opened 2026-01-20
NETOBSERV-2574 Reconcile race condition - API discovery 

jpinsonneau:
## Description

The operator was silently failing to create the `netobserv-hostnetwork-flp` ClusterRoleBinding on OpenShift clusters when the API server was slow or timing out during startup. This caused flowlogs-pipeline pods to fail with SCC (Security Context Constraint) violations.

### Root Cause

When the Kubernetes API server was slow during operator startup, the `fetchAvailableAPIs()` function would receive partial discovery results with the OpenShift SCC API (`security.openshift.io`) marked as failed. The code treated this as success, setting `IsOpenShift()` to `false`, which caused the operator to skip creating required ClusterRoleBindings.

## Solution

Fail fast when critical APIs fail to be discovered:

- Added validation to detect when the OpenShift SCC API is unavailable
- Return an error instead of silently continuing with incorrect cluster detection
- OLM will retry operator startup until the API server is responsive

### Before
- API discovery could fail silently during slow API server conditions
- Operator would start with incorrect cluster type detection
- Required ClusterRoleBindings were never created
- Pods failed to start with SCC violations
- Issue persisted for 10+ days until manual restart

### After
- Critical API discovery failures are detected immediately
- Operator refuses to start with clear error message
- OLM automatically retries until API server is healthy

## Dependencies

n/a

## Checklist

### DIFF
--- a/internal/pkg/cluster/cluster.go
+++ b/internal/pkg/cluster/cluster.go
@@ -75,6 +75,11 @@ func (c *Info) fetchAvailableAPIs(ctx context.Context) error {
 
 // fetchAvailableAPIsInternal discovers available APIs and optionally allows continuing despite critical API failures
 // allowCriticalFailure should be true during refresh loops to allow recovery from transient API server issues
+//
+// API Discovery Policy:
+// - APIs are marked as available when first discovered
+// - Once discovered, APIs are never marked unavailable (to avoid transient discovery issues)
+// - Operator restart is required to detect removed APIs (rare in practice)
 func (c *Info) fetchAvailableAPIsInternal(ctx context.Context, allowCriticalFailure bool) error {
 	log := log.FromContext(ctx)
 	_, resources, err := c.dcl.ServerGroupsAndResources()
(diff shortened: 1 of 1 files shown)


internal/pkg/cluster/cluster.go
```diff
@@ -80,19 +92,35 @@ func (c *Info) fetchAvailableAPIs(ctx context.Context) error {
...
 	for apiName := range apisMap {
 		if hasAPI(apiName, resources) {
 			apisMap[apiName] = true
-		} else if discErr != nil {
+		} else if hasDiscoveryError {
 			// Check if the wanted API is in error
-			for gv, err := range discErr.Groups {
+			for gv, gvErr := range discErr.Groups {
 				if strings.Contains(apiName, gv.String()) {
-					log.Error(err, "some API-related features are unavailable; you can check for stale APIs with 'kubectl get apiservice'", "GroupVersion", gv.String())
```
jotak, line 90: @jpinsonneau was that message displayed in the logs or not, in the support case? I believe it should have?

jpinsonneau: I haven't seen it but I'm having only partial logs here.

That would either confirm the API discovery completely failing and the code never reaching that log or the failure happend earlier and the discovery never been refreshed.

internal/pkg/cluster/cluster.go
```diff
@@ -80,19 +92,35 @@ func (c *Info) fetchAvailableAPIs(ctx context.Context) error {
...
+					log.Error(gvErr, "some API-related features are unavailable; you can check for stale APIs with 'kubectl get apiservice'", "GroupVersion", gv.String(), "api", apiName)
+
+					// OCP Security API is critical - we MUST know if we're on OpenShift
```
jotak: I think the logic works fine for detecting openshift, but this case uncovers other issues that we may have with other APIs - even if that's not as critical, it's also a bug if, for instance, the servicemonitors API is temporarily down at startup, then later comes up: we would always consider it down.
So I think we should include the discovering process in the refresh mechanism

jpinsonneau: That should make it: https://github.com/netobserv/network-observability-operator/pull/2348/commits/0cb952ae67f6df161c7e7b8b9c70a5d5e25c9ba6

Also took the opportunity to implement an exponential backoff retry mechanism from 30s to 10min max when failing.

internal/pkg/cluster/cluster.go
```diff
@@ -65,39 +70,89 @@ func NewInfo(ctx context.Context, cfg *rest.Config, dcl *discovery.DiscoveryClie
...
+			oldAvailable := oldApisMap[apiName]
+			if !oldAvailable && newAvailable {
+				log.Info("API recovered and is now available", "api", apiName)
+				apisRecovered = true
+			}
+		}
+	}
+	c.apisMap = newApisMap
+
+	log.Info("API detection finished", "apis", newApisMap)
```
jotak: I may have another concern, which I'm not sure of, it's how sounded is the API discovery/availability, as we've seen in the past API that could become "unavailable", which doesn't mean it's not supposed to be there. What to do if an API becomes unavailable, but still installed?

An option is to only update newly available APIs, not the newly unavailable ones. The downside is that we never remove APIs, but should we expect it to happen really?

Anyway, let me know your thoughts; I've tested the code below against your unit tests and that still passes

```suggestion

	// Track which critical APIs failed discovery
	criticalAPIFailed := false
	apisRecovered := false
	firstRun := false

	c.apisMapLock.Lock()
	defer c.apisMapLock.Unlock()
	if c.apisMap == nil {
		c.apisMap = map[string]bool{
			consolePlugin:  false,
			cno:            false,
			svcMonitor:     false,
			promRule:       false,
			ocpSecurity:    false,
			endpointSlices: false,
		}
		firstRun = true
	}

	for apiName, discovered := range c.apisMap {
		// Never remove a discovered API, to avoid transient staleness issues triggering changes continuously
		if !discovered {
			if hasAPI(apiName, resources) {
				c.apisMap[apiName] = true
				if !firstRun {
					apisRecovered = true
					log.Info("API recovered and is now available", "api", apiName)
				}
			} else if hasDiscoveryError {
				// Check if the wanted API is in error
				for gv, gvErr := range discErr.Groups {
					if strings.Contains(apiName, gv.String()) {
						log.Error(gvErr, "some API-related features are unavailable; you can check for stale APIs with 'kubectl get apiservice'", "GroupVersion", gv.String(), "api", apiName)

						// OCP Security API is critical - we MUST know if we're on OpenShift
						// to avoid wrong security context configurations
						if apiName == ocpSecurity {
							criticalAPIFailed = true
						}
					}
				}
			}
		}
	}

	if firstRun {
		log.Info("API detection finished", "apis", c.apisMap)
	}
```

jpinsonneau: Sounds valid to me :ok_hand: We can improve later if we find a corner case

jpinsonneau: I can't merge your suggestion directly: `Applying suggestions on deleted lines is currently not supported. `
Applied the changes here with a comment: https://github.com/netobserv/network-observability-operator/pull/2348/commits/e039f00f95eebf0b21c22beb3c8ec06b910d2623

PR #2420, opened 2026-02-05
NETOBSERV-2477: network predefined metrics + add missing ipsec

jpinsonneau:
## Description

Add missing network predefined metrics 
Added some tests to cover new / existing metrics
Added missing ipsec metrics on namespaces and workloads from the CRD (they were already implemented in code)

## Dependencies

n/a

## Checklist


jotak:
I'm looking into that, and I think we have two options here about the network metrics:
1. add them as a new "family" of metrics, like we have per workload/namespace/node, we're now adding per network, like is done here
2. add them as new labels in an existing family.

The second option is preferable because it allows to correlate networks with other entities, BUT only if it's not completely orthogonal to that family. For example, I wouldn't add the networks labels to nodes or namespaces metrics, because there can be many different networks per node or per namespace, so that would increase cardinality a lot. But I think per workload would be fine. So, instead of creating entirely new metrics, I would just add the labels into the workload metrics.

jotak:
(rebased)

jotak:
(rebased)

jotak:
(rebased)

jotak:
(cherry-picked manually)

PR #2421, opened 2026-02-05
NETOBSERV-2610 fix defaults / optional enable fields

jpinsonneau:
## Description

Fix missing annotations on `enable` flags in the API to ensure `openAPIV3Schema` is correctly set.

`//+kubebuilder:default:=`
Use when the field has a consistent default value that doesn't depend on runtime conditions. This makes the default visible in CRD schemas and console forms. Omit for conditional defaults (e.g., depends on CNI detection) where the operator needs to distinguish "not set" from "explicitly set to false".

`//+optional`
Always add for pointer types (*bool, *int32, etc.) since they can be nil. This documents the field as optional in generated API references and follows Kubernetes conventions.

Also remove unecessary fields in the example (since those are handled by the defaults now).

## Dependencies

n/a

## Checklist

### DIFF
--- a/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
@@ -48,12 +48,14 @@ metadata:
                 "sampling": 50
               }
             },
+            "deploymentModel": "Service",
             "loki": {
               "mode": "Monolithic",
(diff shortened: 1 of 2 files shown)


jpinsonneau:
Opened https://github.com/netobserv/network-observability-operator/pull/2423 as a quick win

PR #2443, opened 2026-02-10
Update recommendations:

jotak:
- Refresh from the downstream doc, some items were outdated (e.g. test beds still had the old 65 nodes)
- Extract test bed data out of the table, to clarify these are *not* recommendations
- Add info about where to find the mentioned settings
- Refresh cacheMaxSize with updated information from this release
- Rename Kafka consumer replicas => Consumer replicas, according to this release updates
- Add recommended deployment model more explicitly
- Recommend Service rather than Direct in 10-nodes clusters

cc @memodi @gwynnemonahan

### DIFF
--- a/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/netobserv-operator.clusterserviceversion.yaml
@@ -586,13 +586,15 @@ spec:
     kubectl edit flowcollector cluster
     ```
 
-    As it operates cluster-wide on every node, only a single `FlowCollector` is allowed, and it has to be named `cluster`.
+    Only a single `FlowCollector` is allowed, and it has to be named `cluster`.
 
     A couple of settings deserve special attention:
 
     - Sampling (`spec.agent.ebpf.sampling`): a value of `100` means: one flow every 100 is sampled. `1` means all flows are sampled. The lower it is, the more flows you get, and the more accurate are derived metrics, but the higher amount of resources are consumed. By default, sampling is set to 50 (ie. 1:50). Note that more sampled flows also means more storage needed. We recommend to start with default values and refine empirically, to figure out which setting your cluster can manage.
 
     - Loki (`spec.loki`): configure here how to reach Loki. The default values match the Loki quick install paths mentioned above, but you might have to configure differently if you used another installation method. Make sure to disable it (`spec.loki.enable`) if you don't want to use Loki.
+
+    - Processor replicas (`spec.processor.consumerReplicas`): how many replicas of `flowlogs-pipeline` should be deployed. Those pods collect, transform and re-export network flows. They can also be configured as unmanaged via `unmanagedReplicas`, if you want to use an auto-scaler.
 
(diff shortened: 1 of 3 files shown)


bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -603,23 +603,24 @@ spec:
...
+    | eBPF agent memory limit<br>*In `FlowCollector` `spec.agent.ebpf.resources`*       | 800Mi (default)        | 800Mi (default)     | 1600Mi               |
+    | eBPF agent cache size<br>*In `FlowCollector` `spec.agent.ebpf.cacheMaxSize`*      | 50,000                 | 120,000 (default)   | 120,000 (default)    |
+    | Processor memory limit<br>*In `FlowCollector` `spec.processor.resources`*         | 800Mi (default)        | 800Mi (default)     | 800Mi (default)      |
+    | Processor replicas<br>*In `FlowCollector` `spec.processor.consumerReplicas`*      | 3 (default)            | 6                   | 18                   |
+    | Deployment model<br>*In `FlowCollector` `spec.deploymentModel`*                   | Service (default)      | Kafka               | Kafka                |
+    | Kafka partitions<br>*In your Kafka installation*                                  | N/A                    | 48                  | 48                   |
+    | Kafka brokers<br>*In your Kafka installation*                                     | N/A                    | 3 (default)         | 3 (default)          |
```
memodi, line 626: we're dropping LokiStack size recommendation? IMO we should keep it 1x.extra-small (10 nodes), 1x.small (25 nodes) and 1x.medium (250 nodes)

jotak: I moved that out of the recommendations, to the test beds. It's still visible above. idk, I think we have a different place where we document the recommended loki stack size with different criteria, I don't want to mix up things here, wdyt?

See also: https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/network_observability/installing-network-observability-operators#loki-deployment-sizing_network_observability

memodi: I see, okay, that sounds good to me then.

stleerh:
Prior to the **Resource considerations** section, it says:

    As it operates cluster-wide on every node, only a single `FlowCollector` is allowed, and it has to be named `cluster`.

    A couple of settings deserve special attention:

The first sentence should be changes since it's not necessarily *every node* anymore.

A new setting that should be added is **consumer replicas**.

jotak:
> The first sentence should be changes since it's not necessarily _every node_ anymore.
> 

You refer to the node toleration settings, right? Hmm... how would you reformulate? If we want to keep the justification why it's a singleton resource, we may run into a convoluted explanation , because it's true that now, it could make sense to have several FlowCollectors.
I would say either we keep it as is as a "globally true-ish assertion, simplified" :-)
Or we just remove the justification.

> A new setting that should be added is **consumer replicas**.

It's not really new when we used Kafka before; but yes, can be added

jotak:
@stleerh done; I've removed the mention of "every nodes"

mffiedler:
LGTM

PR #2444, opened 2026-02-11
Fix doc issues with slices

jotak:
- Mention "EXT:" prefix in slices
- Mention "EXT:" prefix in label names
- Fix bad copy-paste of FlowMetric within Slices doc

Regenerate docs

cc @gwynnemonahan


gwynnemonahan:
Thank you!

gwynnemonahan:
LGTM

PR #2482, opened 2026-02-19
NETOBSERV-1101: TLS/mTLS hardening

### LINKED ISSUE #2360: Helm: setup TLS by default with trust-manager
Currently, our [installation guide](https://github.com/netobserv/network-observability-operator/blob/main/README.md#install-with-helm) suggests disabling TLS (by default) between the agent pods and flowlogs-pipeline; if they want TLS, users need to provide their certificates in the right format.

The installation process currently relies just on cert-manager. It's not sufficient, or not well adapted, for agent-to-flp communication, as the two components are in different namespaces. With just cert-manager, we can easily create a certificate for the flowlogs-pipeline service, but not have the CA available in the agent namespace, netobserv-privileged.

[Trust-manager](https://cert-manager.io/docs/trust/trust-manager/) seems to be the solution. It should allow us to synchronize the desired CA across namespaces.

> It's designed to complement cert-manager and works well when consuming CA certificates used by a cert-manager Issuer or ClusterIssuer

It might also be useful for mTLS.

jotak:
## Description

Fixes #2360

- [Upstream] Use trust-manager for agent->flp TLS
- New available configs for TLS/mTLS between agent and FLP

This is adding a new API config:
`spec.processor.service`, where users can configure TLS/mTLS.

- Improve certificates config with:
  - An optional self-signed issuer
  - A netobserv-ca certificate, issued by either provided or self-signed |-> A new netobserv-issuer that uses this CA
  |   |-> which issues operator webhook & metrics certs
  |   |-> now also issues FLP service cert
  |-> Copied to netobserv-privileged with trust-manager Bundle
- Add knob to disable cert/trust managers
  - With a new TLS.md doc to describe which certificates must be
    provided by other means, if disabled
- Add NOTES.txt which is displayed after install
- Some minor renaming

## Dependencies

- https://github.com/netobserv/network-observability-console-plugin/pull/1302

## Checklist

### DIFF
--- a/api/flowcollector/v1beta2/flowcollector_types.go
+++ b/api/flowcollector/v1beta2/flowcollector_types.go
@@ -1120,7 +1120,7 @@ type FlowCollectorConsolePlugin struct {
 	// Deploy as a standalone console, instead of a plugin of the OpenShift Console.
 	// This is not recommended when using with OpenShift, as it doesn't provide an integrated experience.
 	// [Unsupported (*)].
-	Standalone bool `json:"standalone,omitempty"`
+	Standalone *bool `json:"standalone,omitempty"`
 
 	//+kubebuilder:validation:Minimum=0
 	//+kubebuilder:default:=1
--- a/api/flowcollector/v1beta2/helper.go
+++ b/api/flowcollector/v1beta2/helper.go
@@ -55,10 +55,16 @@ func (spec *FlowCollectorSpec) UsePrometheus() bool {
 	return spec.Prometheus.Querier.Enable == nil || *spec.Prometheus.Querier.Enable
 }
 
-func (spec *FlowCollectorSpec) UseConsolePlugin() bool {
+func (spec *FlowCollectorSpec) UseWebConsole() bool {
 	return (spec.UseLoki() || spec.UsePrometheus()) &&
 		// nil should fallback to default value, which is "true"
 		(spec.ConsolePlugin.Enable == nil || *spec.ConsolePlugin.Enable)
+}
+
+func (spec *FlowCollectorSpec) UseStandaloneConsole(hasPluginAPI bool) bool {
+	// defaults to true if there's no plugin API, false otherwise
+	return (spec.ConsolePlugin.Standalone != nil && *spec.ConsolePlugin.Standalone ||
+		spec.ConsolePlugin.Standalone == nil && !hasPluginAPI)
 }
 
 func (spec *FlowCollectorSpec) UseHostNetwork() bool {
--- a/api/flowcollector/v1beta2/zz_generated.deepcopy.go
+++ b/api/flowcollector/v1beta2/zz_generated.deepcopy.go
@@ -589,6 +589,11 @@ func (in *FlowCollectorConsolePlugin) DeepCopyInto(out *FlowCollectorConsolePlug
 		*out = new(bool)
 		**out = **in
 	}
+	if in.Standalone != nil {
+		in, out := &in.Standalone, &out.Standalone
+		*out = new(bool)
+		**out = **in
+	}
 	if in.Replicas != nil {
 		in, out := &in.Replicas, &out.Replicas
 		*out = new(int32)
--- a/internal/controller/consoleplugin/consoleplugin_objects.go
+++ b/internal/controller/consoleplugin/consoleplugin_objects.go
@@ -44,13 +44,14 @@ const metricsPort = 9002
 const metricsPortName = "metrics"
 
 type builder struct {
-	info     *reconcilers.Instance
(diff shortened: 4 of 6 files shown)


jotak:
(keeping as draft for now, I'll see if I can even have mTLS)

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -547,7 +548,22 @@ type ServerTLS struct {
 
 	// Reference to the CA file when `type` is set to `Provided`.
 	// +optional
-	ProvidedCaFile *FileReference `json:"providedCaFile,omitempty"`
+	ProvidedCAFile *FileReference `json:"providedCaFile,omitempty"`
+}
+
+// `ClientServerTLS` define the TLS configuration for both client and server sides
+type ClientServerTLS struct {
```
jotak, line 561: Hey @OlivierCazade - you introduced recently "ServerTLS", and now I need yet something slightly different, to handle both sides in 1 config
If you have an idea to not have yet-another-different TLS config, let me know; But I think it's hard to do without breaking changes... So I've added a bullet in https://issues.redhat.com/browse/NETOBSERV-1735 (the v1beta3 wishlist)

README.md
```diff
@@ -54,48 +57,17 @@ Loki is not mandatory but improves the overall experience with NetObserv.
...
 
 # OR minimal install (Prometheus/Loki must be installed separately)
-helm install my-netobserv -n netobserv --create-namespace netobserv/netobserv-operator
+helm install netobserv -n netobserv --create-namespace netobserv/netobserv-operator
 ```
 
-You can now create a `FlowCollector` resource. Refer to the [Configuration section](#configuration) of this document. A short `FlowCollector` should work, using most default values, plus with the standalone console enabled:
-
-```bash
-cat <<EOF | kubectl apply -f -
```
leandroberetta, line 66: Can we keep this yaml snippet here too? I think it's useful for a quick view.

api/flowcollector/v1beta2/flowcollector_types.go
```diff
@@ -1510,6 +1530,22 @@ type SubnetLabel struct {
...
+	// - `Auto` (default) to try to determine if TLS can be enabled based on the running environment.
+	// - `Auto-mTLS` to preconfigure mTLS. [Unsupported (*)].
+	// +kubebuilder:validation:Enum:="Disabled";"Provided";"Auto";"Auto-mTLS"
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:="Auto"
+	TLSType TLSConfigType `json:"tlsType,omitempty"`
+
+	// TLS or mTLS configuration when `type` is set to `Provided`.
+	// +optional
+	ProvidedCertificates *ClientServerTLS `json:"providedCertificates,omitempty"`
```
leandroberetta, line 1553: Should we add webhooks validations for the tls feature such as checking for these certs if provided is configured?.

README.md
jotak: yes, np - I hesitated to remove it (for less duplication) - but I agree it's useful here as well

api/flowcollector/v1beta2/flowcollector_types.go
jotak: I think that would go against good practices, although I'm right now trying to find a reference and don't find it. Basically, since validation webhooks live in the `api` package, they should in theory not rely on dependencies (or be minimalist about it), because that's typically a package that others may need to pull (e.g. another project that integrates with netobserv) and having dependencies on things like k8s clients could be troublesome. So, we try to limit dependencies here. Plus, I think to remember that we don't want to do i/o during webhook calls, but not 100% sure on that.

You can note that we're already not the good guys here, because we have a dependency on `internal/pkg/cluster`, which itself depends on... kube clients and so on ; that's something we should refactor at some point, but currently this refactoring would be simple, because we only use that dependency to get the `cluster.Info` struct, not for calling functions that would run i/o. If we start adding i/o here, we won't be able to do the refactoring anymore. (It makes me think, maybe it's time to do that refactoring..)

jotak: hey, I'm not sure if that's what you was thinking about, but I added a simple check that the Provided config is not empty (it won't check live the presence of certificates, but at least, it makes sure the configuration is consistent)

jotak:
@leandroberetta I fixed a bug, added tests and improved the doc in my last commit

api/flowcollector/v1beta2/flowcollector_types.go
leandroberetta: Yes, I was thinking about that, thanks!

jotak:
merging for the community release ; will be qe-approved post-merge

PR #2520, opened 2026-03-10
NETOBSERV-2375: make consistent status handling

jotak:
## Description

Address some old "TODOs" in status management. All statuses now follow this pattern:

- They are initialized unknown
- When controller found unused, status set to unused
- At any point, if an error occurs, it's set with that error
- Deployment/DaemonSet check progress set status either as InProgress, or Ready
- New Degraded condition can be optionally set; a degraded status does not prevent it from being "ready".

Make the LokiStack status fit in this model (it was previously using an entirely different pattern)

Also, previously, in 2 different places LokiStack was fetched to get its status (for propagating status into FlowCollector status and into  console plugin config) Now consolidated into a single entry point to get the loki status.

Status changes:
- New status: WaitingEBPFAgents
- New status: WaitingWebConsole
- New status: WaitingDemoLoki
- New status: WaitingLokiStack
- Renamed status: WaitingFlowCollectorLegacy => WaitingFlowCollectorController
- Removed: LokiIssue and LokiWarning

## Dependencies

- https://github.com/netobserv/netobserv-web-console/pull/1319

## Checklist


internal/controller/flowcollector_controller.go
```diff
@@ -116,62 +102,22 @@ func (r *FlowCollectorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
...
-	r.status.SetReady()
 	defer r.status.Commit(ctx, r.Client)
 
 	err = r.reconcile(ctx, clh, desired)
 	if err != nil {
 		l.Error(err, "FlowCollector reconcile failure")
 		// Set status failure unless it was already set
 		if !r.status.HasFailure() {
-			r.status.SetFailure("FlowCollectorGenericError", err.Error())
+			r.status.SetFailure("ChildReconcilerError", err.Error())
```
jpinsonneau, line 112: I'm not convinced by this namming but I don't know what could be better here actually 🤔

jpinsonneau (approved):
Nice work ! That will help a lot the users to have a clear view of what's going on !

Thanks @jotak

internal/controller/flowcollector_controller.go
jotak: "SubReconcilerError" ? "DelegateReconcilerError" ? "DependentReconcilerError" ?

Amoghrd:
@jotak PR looks good overall except I am not able to simulate a degraded state.
- I tried high memory requests for FLP(100Gi) it showed failed.
- I tried setting an incorrect nodeSelector for eBPF pods such that it does not get deployed on any node but the eBPF pod status still showed Ready with no failure which was weird too.

So let me know how do I hit the degraded state scenario

jotak:
@Amoghrd the degraded condition should show up in the same conditions that previously would trigger a "LokiWarning" condition, so if you had tests previously with LokiWarning, you should be able to reuse the same. It's currently the only thing that triggers degraded.

Looking at the [unit tests](https://github.com/netobserv/netobserv-operator/pull/2520/changes#diff-4138b6120958f6620b50ef19977e3c5a82b3f47df88baf009e03f02637cb1a2eR350-R355), an example of issue that triggers it is this one: `Warning: The schema configuration does not contain the most recent schema version`

> I tried setting an incorrect nodeSelector for eBPF pods such that it does not get deployed on any node but the eBPF pod status still showed Ready with no failure which was weird too.

would be good to check if it's a regression of this PR; if not, we should open a bug

Amoghrd:
Yup the eBPF with nodeSelector bug is a regression, created a bug https://redhat.atlassian.net/browse/NETOBSERV-2674

PR #2546, opened 2026-03-13
NETOBSERV-2655: additionalIncludeList field to append metrics to default

Amoghrd:
## Description

Adds a new `additionalIncludeList` field to allow appending metrics to the defaults without breaking backward compatibility.

**Problem**: The original `includeList` field overwrites the default metrics list. Users who want to add a few extra metrics need to specify all defaults + their additions, which is cumbersome.

**Solution**: 
- Keep `includeList` behavior unchanged (overwrites defaults) - **backwards compatible**
- Add new `additionalIncludeList` field that appends to defaults
- Fields are mutually exclusive - if both are set, `includeList` takes precedence and a validation warning is shown

**Usage Examples**:

Append to defaults:
```yaml
spec:
  processor:
    metrics:
      additionalIncludeList:
        - namespace_egress_bytes_total
        - namespace_ingress_bytes_total
```

Override defaults (existing behavior):
```yaml
spec:
  processor:
    metrics:
      includeList:
        - node_ingress_bytes_total
```

## Changes
- Added `additionalIncludeList` field to `FLPMetrics` API
- Updated `GetIncludeList()` to support the new field
- Added validation to warn when both fields are set
- Added comprehensive tests for new functionality
- Updated CRDs, bundle, and documentation

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Add an optional "additional include list" to append extra metrics to the default metrics; ignored when an explicit include list is provided.
  * Added a runtime validation warning describing the mutual-exclusivity/precedence between the two lists.

* **Documentation**
  * Docs, CRD schema, and UI descriptors updated to document the new field, allowed values, and behavior.

* **Tests**
  * Added unit tests for appending, deduplication, and precedence between include lists.


jotak (changes requested):
We can't change the behaviour of the existing API, it would be a breaking change.
We need instead to create a new setting, such as "additionalIncludeList", and keep the existing as it is.

Amoghrd:
Sounds good, will do that

api/flowcollector/v1beta2/flowcollector_alert_types.go
```diff
@@ -114,17 +114,31 @@ type HealthRuleThresholds struct {
...
 		} else {
 			// When loki is disabled, increase what's available through metrics by default, to minimize the loss of information
 			list = DefaultIncludeListLokiDisabled
 		}
+
+		// AdditionalIncludeList only applies when using defaults (IncludeList is not set)
+		if s.Processor.Metrics.AdditionalIncludeList != nil {
+			for _, m := range *s.Processor.Metrics.AdditionalIncludeList {
+				metricName := string(m)
+				if !containsMetric(list, metricName) {
```
jotak: `slices.Contains` should do the same, I don't think a new function is needed

jotak (commented):
just a small comment, LGTM otherwise

jotak (dismissed):
Thanks @Amoghrd , lgtm

Amoghrd:
(Rebased without any changes)
Could I get LGTM, approve labels for merge? TIA!

jpinsonneau (approved):
LGTM in terms of code

PR #2552, opened 2026-03-16
NETOBSERV-2234: Validation warning about missing metrics

Amoghrd:
## Description

Warning saying that some of the queries run from the console plugin will fail when user does not add some of the default metrics in includeList

## Dependencies

https://github.com/netobserv/netobserv-operator/pull/2546

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently filter out metrics tied to disabled features from both default and user-specified metric include lists.
  * Added a validation warning when the console plugin is enabled in Prometheus-only mode and a custom metrics list omits required default metrics.

* **Tests**
  * Expanded tests covering console-plugin validation and missing-default scenarios with customized include lists.


jpinsonneau (dismissed):

api/flowcollector/v1beta2/flowcollector_validation_webhook.go
```diff
@@ -435,6 +436,42 @@ func (v *validator) validateFLPMetricsForAlerts() {
 	}
 }
 
+func (v *validator) validateFLPMetricsForConsolePlugin() {
```
jotak, line 440: I don't think this warning is going to be always relevant. Some metrics from the defaults can be disabled or modified without necessarily being a problem for the console display. I can think for example about "node_to_node_ingress_flows_total" which is used in default alerts but not in the console queries (and it generates already a warning if it's disabled AND the related alert is in use)

Which means sometimes the warning would be triggered as a false positive.

jotak: TBH, we have improved the situation lately about missing metrics in the console, I wonder if it's still worth it to try doing something in the webhook

Amoghrd: So is it better to have a separate list of metrics only required for console which does not include alerts?
Or since we implemented incremental metrics too, is it better to not have this warning at all?

Amoghrd:
Rebased with main and updated wrt coderabbit review comment

api/flowcollector/v1beta2/flowcollector_validation_webhook.go
Amoghrd: @jotak Any comments for above approach? ^^

jotak: I think the difficult thing here is to make the operator aware of what the console plugin is going to query. IMO that's something potentially error prone in the long run / could be a maintenance pain.

Also, we need to put ourselves in the user's shoes when they want to disable some metrics - what's their reason: it's likely because they want to reduce the load/storage on prometheus. A warning could be annoying for them, if they did that on purpose. I think we've done some effort on the console side to better clarify why some queries don't work, ie. because some metrics are disabled, and so this could be sufficient and a warning is not needed anymore? wdyt?

PS: sorry for the late reply!

Amoghrd: yeah sounds good, will close out this PR and mark the JIRA as WontDo. Can reopen it in the future if there is a need

PR #2598, opened 2026-03-27
NETOBSERV-2701 handle plugin images from CSV

jpinsonneau:
## Description

- Handle multiple plugin images from CSV
- Bind OpenShift version to each image directly from the controller flag

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Operator now supports multiple console-plugin variants (PF4, PF5, and newer) and automatically selects the appropriate plugin for your cluster version.

* **Refactor**
  * Console-plugin configuration moved to a version-aware, semicolon-separated mapping with runtime resolution and fallback semantics.

* **Chores**
  * CLI and startup now use a single console-plugin-images flag and validate image mappings at startup, failing fast on invalid input.

* **Tests**
  * Added unit tests for parsing, validation, and resolution of console-plugin image variants.


bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -990,8 +990,7 @@ spec:
                 - --leader-elect
                 - --ebpf-agent-image=$(RELATED_IMAGE_EBPF_AGENT)
                 - --flowlogs-pipeline-image=$(RELATED_IMAGE_FLOWLOGS_PIPELINE)
-                - --console-plugin-image=$(RELATED_IMAGE_CONSOLE_PLUGIN)
-                - --console-plugin-compat-image=$(RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT)
+                - --console-plugin-images=4.0.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF4);4.15.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF5);4.22.0=$(RELATED_IMAGE_CONSOLE_PLUGIN)
```
memodi:
```suggestion
                - --console-plugin-images=4.14.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF4);4.15.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF5);4.22.0=$(RELATED_IMAGE_CONSOLE_PLUGIN)
```

jpinsonneau: Addressed in https://github.com/netobserv/netobserv-operator/pull/2598/commits/0b1f341b17252cb21b61bca499516f3f3cc78bef

Also took the opportunity to raise error if the minimum version set in the array doesn't match current version. It means the operator manager will fail to reconcile and the user will have to change the env variable to continue.

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
```diff
@@ -990,8 +990,7 @@ spec:
                 - --leader-elect
                 - --ebpf-agent-image=$(RELATED_IMAGE_EBPF_AGENT)
                 - --flowlogs-pipeline-image=$(RELATED_IMAGE_FLOWLOGS_PIPELINE)
-                - --console-plugin-image=$(RELATED_IMAGE_CONSOLE_PLUGIN)
-                - --console-plugin-compat-image=$(RELATED_IMAGE_CONSOLE_PLUGIN_COMPAT)
+                - --console-plugin-images=4.14.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF4);4.15.0=$(RELATED_IMAGE_CONSOLE_PLUGIN_PF5);4.22.0=$(RELATED_IMAGE_CONSOLE_PLUGIN)
```
jotak: I think we should continue to have it work with images below 4.14 (especially 4.12 and 4.13, but I would even say anything below 4.14)
IMO the fact that we don't support it officially doesn't mean to make it suddenly break

internal/pkg/manager/config.go
```diff
@@ -24,15 +36,85 @@ type Config struct {
...
+// It iterates ConsolePluginImageVariants (sorted ascending by MinVersion) and returns the image from
+// the last variant whose MinVersion is satisfied. On non-OpenShift clusters, the last entry
+// (highest MinVersion, most current) is used as default. Returns an error if no variant matches.
+func (cfg *Config) ResolveConsolePluginImage(clusterInfo *cluster.Info) (string, error) {
+	if len(cfg.ConsolePluginImageVariants) == 0 {
+		return "", fmt.Errorf("no console plugin image variants configured")
+	}
+	if !clusterInfo.IsOpenShift() {
+		return cfg.ConsolePluginImageVariants[len(cfg.ConsolePluginImageVariants)-1].Image, nil
+	}
```
jotak, line 111: If I understand currently, non-openshift users would have to configure "--console-plugin-images fakeversion=$(RELATED_IMAGE_CONSOLE_PLUGIN)" so that it works?
If so, can we find something else ; maybe simply returning early the value of $(RELATED_IMAGE_CONSOLE_PLUGIN) in that case?

jotak: maybe we need actually 2 different flags, mutually exclusive:
- --openshift-console-plugin-images for openshift
- --web-console-image for non-openshift

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: That was based on that feedback https://github.com/netobserv/netobserv-operator/pull/2598#discussion_r3001211305

That said, we can change the behavior to have a `default=` image in the array that will be taken if no other versions match and also handle a single image without any version defined to [address the upstream case](https://github.com/netobserv/netobserv-operator/pull/2598#discussion_r3032652077)

WDYT ?

jotak: yes, that would work for me!

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jotak: @memodi I don't think we should enforce that kind of thing in code; supported versions is one thing but it doesn't mean to break things intentionally for unsupported versions

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/netobserv-operator/pull/2598/commits/2e7e6eaf22b74c872477426451c9bbed15f188f4

jpinsonneau: done in https://github.com/netobserv/netobserv-operator/pull/2598/commits/62ecc7c729071dfb479b830d3cfe4d6faeb4575c

bundle/manifests/netobserv-operator.clusterserviceversion.yaml
jpinsonneau: ended up with `default` entry. See https://github.com/netobserv/netobserv-operator/pull/2598/commits/62ecc7c729071dfb479b830d3cfe4d6faeb4575c

jpinsonneau:
Rebased and addressed feedback

jpinsonneau:
Rebased without changes

jpinsonneau:
Rebased without changes

memodi (dismissed):
/lgtm

/cc @kapjain-rh for testing.

Amoghrd:
I am assigned for this task, so will go ahead and test it out

PR #2706, opened 2026-05-04
NETOBSERV-2705: Filter disabled feature fields from OTEL exports

Amoghrd:
## Summary

Fixes [NETOBSERV-2705](https://redhat.atlassian.net/browse/NETOBSERV-2705): Filter out fields for disabled agent features from OpenTelemetry exports to prevent null values in exported logs.

## Changes

Added a `TransformFilter` stage before the OTEL `TransformGeneric` stage in the FLP pipeline to remove fields associated with disabled features.

### Modified Files
- `internal/controller/flp/flp_pipeline_builder.go`:
  - New method `getOtelFieldFilterRules()` generates filter rules based on enabled agent features
  - Updated `createOpenTelemetryStage()` to apply field filtering before OTEL transformation
- `internal/controller/flp/flp_pipeline_builder_test.go`: Minor formatting

### Fields Filtered (when features disabled)
- **DNSTracking** (6 fields): `DnsErrno`, `DnsFlags`, `DnsFlagsResponseCode`, `DnsId`, `DnsLatencyMs`, `DnsName`
- **PacketDrop** (5 fields): `PktDropBytes`, `PktDropPackets`, `PktDropLatestDropCause`, `PktDropLatestFlags`, `PktDropLatestState`
- **FlowRTT** (1 field): `TimeFlowRttNs`
- **IPSec** (1 field): `IPSecStatus`
- **UDNMapping** (2 fields): `SrcK8S_NetworkName`, `DstK8S_NetworkName`

## Test Plan

OTEL exporter test (test_exporters.go:74977) verifies that:
1. OTEL collector receives flow records with TLS encryption
2. Fields for disabled features are not present in exported logs (previously showed as null)

## Related Issues

- JIRA: [NETOBSERV-2705](https://redhat.atlassian.net/browse/NETOBSERV-2705)

## Summary by CodeRabbit

* **Refactor**
  * Optimized OpenTelemetry export field handling to conditionally remove unused fields when agent features are disabled.

* **Tests**
  * Minor formatting updates to test configuration verification.

### DIFF
--- a/internal/controller/flp/flp_pipeline_builder.go
+++ b/internal/controller/flp/flp_pipeline_builder.go
@@ -728,12 +728,21 @@ func (b *PipelineBuilder) createOpenTelemetryStage(name string, spec *flowslates
 	metricsEnabled := spec.Metrics.Enable != nil && *spec.Metrics.Enable
 
 	if logsEnabled || metricsEnabled {
+		// Filter out fields for disabled features before transforming to OTEL format
+		filterRules := b.getOtelFieldFilterRules()
+		var currentStage config.PipelineBuilderStage
+		if len(filterRules) > 0 {
+			currentStage = fromStage.TransformFilter(fmt.Sprintf("%s-field-filter", name), api.TransformFilter{Rules: filterRules})
+		} else {
+			currentStage = *fromStage
+		}
+
 		// add transform stage
 		transformCfg, err := otelConfig.GetOtelTransformConfig(spec.FieldsMapping)
 		if err != nil {
 			return err
 		}
-		transformStage := fromStage.TransformGeneric(fmt.Sprintf("%s-transform", name), *transformCfg)
+		transformStage := currentStage.TransformGeneric(fmt.Sprintf("%s-transform", name), *transformCfg)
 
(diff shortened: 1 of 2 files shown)


Amoghrd:
Verified by running the OTEL QE automation. Works as expected!
I dont think this needs QE approval, marking the PRs as no-qe. Let me know if otherwise

jotak (changes requested):
I don't think there's anything to change in the operator here ; even if that works, it would add a filtering step in FLP which consumes CPU.
The underlying problem is that the "transform generic" logic in FLP is unsound, it's just not right that it creates an output key when the input key is missing. It's something to fix in FLP, and it shouldn't add any CPU cycle

jotak:
=> this would be the fix in FLP that solves the same problem : https://github.com/netobserv/flowlogs-pipeline/pull/1259

Amoghrd:
Ah okay, that is definitely better. I took the approach of how we do in normal flows wherein we remove fields if the features are not enabled but your PR prevents it from creating in the first place.
I will close out these PRs and test yours

PR #2793, opened 2026-06-04
NETOBSERV-2575: allow setting toleration for static plugin

jotak:
## Description

New env "STATIC_PLUGIN_INHERIT_TOLERATION_SUBSCRIPTION", to provide the name of the subscription resource to inherit toleration config from.

Also propagate node selector and affinity

## Dependencies

n/a

## Checklist


main.go
```diff
@@ -238,3 +239,9 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func readConfigFromEnv(c *manager.Config) {
+	c.StaticPluginConfig = manager.StaticPluginConfig{
```
jotak, line 246: team / @jpinsonneau : I'm thinking about generalizing this approach, of reading directly config from env, rather than through command line arguments in the operator. In the YAML, instead of having this pattern "ENV => command line injection => code reads from command line", we'd just have "ENV => code reads from env", so a little bit simpler.

wdyt? do you think we're losing something in not driving that through args?

jpinsonneau: That's probably cleaner. We just need to ensure it doesn't break any tests / CI runs that would pass args directly without setting the ENVs.

Also, if you plan to replace existing args, we need to handle defaults properly.

jotak: yes, ok. I'll leave this PR just as that a prepare a refactoring in another one

jotak: @jpinsonneau I have my next PR ready to push for replacing the existing args, so I'd like this one to go in first

jpinsonneau (dismissed):

Amoghrd:
Hey @jotak I am not able to find the env var "STATIC_PLUGIN_INHERITED_TOLERATION_SUBSCRIPTION" in the CSV. Debugging with claude and it mentioned the var not being present in the config/manager/manager.yaml file.
Screenshot 2026-06-23 at 2 22 58 PM

Amoghrd:
Also I see the operator pod kept crashing and restarting with the above PR image. Debugging this is what claude found:
```
- Line 8: olm "github.com/operator-framework/api/pkg/operators/v1alpha1" - the import is there in the controller
- Line 56: Watches(&olm.Subscription{}, ...) - trying to watch Subscription objects
  
But in main.go, the olm package is NOT imported and olm.AddToScheme(scheme) is NOT called in the init() function!
  
This is a critical bug in PR #2793. The fix is simple - add these two lines to main.go:
  
1. Import: operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
2. In init(): utilruntime.Must(operatorsv1alpha1.AddToScheme(scheme))
  
Summary of issues found in PR #2793:
1. ✅ Missing scheme registration (causes crash) - Subscription type not registered in main.go
2. ✅ Missing environment variable in CSV - STATIC_PLUGIN_INHERITED_TOLERATION_SUBSCRIPTION not in config/manager/manager.yaml
3. ✅ Missing bundle regeneration - needs make update-bundle after adding env var
  
The operator is crashing because it tries to watch a resource type that isn't registered in the scheme!
```

jotak:
@Amoghrd something might have gone wrong during the rebase, let me check...

jotak:
@Amoghrd ok, I screwed up here .... should be all good now, hopefully!

Amoghrd:
No worries, will test it out once I have a cluster

PR #2798, opened 2026-06-08
Packet Translation and Sampling 50 fix

Amoghrd:
## Description

Fixing PacketTranslation and sampling 50 test to use extra loki params for filtering flows such that the test is more robust and will get the accurate flows

## Dependencies

n/a

## Test Results
4.14: 
- [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-High-79015-Verify PacketTranslation feature [Serial] PASSED [502.917 seconds]
- [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-Longduration-Medium-78480-NetObserv with sampling 50 [Serial][Slow] PASSED [1033.885 seconds]

4.22:
- [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-High-79015-Verify PacketTranslation feature [Serial] PASSED [590.190 seconds]
- [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-Longduration-Medium-78480-NetObserv with sampling 50 [Serial][Slow] PASSED [1152.157 seconds]

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Tightened integration tests for packet translation: queries now require specific translation labels and ports instead of broad zone filters, improving precision.
  * Streamlined validations to assert presence of matching flows with clearer label expectations.
  * Removed a legacy helper and simplified gating/cleanup logic to reduce maintenance and flakiness.


integration-tests/backend/test_flowcollector.go
```diff
@@ -1620,7 +1620,7 @@ var _ = g.Describe("[sig-netobserv] Network_Observability", func() {
 				Template:       flowFixturePath,
 			}
 
-			defer func() { _ = flow.DeleteFlowcollector(oc) }()
+			defer flow.DeleteFlowcollector(oc)
```
oliver-smakal: The linting fails on this. Not a big deal from my side, we can change the linting if we don't want this.

integration-tests/backend/test_flowcollector.go
```diff
@@ -1738,7 +1749,7 @@ var _ = g.Describe("[sig-netobserv] Network_Observability", func() {
 				Template:      flowFixturePath,
 			}
 
-			defer func() { _ = flow.DeleteFlowcollector(oc) }()
```
oliver-smakal, line 1741: The linting fails on this. Not a big deal from my side, we can change the linting if we don't want this.

integration-tests/backend/test_flowcollector.go
Amoghrd: I feel we dont need a func for a single line. There are other single line defer statements which dont have funcs so trying to move away from this

memodi: it's not about having it inside of a function, it's about catching the return value that function returns. Here I kept the behavior as is by ignoring it as we did before, however we could update it catch the error in the defer func block and also make the linting pass. Other single line defer statements would not be an issue since the function called in defer statement may not be returning any values.

I think we should avoid adding custom linting rules and generally follow best practices unless its something that's custom to our modules.

Amoghrd: We can do that in a separate PR since we need to update all tests for defer statements.

memodi: we don't need to update other defer statement, perhaps you missed edit I made on my comment:

> Other single line defer statements would not be an issue since the function called in defer statement may not be returning any values.

Amoghrd: Yup, reverted that change

memodi (approved):
/lgtm
/approve

PR #2799, opened 2026-06-09
NETOBSERV-2768: Implement DUMP_EVENTS_ON_FAILURE env var

Amoghrd:
## Description

- Implement DUMP_EVENTS_ON_FAILURE env var and default it to false.
- Summary of all tests ran at the end of the run
- Fix upgrade test wherein later tests would deploy released version of the operator since we are updating the global var in the test. Now using a local var in the test such that the global var is not polluted
- Fix YAML lint errors

## Dependencies

n/a

## Test Results

Ran 2 tests
Passed: 1, Failed: 1, Skipped: 1
==========================================================================================================

FAILED TESTS (1):
1. [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-Critical-59746-NetObserv upgrade testing [Serial]
   /Users/amoghrd/repos/netobserv/netobserv-operator/integration-tests/backend/test_flowcollector.go:853 [646.025 seconds]
   Error: Expected
    <string>: 1.12.0
not to equal
    <string>: 1.12.0

PASSED TESTS (1):
1. [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-High-79015-Verify PacketTranslation feature [Serial] [695.229 seconds]

SKIPPED TESTS (1):
1. [sig-netobserv] Network_Observability with Loki Author:memodi-NonPreRelease-Medium-77894-TechPreview Network Policies Correlation [Serial] [399.483 seconds]
   Reason: Skipping because the TechPreviewNoUpgrade is not enabled on the cluster.

## Checklist

## Summary by CodeRabbit

* **Tests**
  * New env var support: DUMP_EVENTS_ON_FAILURE enables cluster-wide event/log dumping on test failure for better diagnostics.
  * Improved test reporting: after-suite now lists failed, passed, and skipped tests with clearer totals and per-test timings and failure details.
  * More reliable upgrade tests and cleanup guidance: upgrade flows avoid shared global state and emphasize uninstalling the operator after completion.


integration-tests/backend/backend_suite_test.go
```diff
@@ -24,11 +26,19 @@ var _ = g.BeforeSuite(func() {
...
 	e2eframework.AfterReadingAllFlags(exutil.TestContext)
 
+	// Control verbose event dumping on test failures via DUMP_EVENTS_ON_FAILURE env variable
+	// Default: disabled (false)
+	// Set DUMP_EVENTS_ON_FAILURE=true to enable verbose cluster-wide event dumps for debugging
+	dumpEvents, err := strconv.ParseBool(os.Getenv("DUMP_EVENTS_ON_FAILURE"))
+	if err != nil {
+		dumpEvents = false
+	}
+	exutil.TestContext.DumpLogsOnFailure = dumpEvents
```
memodi: since we're trying to reduce the dependency on openshift/origin, how about we directly use e2eframework to set the context that exutil is using setting here https://github.com/openshift/origin/blob/main/pkg/clioptions/clusterdiscovery/provider.go#L75 ?

Amoghrd: Done.
Tested by adding a debug log. Ran a test by passing `DUMP_EVENTS_ON_FAILURE=true` and could see the log showing `DEBUG: DUMP_EVENTS_ON_FAILURE env=true, parsed=true, TestContext.DumpLogsOnFailure=true`

integration-tests/backend/backend_suite_test.go
```diff
@@ -22,13 +24,21 @@ var _ = g.BeforeSuite(func() {
 	flag.Parse()
 
 	// Set up provider config after parsing flags
-	e2eframework.AfterReadingAllFlags(exutil.TestContext)
+	e2eframework.AfterReadingAllFlags(&e2eframework.TestContext)
```
memodi, line 27: @oliver-smakal - do you see any side effect if we didn't use `exutil.TestContext`?

oliver-smakal: No side effect here, it should be exactly the same thing (not only the same type/value, they point to the same space in memory):

[test/extended/util/test_setup.go:](https://github.com/openshift/origin/blob/02b417258847387a7d11b374a2f9f4e87bfdea38/test/extended/util/test_setup.go)
```go
import (
	"k8s.io/kubernetes/test/e2e/framework"
)

var (
	// TestContext which allows injecting context data into tests
	TestContext *framework.TestContextType = &framework.TestContext
)
```

That means that even if we would use the exutil package and it would modify  `exutil.TestContext` due to some side effects, it is the same thing as `&framework.TestContext` so it will not break anything.

memodi:
new tests are now running for e2etests in CI, the lint-fmt test is failing

Amoghrd:
@memodi Ran YAML lint locally and pushed changed. CI has an issue with YAML lint command not found. Fix in [release#80672](https://github.com/openshift/release/pull/80672)

PR #2801, opened 2026-06-11
NETOBSERV-2676: TLS Alert

Amoghrd:
## Description

Create TLS Alert for TLS traffic < 1.2 in cluster.
Integrated with HealthRule and also created a sample YAML.
Monitors for TLS 1.0, 1.1, or SSL traffic. Alert fires when >5% of TLS flows use insecure versions (2-minute rate window, 5-minute duration).
Sample YAML: Has above rule and also has a warning for TLS1.2. Not incorporated this into HealthRule since it might be too much. 

Also updated the QE integration test to check for alert and TLS 1.1 traffic

## Dependencies

n/a

## Test Results
[sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-High-88455-Verify TLS Tracking feature [Serial]
  /Users/amoghrd/repos/netobserv/netobserv-operator/integration-tests/backend/test_flowcollector.go:2750
  • PASSED [802.037 seconds]
  •==========================================================================================================
  Backend Suite - 1/1 specs • SUCCESS! [802.990 seconds]
  
  Ran 1 tests
  Passed: 1, Failed: 0, Skipped: 0
  ==========================================================================================================
  
  PASSED TESTS (1):
  1. [sig-netobserv] Network_Observability with Loki Author:aramesha-NonPreRelease-High-88455-Verify TLS Tracking feature [Serial] [802.037 seconds]

   SUCCESS! 13m22.989751874s --- PASS: TestBackend (803.29s)
  PASS
  ok    e2etests        809.051s

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added `TLSInsecureVersion` health-rule support to flag deprecated/insecure TLS versions and emit a warning alert (grouped by namespace) when TLS tracking is enabled.
  * Included the new rule in the default health-rule set and allowed it to be disabled via `spec.processor.metrics.disableAlerts`.

* **Documentation**
  * Updated FlowCollector docs, Health Rules docs, and CRD/Helm schemas to include `TLSInsecureVersion`.
  * Added a TLS security `PrometheusRule` sample.

* **Tests**
  * Extended TLS end-to-end and unit tests to cover TLS 1.3 and insecure TLS 1.1 scenarios.


internal/pkg/metrics/alerts/tls_insecure_version.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (r *tlsInsecureVersion) RecordingName() string {
+	return buildRecordingRuleName(r.ctx, "tls_insecure_version", "2m")
+}
+
+func (r *tlsInsecureVersion) GetAnnotations() (map[string]string, error) {
+	healthAnnot := newHealthAnnotation(r.ctx)
+	healthAnnot.TrafficLink = &trafficLink{
+		BackAndForth:      false,
+		ExtraFilter:       `tls_version=~"TLS 1.0|TLS 1.1|SSL.*"`,
```
jotak: did you check if that works in the "Network traffic" link? I have a doubt, but not sure...

internal/pkg/metrics/alerts/tls_insecure_version.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	return map[string]string{
+		"summary": "Insecure TLS version detected",
+		"description": fmt.Sprintf(
+			"NetObserv is detecting more than %s%% of traffic using insecure TLS versions (TLS 1.0, TLS 1.1, or SSL)%s.",
+			r.ctx.getLowestThreshold(),
+			getAlertLegend(r.ctx),
+		),
+		"runbook_url":       buildRunbookURL(r.ctx.template),
```
jotak, line 36: we'll need another PR for the runbook

internal/pkg/metrics/alerts/tls_insecure_version.go
```diff
@@ -0,0 +1,57 @@
...
+		healthAnnotationKey: encodeHealthAnnotation(healthAnnot),
+	}, nil
+}
+
+func (r *tlsInsecureVersion) Build() (*monitoringv1.Rule, error) {
+	// TLS egress traffic monitoring
+	if r.ctx.side == asDest {
```
jotak: why only egress? An in-cluster client connecting to an external service would have only the client request covered but not the server response

internal/pkg/metrics/alerts/tls_insecure_version.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, nil
+	}
+
+	metric, totalMetric := getMetricsForRule(r.ctx)
+	// Filter for insecure TLS versions: TLS 1.0, TLS 1.1, or SSL variants
+	filter := getPromQLFilters(r.ctx, `TLSVersion=~"TLS 1\\.0|TLS 1\\.1|SSL.*"`)
+	totalFilter := getPromQLFilters(r.ctx, "")
```
jotak: I think we need to filter the total to only include flows having TLSVersion defined, else it would misrepresent the percentage. Because TLSVersion is only known on handshake packets, not on AppData packets, so there can be a lot of flows that are TLS but we don't know which version.

I think this would work:

```suggestion
	filter := getPromQLFilters(r.ctx, `TLSVersion=~"TLS 1\\.0|TLS 1\\.1|SSL.*"`)
	totalFilter := getPromQLFilters(r.ctx, `TLSVersion!=""`)
```

jotak:
Thanks @Amoghrd ! Review done, just a couple of comments, nothing big

leandroberetta (dismissed):
lgtm, thanks @Amoghrd !

internal/pkg/metrics/alerts/tls_insecure_version.go
Amoghrd: removed the if condition and added backAndForth = True in netflow_traffic link

internal/pkg/metrics/alerts/tls_insecure_version.go
Amoghrd: Yup this didnt work, updating similar to DNSAlert to see if that works

Amoghrd: Updated the filter and now its working as expected
Screenshot 2026-06-22 at 12 40 18 PM

Amoghrd:
Sorry @leandroberetta had to fix a small bug in the extra_filter section of the alert which @jotak had pointed out before. Its working now!

internal/pkg/metrics/alerts/tls_insecure_version.go
Amoghrd: Runbook PR https://github.com/openshift/runbooks/pull/446

jotak (approved):
thanks @Amoghrd , LGTM
/lgtm

PR #2814, opened 2026-06-22
NETOBSERV-2390 Trimmed junit, linux util commands and e2etests labelling workflow

memodi:
## Description

[NETOBSERV-2390](https://redhat.atlassian.net/browse/NETOBSERV-2390) Trimmed junit, linux util commands and e2etests labelling

- Now junit xml files will only have NetObserv tests reporting when `JUNIT_REPORT_FILE` is used, usage: `export JUNIT_REPORT_FILE=junit_netobserv_e2e.xml`. Using `--junit-report` flag would have unfiltered xml file.
- e2etests images now include common linux utility tools and commands.
- New PRs with only changes in integration-tests/backend directory will be automatically labelled with "integration-tests".

## Dependencies

n/a

## Checklist

  * [X] Regression tests only (e.g. refactoring with no user-facing change).

## Summary by CodeRabbit

* **Chores**
  * Added automated PR labeling for integration test changes.
  * Enhanced backend test environment with additional OS utilities.

* **Tests**
  * Implemented JUnit report generation for integration test suite with filtering.


.github/workflows/label-integration-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+        script: |
+          const { data: files } = await github.rest.pulls.listFiles({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: context.payload.pull_request.number,
+            per_page: 100,
+          });
+          const allUnderIntegrationTests = files.every(f =>
+            f.filename.startsWith('integration-tests/backend/')
+          );
```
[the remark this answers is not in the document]
memodi: 100 files limit is good for integration-tests

oliver-smakal:
> Now junit xml files will only have NetObserv tests reporting when JUNIT_REPORT_FILE is used, usage: export JUNIT_REPORT_FILE=junit_netobserv_e2e.xml. Using --junit-report flag would have unfiltered xml file.

Wouldn't it be better to have the filter work with the flag `--junit-report` flag directly? I don't really see a use when we would need the result unfiltered and this way, it would behave like regular ginkgo and no additional documentation/knowledge would be necessary.

memodi:
> Wouldn't it be better to have the filter work with the flag `--junit-report` flag directly? I don't really see a use when we would need the result unfiltered and this way, it would behave like regular ginkgo and no additional documentation/knowledge would be necessary.

Yes, I agree, that would have been perfect and I tried to have claude do it that way, however 
registerReportAfterSuiteNodeForAutogeneratedReports is called from RunSpecs ([core_dsl.go](https://github.com/openshift/onsi-ginkgo/blob/v2.27.2-openshift-4.22/core_dsl.go#L304)) — which runs AFTER var _ = g.ReportAfterSuite(...) declarations at init time. So ginkgo's autogenerated node is pushed LAST, meaning it runs AFTER ours. Our filter runs first, then ginkgo overwrites with the full unfiltered XML.

Amoghrd:
LGTM

.github/workflows/label-integration-tests.yaml
```diff
@@ -0,0 +1,34 @@
+name: Label integration-tests PRs
+on:
+  pull_request_target:
```
jotak, line 3: workflow looks good to me
... we're close to the danger zone with pull_request_target, maybe some day we should write a lint script that makes sure we never introduce weaknesses unintentionally there.
But this one is fine
/lgtm

memodi: > we're close to the danger zone with pull_request_target.

I understand pull_request is more secure than pull_request_target but here it needs for secrets to edit labels. can you elaborate what do you mean by danger zone?

jotak: That's still the good old "[pwn request](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)" vulnerability. No problem here, because you don't check out the code, run risky script, or use some untrusted/unpinned 3rd party action.
The tl;dr is that `pull_request_target` runs with more privileges than `pull_request` , e.g. it has access to secrets, and is sometimes misused to run potentially untrusted code, which an attacker can leverage to steal secrets. Even variable interpolation must be checked carefully because attackers can do script injection, for example, if we used `${{ github.ref_name }}` or other user-controlled variables.

So this type of workflow must be audited carefully.

jotak: => https://github.com/netobserv/netobserv-operator/pull/2820

memodi:
tested workflow on fork repo PR https://github.com/memodi/netobserv-operator/pull/2/ (thanks @Amoghrd )

/approve

memodi:
the e2etest failed appears to be a bug in operator, I filed it here https://redhat.atlassian.net/browse/NETOBSERV-2839

PR #2823, opened 2026-06-25
NETOBSERV-2515: Implement dynamic TLS configuration from OpenShift API Server

leandroberetta:
OpenShift lets cluster admins configure a cluster-wide `TLSSecurityProfile` on the `APIServer` resource (`Old` / `Intermediate` / `Modern` / `Custom`). NetObserv Operator watches this setting and propagates the resulting TLS minimum version and cipher suites to all the components it manages: **flowlogs-pipeline (FLP)**, the **eBPF agent**, and the **console plugin(s)** (`netobserv-plugin` / `netobserv-plugin-static`). When the profile changes, the operator gracefully restarts itself, which triggers a full reconcile of the `FlowCollector` and rolls out the new settings to every managed workload.

## Test Case 0 — No profile configured behaves like Intermediate

**Objective**: confirm that when `spec.tlsSecurityProfile` is absent, the operator behaves as if `Intermediate` were explicitly set — both for the TLS config it actually applies and for the baseline it uses to detect future changes. This matters because it explains an easy false negative: patching the profile to `Intermediate` as a first step produces **no observable change** (nothing actually changed from the operator's point of view).

### Steps

1. Ensure no profile is set:
   ```
   oc patch apiserver cluster --type=merge -p '{"spec":{"tlsSecurityProfile":null}}'
   ```
2. Check operator logs on next restart (or current logs if already running with no profile):
   ```
   oc logs -n netobserv deployment/netobserv-controller-manager | grep "TLS profile"
   ```
3. Check `TLS_MIN_VERSION` on any managed pod:
   ```
   oc get pod -n netobserv -l app=flowlogs-pipeline -o jsonpath='{.items[0].spec.containers[0].env}' | python3 -m json.tool | grep -A1 'TLS_'
   ```
4. Patch the profile explicitly to `Intermediate`:
   ```
   oc patch apiserver cluster --type=merge -p '{"spec":{"tlsSecurityProfile":{"type":"Intermediate","intermediate":{}}}}'
   ```

### Expected Results

- Step 2 log line: `OpenShift detected but no TLS profile configured in APIServer, using secure defaults`.
- Step 3: `TLS_MIN_VERSION` = `771` (TLS 1.2, same as `Intermediate`).
- Step 4: **no operator restart, no pod rollout** — confirms the no-profile default and the explicit `Intermediate` profile are treated as identical (no spurious restarts on a no-op change).

## Test Case 1 — TLS profile propagation on OpenShift

**Objective**: confirm that changing the cluster TLS profile cascades automatically to all NetObserv-managed components.

### Steps

1. Check the current TLS profile (likely unset — see Test Case 0):
   ```
   oc get apiserver cluster -o jsonpath='{.spec.tlsSecurityProfile}'
   ```
2. Set the profile to `Modern` (TLS 1.3 only — chosen because it's guaranteed to differ from the implicit default, so the change is actually observable):
   ```
   oc patch apiserver cluster --type=merge -p '{"spec":{"tlsSecurityProfile":{"type":"Modern","modern":{}}}}'
   ```
3. Watch the operator logs for the profile change detection and restart:
   ```
   oc logs -n netobserv deployment/netobserv-controller-manager -f
   ```
4. Wait ~30–60s for the operator pod to restart and for the FlowCollector to reconcile, then check the TLS env vars on managed pods:
   ```
   oc get pod -n netobserv -l app=flowlogs-pipeline -o jsonpath='{.items[0].spec.containers[0].env}' | python3 -m json.tool | grep -A1 'TLS_'
   oc get pod -n netobserv-privileged -l app=netobserv-ebpf-agent -o jsonpath='{.items[0].spec.containers[0].env}' | python3 -m json.tool | grep -A1 'TLS_'
   oc get pod -n netobserv -l app=netobserv-plugin -o jsonpath='{.items[0].spec.containers[0].env}' | python3 -m json.tool | grep -A1 'TLS_'
   ```

### Expected Results

- Operator log shows, right after restart:
  ```
  Using OpenShift TLS profile	{"profileType": "Modern"}
  Setting up TLS profile watcher for graceful restart on profile changes
  ```
- The operator pod is recreated (`oc get pod -n netobserv -l app=netobserv-operator` shows a new pod name / fresh age).
- All managed workloads restart within roughly the same time window as the operator (FLP, eBPF agent, console plugin(s)) — check with:
  ```
  oc get pods -n netobserv -o wide
  oc get pods -n netobserv-privileged -o wide
  ```
  All pods should show an `AGE` close to the moment the profile was changed (a few seconds apart is normal, since each Deployment/DaemonSet rolls independently once the operator updates its spec).
- Each managed pod shows:
  ```
  TLS_MIN_VERSION=772          (0x0304 = TLS 1.3)
  TLS_CIPHER_SUITES=4865,4866,4867
  TLS_CURVE_PREFERENCES=4588,29,23,24
  ```

## Test Case 3 — Custom TLS profile

**Objective**: confirm the `Custom` profile type (explicit `minTLSVersion` + `ciphers` list) is correctly parsed and propagated. This is implemented and unit-tested in `internal/pkg/tlsconfig/config.go` (`TestComposeTLSConfig_CustomProfile`, `TestConfigToEnvVars_CustomProfile`), including the edge case of a malformed/empty `custom` block.

### Steps

1. Apply a custom profile (including an explicit `groups` list):
   ```
   oc patch apiserver cluster --type=merge -p '{
     "spec": {
       "tlsSecurityProfile": {
         "type": "Custom",
         "custom": {
           "minTLSVersion": "VersionTLS12",
           "ciphers": [
             "ECDHE-ECDSA-CHACHA20-POLY1305",
             "ECDHE-RSA-CHACHA20-POLY1305",
             "ECDHE-ECDSA-AES128-GCM-SHA256",
             "ECDHE-RSA-AES128-GCM-SHA256"
           ],
           "groups": ["secp384r1", "X25519"]
         }
       }
     }
   }'
   ```
2. Watch operator logs and pod rollouts (same as Test Case 1, steps 4–5).
3. Check env vars on any managed pod.

### Expected Results

- `TLS_MIN_VERSION` = `771` (VersionTLS12, per the reference table).
- `TLS_CIPHER_SUITES` contains exactly these 4 decimal IDs (order may vary): `52393,52392,49195,49199` — corresponding to the 4 OpenSSL cipher names requested, translated to Go's numeric IDs.
- `TLS_CURVE_PREFERENCES` = `24,29` (in that order: `secp384r1`=24, then `X25519`=29

### Negative sub-case: malformed Custom profile

```
oc patch apiserver cluster --type=merge -p '{"spec":{"tlsSecurityProfile":{"type":"Custom"}}}'
```
(`type: Custom` with no `custom` block.)

**Expected**: operator logs an error (`custom TLS profile specified but Custom field is nil`) and does **not** crash or leave managed components without a TLS config — it falls back gracefully rather than propagating a broken config. Confirm the operator pod stays `Running` (no CrashLoopBackOff) and existing managed pods are not disrupted.

## Dependencies

https://github.com/netobserv/flowlogs-pipeline/pull/1297
https://github.com/netobserv/netobserv-ebpf-agent/pull/1015
https://github.com/netobserv/netobserv-web-console/pull/1610

Will be addressed in a follow-up:
https://github.com/netobserv/netobserv-cli/pull/552

## Checklist

* [x] Does the changes in PR need specific configuration or environment set up for testing?


internal/controller/reconcilers/common.go
```diff
@@ -8,17 +8,20 @@ import (
...
-	Namespace   string
-	ClusterInfo *cluster.Info
-	Loki        *helper.LokiConfig
-	Vendor      constants.Vendor
+	Watcher      *watchers.Watcher
+	Namespace    string
+	ClusterInfo  *cluster.Info
+	Loki         *helper.LokiConfig
+	Vendor       constants.Vendor
+	IsDownstream bool
```
jpinsonneau: Is that used somewhere ?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/netobserv/netobserv-operator
 
-go 1.25.7
+go 1.26
```
jpinsonneau: any reason bumping that in that PR ? It doesn't seems related 🤔

main.go
```diff
@@ -228,6 +237,33 @@ func defaultStringEnv(env, def string) string {
...
+		return nil
+	}
+	setupLog.Info("Setting up TLS profile watcher for graceful restart on profile changes")
+	return (&tlspkg.SecurityProfileWatcher{
+		Client:                mgr.GetClient(),
+		InitialTLSProfileSpec: *tlsProfileSpec,
+		OnProfileChange: func(_ context.Context, oldSpec, newSpec configv1.TLSProfileSpec) {
+			setupLog.Info("TLS profile has changed, initiating graceful shutdown to reload",
+				"oldProfile", oldSpec, "newProfile", newSpec)
+			os.Exit(0)
```
jpinsonneau: We should cancel the manager context instead of calling Exit(0) here.

Something like:
```
	ctx, stop := context.WithCancel(ctrl.SetupSignalHandler())
	defer stop()
```
in main function

and then call `stop` from `OnProfileChange` / `OnAdherencePolicyChange` funcs

WDYT ?

internal/pkg/tlsconfig/platform.go
```diff
@@ -0,0 +1,56 @@
...
+	if err := c.Get(ctx, client.ObjectKey{Name: "cluster"}, apiServer); err != nil {
+		if k8serrors.IsNotFound(err) {
+			// APIServer object not present (e.g. test environments) — treat as no profile configured
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get APIServer config: %w", err)
+	}
+
+	// Return the TLS security profile (may be nil if not set)
+	return apiServer.Spec.TLSSecurityProfile, nil
```
jpinsonneau, line 43: So we capture only `TLSSecurityProfile` and not `TLSAdherence` ?

internal/controller/reconcilers/common.go
jotak: it probably reappeared after a rebasing conflict, @leandroberetta ? I removed it in a PR, which was merged yesterday

internal/pkg/manager/manager.go
```diff
@@ -56,7 +59,7 @@ type PostCreateHook = func(ctx context.Context) error
 
 type Manager struct {
 	manager.Manager
-	ClusterInfo *cluster.Info
+	ClusterInfo *cluster.Info // Contains TLSProfile via ClusterInfo.GetTLSProfile()
```
jotak: nit: I'm not sure the comment is really necessary here? (why saying it contains TLSProfile more than the other fields it contains?)

internal/pkg/manager/manager.go
```diff
@@ -38,7 +41,7 @@ import (
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;create;update;watch
 //+kubebuilder:rbac:groups=apiregistration.k8s.io,resources=apiservices,verbs=list;get;watch
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;prometheusrules,verbs=get;create;delete;update;patch;list;watch
-//+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions;networks,verbs=get;list;watch
+//+kubebuilder:rbac:groups=config.openshift.io,resources=apiservers;clusterversions;networks,verbs=get;list;watch
```
jotak: there has been another issue here with a rebase conflict, but it seems it was from your previous PR https://github.com/netobserv/netobserv-operator/pull/2650/changes#diff-7cb1cf7698036af9cdd80b008abcc16dc4e146182d9e8d8429c977f8687979ad

I moved all these kubebuilder annotations in `roles.go` , now they are duplicated. Can you remove the duplicates? (ie. remove everything from here and just update roles.go for apiservers rbac)

internal/pkg/cluster/cluster.go
```diff
@@ -56,6 +60,8 @@ var (
 	LokiStack      APIName = APIName("lokistacks." + lokiv1.GroupVersion.String())
 )
 
+// NewInfo creates cluster Info, discovering available APIs.
+// On OpenShift, the TLS profile is fetched from the APIServer and refreshed periodically by the refresh loop.
 func NewInfo(ctx context.Context, cfg *rest.Config, dcl *discovery.DiscoveryClient, onRefresh func()) (*Info, func(ctx context.Context) error, error) {
```
jotak: sounds like the `onRefresh` arg is now unused (never set from NewInfo call), I think we can remove from that func signature?

internal/pkg/manager/manager.go
```diff
@@ -73,6 +76,49 @@ func NewManager(
...
+	flowslatest.CurrentClusterInfo = info
+
+	// Step 3: Configure TLS for metrics and webhook servers
+	tlsProfile := info.GetTLSProfile()
+	if tlsProfile != nil {
+		log.Info("Using OpenShift TLS profile", "profileType", tlsProfile.Type)
+	} else if info.IsOpenShift() {
+		log.Info("OpenShift detected but no TLS profile configured in APIServer, using secure defaults")
+	} else {
+		log.Info("Not running on OpenShift, using secure TLS defaults")
```
jotak: I would not even mention OpenShift here. That would look weird for users running another vendor / vanilla

```suggestion
		log.Info("Using secure TLS defaults")
```

internal/pkg/tlsconfig/config.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	tlsConfig.MinVersion = minVersion
+	tlsConfig.CipherSuites = cipherSuites
+
+	return tlsConfig, nil
+}
+
+// profileToTLSConfig converts an OpenShift TLSSecurityProfile to TLS version and cipher suites
+func profileToTLSConfig(profile *configv1.TLSSecurityProfile) (uint16, []uint16, error) {
```
jotak: I think error handling should be done differently, for a few reasons:
- to better surface errors (in status conditions)
- to fallback on secure defaults in case of error
- one issue would be silently ignored here, it's if the TLS version is unknown, we fall back to 1.2 (imagine tomorrow TLS 1.4 is out and we didn't update our code, we would silently ignore 1.4 and use 1.2 instead)

Also, if cipher suites are empty but TLSVersion is correctly set, we still want to use the correct TLS version (and vice-versa) ; ie. do not fail everything for 1 error.

internal/pkg/tlsconfig/config.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if len(cipherSuites) == 0 {
+		return 0, nil, fmt.Errorf("no valid cipher suites found in profile")
+	}
+
+	return minVersion, cipherSuites, nil
+}
+
+// tlsVersionFromString converts TLS version string to uint16 constant
+func tlsVersionFromString(version configv1.TLSProtocolVersion) uint16 {
```
jotak: related to my other comment, this should return an error if the version is unknown/empty

internal/pkg/tlsconfig/config.go
```diff
@@ -0,0 +1,152 @@
...
+
+	case configv1.TLSProfileCustomType:
+		if profile.Custom == nil {
+			return 0, nil, fmt.Errorf("custom TLS profile specified but Custom field is nil")
+		}
+		minVersion = tlsVersionFromString(profile.Custom.MinTLSVersion)
+		cipherNames = profile.Custom.Ciphers
+
+	default:
+		// Unknown profile type - use intermediate as default
```
jotak: here also, surface in status condition that the profile is unkown

internal/pkg/tlsconfig/envvars.go
```diff
@@ -0,0 +1,130 @@
...
+
+// ProfileToEnvVars converts an OpenShift TLS profile to environment variables
+// suitable for injection into component containers.
+// Returns nil if profile is nil (components will use Go defaults).
+func ProfileToEnvVars(profile *configv1.TLSSecurityProfile) ([]corev1.EnvVar, error) {
+	if profile == nil {
+		return nil, nil
+	}
+
+	tlsCfg, err := ComposeTLSConfig(profile)
```
jotak: I think it would be simpler if `ComposeTLSConfig` was called only from the manager / clusterInfo (e.g. change `ClusterInfo.GetTLSProfile` to return `tls.Config` instead of the TLSProfile) : that way, you don't have to deal with errors every time you try to get the `tls.Config`. Would that work? It would simplify things.

internal/pkg/tlsconfig/envvars.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for _, curve := range curves {
+		names = append(names, curveIDToString(curve))
+	}
+
+	return strings.Join(names, ",")
+}
+
+// curveIDToString converts a curve ID to its name
+func curveIDToString(curve tls.CurveID) string {
```
jotak: here and above: not really sure what's best here, is there really a benefit to convert everything to human readable strings? The alternative being be to directly output the uint16s as strings. The latter would avoid conversions back and forth, and also is more future-proof if the standard evolves with new ids.

go.mod
jotak: that's probably required by openshift/api
Anyway, there are other PR opened that do the same, we'll have to go there

main.go
```diff
@@ -228,6 +237,33 @@ func defaultStringEnv(env, def string) string {
 	return def
 }
 
+// setupTLSProfileWatcher configures the SecurityProfileWatcher on OpenShift clusters.
+// On OpenShift with no explicit TLS profile, OCP uses the Intermediate default; we still
+// set up the watcher so that a future explicit profile change is detected and triggers a restart.
+func setupTLSProfileWatcher(mgr *manager.Manager) error {
+	if !mgr.ClusterInfo.IsOpenShift() {
```
jotak, line 263: here and in other places where you check `IsOpenShift` : what about older versions of openshift? I'm not sure TLS profiles existed back in 4.12, unless it was backported...
This feature started to be mandatory in 4.22, so maybe we can check based on version. Maybe add a `UseTLSProfile() bool` func in clusterInfo, that returns true if on OpenShift >= 4.22 ?

internal/pkg/tlsconfig/platform.go
leandroberetta: Good point! TLSAdherence controls whether components should strictly follow the configured profile. For now we only read TLSSecurityProfile and apply it when set, which is ok to me but we can add this logic. For OpenShift as it is would be ok, not sure for upstream.

internal/pkg/tlsconfig/envvars.go
leandroberetta: Good point, I implemented it. Thanks!

main.go
jotak: hmm I see you added UseTLSProfile() but it's called only here in the watcher, and not in other places where we fetch the tls profile; I don't think it makes sense, because on older openshift, TLS Profile would be fetched once and then unwatched. IMO we should either use it everywhere, or rollback removing UseTLSProfile().

I guess the reason why you don't call it from cluster.NewInfo() is because, at this stage, the openshift version is still unknown, so it's not gonna work (unless doing a bigger refactoring). I'm fine to roll back to your initial code without `UseTLSProfile`. We must just make sure this code works on older openshift that don't have the TLSProfiles / APIServer APIs.

jotak (commented):
one last comment other than that lgtm

main.go
leandroberetta: Ok, I rollbacked the change. For what I see, the API change was way before 4.12 so we should be good.

jpinsonneau (dismissed):

leandroberetta:
CrashLoopBackOff Issue

Changing the cluster's APIServer TLS security profile (apiserver/cluster spec.tlsSecurityProfile, e.g. Modern ↔ Intermediate) triggers a control-plane rollout (kube-apiserver + openshift-apiserver redeploy) and MCO node reboots. During that window the operator could land in CrashLoopBackOff. 

Investigation found two distinct mechanisms behind the symptom:

1. Startup API-discovery crash (exit 1 — preventable, fixed here). 

While openshift-apiserver rolls out, aggregated discovery of security.openshift.io is transiently unavailable. cluster.NewInfo → fetchAvailableAPIs treated this as fatal, so the process called os.Exit(1) — a
  real startup crash. Kubelet's back-off then produced CrashLoopBackOff and kept the operator down for the duration of the outage. This is preventable: by retrying discovery within a single start, the container rides through the transient outage and comes up cleanly, so the crash never happens.

2. Graceful reload on a stale-baseline read (exit 0 — correct by design, not fixed).

To know when to reload, the SecurityProfileWatcher compares the live APIServer profile against a baseline snapshot taken once at container startup — a direct read of apiserver/cluster at boot. This baseline is not the administrator's configured value; it is simply whatever that point-in-time read returned. On a profile change the watcher reloads by exiting 0 so the process restarts and reconfigures its servers with the new profile.

The issue is that, during a control-plane rollout (made worse by two overlapping profile changes), the boot read can return a stale/lagging value. The container then latches that stale value as its baseline. Moments later the watcher's informer delivers the already-settled, correct value, and the comparison reports a "change" — even though the administrator changed the profile only once, before this container even started. What the watcher detects is not a real profile change; it is the stale boot snapshot correcting itself.

Because the baseline is re-captured on every container start (it is not persisted across restarts), a container that starts while the read is still unsettled can latch a stale value again, so the exit-0 reloads repeat and cluster together within the rollout window. Kubelet flags the clustered restarts as CrashLoopBackOff regardless of exit code.

Scope of the fix

This PR fixes only mechanism 1, the exit-1 startup crash, which can be prevented outright.

Mechanism 2 is intentionally NOT fixed. Both mechanisms recover on their own once the rollout settles, but they differ in nature: 1 is a genuine crash we can eliminate by riding through the transient outage, whereas 2 is a series of correct restarts — each reload is required because the container really did start with the wrong (stale) profile and must restart to apply the real one. The CrashLoopBackOff status is only a cosmetic side effect of those correct restarts clustering together, has no functional impact, and clears itself.

Changes

- Extend startup discovery retry (internal/pkg/cluster/cluster.go): startupAPIDiscoveryBackoff retries transient critical-API discovery failures for ~5 min (sized to outlast a single 3-replica control-plane rollout; independent of worker-node count) instead of crashing. Fixes mechanism 1.
- Add startupProbe (config/manager/manager.yaml, helm chart, CSV bundles): suspends liveness/readiness until /healthz first responds (budget 30×10s = 5 min ≥ retry window), so the container isn't killed while discovery is still retrying. Required to make the extended retry safe.
- Document mechanism 2 (main.go): comment in setupTLSProfileWatcher explaining the stale-baseline graceful reloads, why the resulting CrashLoopBackOff is a cosmetic side effect of correct restarts, and why it is left unaddressed.

Testing

Live on OCP: none→Modern and chained Modern→Intermediate. Confirmed both rollouts occur (MCP + openshift-apiserver roll). The operator survived across node reboots with zero exit-1 startup crashes and zero liveness kills. Any CrashLoopBackOff observed corresponded to mechanism 2 (exit-0 graceful reload) and self-recovered without intervention.

@kapjain-rh after several test and rounds of testing, above are my findings.

@jotak @jpinsonneau I would to have your opinions and eyes on this situation. Thanks!

jotak:
@leandroberetta so if I understand correctly, your fix works as intended?

What I would have liked to know though, if whether, without any change, it would eventually self-heal as well. 
The "1. Startup API-discovery crash (exit 1 — preventable, fixed here)." could have been approached the same way as point 2, ie. an expected crash? Being unable to fetch the discovery API results in the operator potentially taking wrong decisions about what to deploy (e.g. it thinks it's not on openshift, so it won't deploy openshift specific stuff, or OVN network policy). Of course it will eventually recover, but there's this lapse of time where it could take inconsistent decisions. So I wonder if we shouldn't just let it crash & retry until everything is working correctly and self-heals?

jpinsonneau:
The fix looks correct to me :ok_hand: 

@jotak about 1, without the fix it would self-heal, but via `CrashLoopBackOff` it could keep the operator down up to 5 minutes. The retry avoids the "wrong decisions" and won't proceed with partial results. If retries are exhausted it still errors out cleanly.

@leandroberetta about 2, it might be worth opening a follow-up issue to wait for the value to stabilize before latching, which would reduce the cosmetic `CrashLoopBackOff` noise during rollouts.

WDYT ?

leandroberetta:
> The fix looks correct to me 👌
> 
> @jotak about 1, without the fix it would self-heal, but via `CrashLoopBackOff` it could keep the operator down up to 5 minutes. The retry avoids the "wrong decisions" and won't proceed with partial results. If retries are exhausted it still errors out cleanly.
> 
> @leandroberetta about 2, it might be worth opening a follow-up issue to wait for the value to stabilize before latching, which would reduce the cosmetic `CrashLoopBackOff` noise during rollouts.
> 
> WDYT ?

Sounds ok to me, I will create the follow-up, I prefer to address it after this feature lands on main.

kapjain-rh:
/label qe-approved

netobserv-controller-manager goes into crashloopback back and takes time in reconcile but it is at cluster level reconciliation and it takes time to stabilized so it is expected.

PR #2851, opened 2026-07-21
Fix STATIC_PLUGIN_INHERIT_TOLERATION_SUBSCRIPTION should not be set by default

jotak:
The default config was making an assumption that the subscription is named "netobserv-operator", and that ends up hardcoded in the bundle env (can still be reconfigured manually, but breaking the "out of the box" path)

It's safer to not make this assumption at all. Users who want propagating the toleration from Subscription need to configure manually the env STATIC_PLUGIN_INHERIT_TOLERATION_SUBSCRIPTION with the correct subscription name


memodi:
/lgtm
/label qe-approved
/jira [NETOBSERV-2575](https://redhat.atlassian.net/browse/NETOBSERV-2575)

leandroberetta (approved):
lgtm. Tested and now is working, thanks @jotak !

PR #2892, opened 2026-08-26
fix: add resource requests to static plugin reconciler

### LINKED ISSUE #2891: netobserv-plugin-static Deployment has no resource requests/limits
# `netobserv-plugin-static` Deployment has no resource requests/limits

## Current Behavior

The `netobserv-plugin-static` Deployment deployed by the operator has empty `resources: {}` on its container spec. This means the pod has no CPU/memory requests or limits, which:

- Breaks VPA functionality (VPA requires an existing resource spec)
- Sets QoS class to BestEffort
- Makes scheduling unpredictable

## Root Cause

The static reconciler constructs a fake `FlowCollector` object in code with no `Resources` field set. This bypasses the CRD schema, so the kubebuilder default annotation (`//+kubebuilder:default:={requests:{memory:"50Mi",cpu:"100m"},limits:{memory:"100Mi"}}`) never fires — kubebuilder defaults only apply to CRs created through the Kubernetes API. The zero-value `Resources` flows directly into the pod template, resulting in `resources: {}`.

## Workaround

A Kyverno policy in `openshift-netobserv-operator` injects resource requests/limits at admission time:

```yaml
apiVersion: kyverno.io/v1
kind: Policy
metadata:
  name: add-resource-limits
  namespace: openshift-netobserv-operator
spec:
  rules:
    - name: netobserv-plugin-static
      match:
        any:
          - resources:
              kinds: [Deployment]
              selector:
                matchLabels:
                  app: netobserv-plugin-static
      mutate:
        patchStrategicMerge:
          spec:
            template:
              spec:
                containers:
                  - (name): netobserv-plugin-static
                    resources:
                      limits:
                        cpu: 100m
                        memory: 128Mi
                      requests:
                        cpu: 5m
                        memory: 32Mi
```

## Long-term

A long-term plan should expose resource requests/limits and replica count as tunables in the `FlowCollector` CR. Until then, the static reconciler needs default requests so VPAs can function.

## Environment
- Operator: v1.21.2 (confirmed present in main branch)

ArthurVardevanyan:
The `netobserv-plugin-static` Deployment deployed by the operator has empty `resources: {}` on its container spec. 

This results in:
- **VPA cannot function** (requires an existing resource spec) — this is the primary blocker
- QoS class defaults to BestEffort
- Unpredictable scheduling

Add explicit resource **requests only** (CPU: 10m, memory: 64Mi) to the fake `FlowCollector` object in `consoleplugin_static_reconciler.go`. This ensures the static plugin Deployment has resource requests by default, allowing VPAs to work and pods to get a proper QoS class.

**No limits are included** to avoid breaking existing deployments that rely on the operator-managed replica count and any external autoscaling (HPA/KEDA).

image

## Related Issues

Fixes #2891

## Long-term

This PR is a minimal fix. A long-term plan should expose resource requests/limits and replica count as tunables in the `FlowCollector` CR, allowing users to configure these values alongside the existing console plugin configuration.

## Summary by CodeRabbit

- **New Features**
  - Added configurable CPU and memory requests and limits for the static console plugin.
  - Static plugin deployments now default to 10m CPU and 64Mi memory requests.
  - Deployment configuration exposes environment variables for adjusting static plugin resource settings.
  - Empty limit values indicate that no resource limit is applied.

- **Bug Fixes**
  - Invalid CPU or memory resource values now produce clear configuration errors instead of causing unexpected failures.

### DIFF
--- a/internal/controller/consoleplugin/consoleplugin_static_reconciler.go
+++ b/internal/controller/consoleplugin/consoleplugin_static_reconciler.go
@@ -2,6 +2,7 @@ package consoleplugin
 
 import (
 	"context"
+	"fmt"
 
 	osv1 "github.com/openshift/api/console/v1"
 	olm "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -59,6 +60,10 @@ func (r *StaticReconciler) ReconcileStaticPlugin(ctx context.Context, enable boo
 		}
 	}
 
+	resources, err := buildStaticPluginResources(&r.managerConfig.StaticPluginConfig)
+	if err != nil {
+		return fmt.Errorf("building static plugin resources: %w", err)
+	}
 	// Fake a FlowCollector to create console plugin and expose forms
 	return r.reconcileStatic(ctx, &flowslatest.FlowCollector{
 		Spec: flowslatest.FlowCollectorSpec{
@@ -68,7 +73,7 @@ func (r *StaticReconciler) ReconcileStaticPlugin(ctx context.Context, enable boo
 			ConsolePlugin: flowslatest.FlowCollectorConsolePlugin{
 				Enable:    ptr.To(enable),
 				LogLevel:  "info",
-				Resources: buildStaticPluginResources(r.cfg),
+				Resources: resources,
 				Advanced: &flowslatest.AdvancedPluginConfig{
 					Register:   ptr.To(true),
 					Scheduling: sched,
@@ -78,7 +83,7 @@ func (r *StaticReconciler) ReconcileStaticPlugin(ctx context.Context, enable boo
 	})
 }
 
-func buildStaticPluginResources(cfg *manager.StaticPluginConfig) corev1.ResourceRequirements {
+func buildStaticPluginResources(cfg *manager.StaticPluginConfig) (corev1.ResourceRequirements, error) {
 	cpuRequest := cfg.CPURequest
 	if cpuRequest == "" {
 		cpuRequest = "10m"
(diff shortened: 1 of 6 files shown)


internal/controller/consoleplugin/consoleplugin_static_reconciler.go
```diff
@@ -38,6 +40,12 @@ func (r *StaticReconciler) ReconcileStaticPlugin(ctx context.Context, enable boo
 			ConsolePlugin: flowslatest.FlowCollectorConsolePlugin{
 				Enable:   ptr.To(enable),
 				LogLevel: "info",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("10m"),
+						corev1.ResourceMemory: resource.MustParse("64Mi"),
+					},
+				},
```
jotak: @jpinsonneau what do you think about that? Static plugin is really simple and probably shouldn't ever need something more fancy / configurable, i guess

jpinsonneau: That's on request side so yes it's good enough.

However it's easy to expose that in the CSV so why not adding an env variable just in case ? 😸

ArthurVardevanyan: Updated to include env via csv

jotak (dismissed):
thanks @ArthurVardevanyan !
/lgtm

jpinsonneau (dismissed):

PR #2902, opened 2026-09-01
NETOBSERV-2964 Fix incorrect _count suffix in DNS alert PromQL

jpinsonneau:
## Summary

- Fix [DNSNxDomain and DNSErrors alerts that never fire](https://github.com/netobserv/netobserv-web-console/pull/1700) because the generated PromQL references a non-existent metric name
- The `_count` suffix was incorrectly appended to `dns_flows_total` (a Counter), producing `netobserv_*_dns_flows_total_count` which doesn't exist in Prometheus. Only Histogram metrics have a `_count` sub-metric.
- All other Counter-based alerts (`kernel_drops`, `tls_insecure_version`, `ipsec_errors`, `netpol_denied`) correctly use an empty suffix
- Add PromQL assertion tests for both DNSNxDomain and DNSErrors

## Root cause

In `dns_nxdomain.go` and `dns_errors.go`, `promQLRateFromMetric(metric, "_count", ...)` was used instead of `promQLRateFromMetric(metric, "", ...)`. The `_count` suffix is only valid for Histograms (used correctly in `latency_trend.go` with `_bucket`). Since `dns_flows_total` is registered as a `CounterVec` in FLP, Prometheus only exposes `netobserv_*_dns_flows_total` — there is no `_count` variant.

## Test plan

- [x] Existing alert unit tests pass
- [x] New `TestDNSNxDomainPromql` and `TestDNSErrorsPromql` verify the generated PromQL references the correct metric name with proper filters
- [ ] End-to-end: deploy a FlowCollector with DNSNxDomain health rules and `dns_flows_total` in the metrics includeList, generate NxDomain traffic, verify alerts fire in Prometheus

Made with [Cursor](https://cursor.com)


jotak (approved):
/lgtm
thanks!

PR #2922, opened 2026-09-10
NETOBSERV-2990: manage operands rbac via update-only

jotak:
## Description

Restore operator-managed CRB, with the following guardrails:
- operands CRB are still OLM/helm-installed, but they start as empty shells, with no subject.
- operator has update-only, per-resourceName CRB permission
- replace transitive permissions (operator has same permissions as its operands in order to grant them) with "bind" verb (operator can bind some specific cluster roles)

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added automatic management of component access bindings as FlowCollector features are enabled or disabled.
  - Added cleanup that removes access subjects from preinstalled bindings before a FlowCollector is deleted.
  - Added permissions for the operator to manage these bindings.
  - Added Helm and OpenShift manifest support for the updated access-binding configuration.

- **Documentation**
  - Simplified namespace field descriptions to clarify that they identify where NetObserv pods run.


internal/pkg/roles/roles.go
```diff
@@ -1,65 +1,43 @@
...
-)
+// OperandClusterRoleBindings lists the preinstalled ClusterRoleBindings whose subjects are managed
+// by the operator: subjects are added when operands need them, and emptied when the FlowCollector
+// is deleted. The bindings themselves are preinstalled empty shells and are never created/deleted here.
+var OperandClusterRoleBindings = []ClusterRoleName{
+	LokiWriterRole,
+	FLPInformersRole,
+	HostNetworkRole,
+	ConsoleTokenReviewRole,
+	FlowCollectorViewerRole,
```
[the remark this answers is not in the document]
jotak: done

internal/controller/flp/envtest/flp_controller_envtest.go
```diff
@@ -803,6 +833,13 @@ func ControllerSpecs(env test.Environment, ctxGetter test.ContextGetter) {
 			}, timeout, interval).Should(BeTrue())
 		})
 
+		It("Should deploy Loki roles", func() {
+			By("Expecting FLP Writer ClusterRoleBinding")
+			Eventually(func() interface{} {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, "flowlogs-pipeline")
+			}, timeout, interval).Should(Succeed())
+		})
```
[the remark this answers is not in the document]
jotak: done

internal/controller/reconcilers/reconcilers.go
```diff
@@ -64,6 +68,84 @@ var (
...
+}
+
+// ReconcileClusterRoleBinding updates the current role binding with the provided service account as a subject.
+// It does NOT try to create or delete it: operand CRBs are expected to be preinstalled. The operator does not have create permission.
+func ReconcileClusterRoleBinding(ctx context.Context, cl *helper.Client, namespace, sa string, ref roles.ClusterRoleName, isDelete bool) error {
+	log := log.FromContext(ctx)
+	crb := rbacv1.ClusterRoleBinding{}
+	if err := cl.Get(ctx, types.NamespacedName{Name: string(ref)}, &crb); err != nil {
+		if errors.IsNotFound(err) {
+			return fmt.Errorf("can't reconcile ClusterRoleBinding %s, it should be preinstalled; was it removed? - %w", ref, err)
```
leandroberetta, line 87: If a CRB was previously deleted, EmptyClusterRoleBinding returns an error. As a result, the finalizer is never removed and the FlowCollector remains stuck in Terminating. Since an absent CRB is already effectively empty, NotFound should be treated as success during finalization.

internal/controller/reconcilers/reconcilers.go
```diff
@@ -64,6 +68,84 @@ var (
...
+	ts := o.GetDeletionTimestamp()
+	return ts != nil && !ts.IsZero()
+}
+
+// ReconcileClusterRoleBinding updates the current role binding with the provided service account as a subject.
+// It does NOT try to create or delete it: operand CRBs are expected to be preinstalled. The operator does not have create permission.
+func ReconcileClusterRoleBinding(ctx context.Context, cl *helper.Client, namespace, sa string, ref roles.ClusterRoleName, isDelete bool) error {
+	log := log.FromContext(ctx)
+	crb := rbacv1.ClusterRoleBinding{}
+	if err := cl.Get(ctx, types.NamespacedName{Name: string(ref)}, &crb); err != nil {
```
leandroberetta, line 85: This Get creates a narrow-cache watch for the CRB, but no GetSource/handler is registered with either controller. Therefore, external changes to a managed CRB update the cache but do not trigger reconciliation. Could we register these CRBs as narrow watch sources so removed subjects are restored without waiting for an unrelated event?

leandroberetta (commented):
Thanks for the work @jotak, I left some comments to discuss but looks good overall.

internal/controller/reconcilers/reconcilers.go
jpinsonneau: +1 here


```suggestion
			// A missing binding already satisfies the goal (no subjects). Don't error: this runs from
			// the finalizer, and blocking here (e.g. when CRBs are removed during operator uninstall)
			// would leave the FlowCollector stuck in Terminating forever.
			log.Info("ClusterRoleBinding already absent, nothing to empty", "name", string(ref))
			return nil
```

internal/controller/flp/envtest/flp_controller_envtest.go
jpinsonneau, line 879: Worth adding a cleanup tests also ?

```suggestion

		It("Should remove the Loki writer subject when leaving LokiStack mode", func() {
			By("Expecting the flowlogs-pipeline subject to be removed from the Loki writer ClusterRoleBinding")
			Eventually(func() interface{} {
				// Manual mode no longer needs the Loki writer role: the subject added in LokiStack
				// mode must be removed, leaving the preinstalled binding as an empty shell again.
				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, operatorNamespace /* empty expect list */)
			}, timeout, interval).Should(Succeed())
		})
	})
```

Needs https://github.com/netobserv/netobserv-operator/pull/2922/changes#r4026645693 to pass the expected ns as argument 😉

internal/controller/flp/envtest/flp_controller_envtest.go
```diff
@@ -889,3 +926,19 @@ func checkDigestUpdate(oldDigest *string, annots map[string]string) error {
...
+	if len(rb.Subjects) != len(subjectNames) {
+		return fmt.Errorf("expected %d subjects for %s, got %d; %v", len(subjectNames), role, len(rb.Subjects), rb)
+	}
+	for i, name := range subjectNames {
+		if rb.Subjects[i].Name != name {
+			return fmt.Errorf("expected subject %d for %s to be %s, got %s; %v", i, role, name, rb.Subjects[i].Name, rb)
+		}
+	}
+	return nil
+}
```
jpinsonneau: We should check the namespace too:


```suggestion
func expectClusterRoleBinding(ctx context.Context, k8sClient client.Client, role roles.ClusterRoleName, expectNamespace string, subjectNames ...string) error {
	rb := rbacv1.ClusterRoleBinding{}
	if err := k8sClient.Get(ctx, types.NamespacedName{Name: string(role)}, &rb); err != nil {
		return err
	}
	if len(rb.Subjects) != len(subjectNames) {
		return fmt.Errorf("expected %d subjects for %s, got %d; %v", len(subjectNames), role, len(rb.Subjects), rb)
	}
	for i, name := range subjectNames {
		if rb.Subjects[i].Name != name {
			return fmt.Errorf("expected subject %d for %s to be %s, got %s; %v", i, role, name, rb.Subjects[i].Name, rb)
		}
		// The operator must write the actual operands namespace as the subject namespace
		// (not a hardcoded default): this is what makes custom namespaces work out of the box.
		if rb.Subjects[i].Namespace != expectNamespace {
			return fmt.Errorf("expected subject %d for %s to be in namespace %s, got %s; %v", i, role, expectNamespace, rb.Subjects[i].Namespace, rb)
		}
	}
	return nil
}
```

internal/controller/flp/envtest/flp_controller_envtest.go
```diff
@@ -358,6 +373,21 @@ func ControllerSpecs(env test.Environment, ctxGetter test.ContextGetter) {
...
+
+			By("Not expecting hostnetwork role (not needed with Kafka)")
+			Eventually(func() interface{} {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.HostNetworkRole /* empty expect list */)
+			}, timeout, interval).Should(Succeed())
+
+			By("Not expecting Loki role (requires LokiStack)")
+			Eventually(func() interface{} {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole /* empty expect list */)
+			}, timeout, interval).Should(Succeed())
```
jpinsonneau, line 389:
```suggestion
			By("Expecting informer role binding (independent deployment)")
			Eventually(func() any {
				return expectClusterRoleBinding(ctx, k8sClient, roles.FLPInformersRole, operatorNamespace, "flowlogs-pipeline-informers")
			}, timeout, interval).Should(Succeed())

			By("Not expecting hostnetwork role (not needed with Kafka)")
			Eventually(func() interface{} {
				return expectClusterRoleBinding(ctx, k8sClient, roles.HostNetworkRole, operatorNamespace /* empty expect list */)
			}, timeout, interval).Should(Succeed())

			By("Not expecting Loki role (requires LokiStack)")
			Eventually(func() interface{} {
				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, operatorNamespace /* empty expect list */)
			}, timeout, interval).Should(Succeed())
```

As suggested in https://github.com/netobserv/netobserv-operator/pull/2922/changes#r4026645693

internal/controller/flp/envtest/flp_controller_envtest.go
```diff
@@ -803,6 +833,13 @@ func ControllerSpecs(env test.Environment, ctxGetter test.ContextGetter) {
 			}, timeout, interval).Should(BeTrue())
 		})
 
+		It("Should deploy Loki roles", func() {
+			By("Expecting FLP Writer ClusterRoleBinding")
+			Eventually(func() interface{} {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, "flowlogs-pipeline")
```
jpinsonneau:
```suggestion
				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, operatorNamespace, "flowlogs-pipeline")
```

As suggested in https://github.com/netobserv/netobserv-operator/pull/2922/changes#r4026645693

internal/controller/flp/envtest/flp_controller_envtest.go
```diff
@@ -158,6 +158,21 @@ func ControllerSpecs(env test.Environment, ctxGetter test.ContextGetter) {
...
+			}
+
+			Eventually(func() any {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.FLPInformersRole, "flowlogs-pipeline-informers")
+			}, timeout, interval).Should(Succeed())
+
+			By("Not expecting Loki role (requires LokiStack)")
+			Eventually(func() interface{} {
+				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole /* empty expect list */)
+			}, timeout, interval).Should(Succeed())
```
jpinsonneau, line 174:
```suggestion
			if env == test.EnvOpenShift {
				Eventually(func() any {
					return expectClusterRoleBinding(ctx, k8sClient, roles.HostNetworkRole, operatorNamespace, "flowlogs-pipeline")
				}, timeout, interval).Should(Succeed())
			}

			Eventually(func() any {
				return expectClusterRoleBinding(ctx, k8sClient, roles.FLPInformersRole, operatorNamespace, "flowlogs-pipeline-informers")
			}, timeout, interval).Should(Succeed())

			By("Not expecting Loki role (requires LokiStack)")
			Eventually(func() interface{} {
				return expectClusterRoleBinding(ctx, k8sClient, roles.LokiWriterRole, operatorNamespace /* empty expect list */)
			}, timeout, interval).Should(Succeed())
```

As suggested in https://github.com/netobserv/netobserv-operator/pull/2922/changes#r4026645693

internal/controller/reconcilers/reconcilers.go
jpinsonneau, line 1: Should we add a `reconcilers_test.go` here to ensure some critical scenarios are covered such as a `NotFound` binding not returning an error that will block FlowCollector deletion or simple add / remove subjects cases ?

jpinsonneau (changes requested):

internal/controller/reconcilers/reconcilers.go
jotak: I don't understand the point here. CRB are OLM-installed, if users want to mess up uninstalling that, they'll have problems. You want to make that a success, but if they recreate a flowcollector later, they'll get errors because the operator doesn't recreate CRB. It's pure OLM. So, it's an error if the CRB is missing, even when we're finalizing flowcollector deletion.
On flowcollector deletion, CRB should retrieve their initial state, ie. being present as empty shells.

jotak:
(rebased)

internal/controller/reconcilers/reconcilers.go
jpinsonneau: That's a fair point Joel. In that case why not making the distinction between the two paths ?
- throw error on FC creation / update
- skip on FC deletion

```suggestion
			// We only need the binding to exist when adding a subject (an operand needs the permission).
			// When removing, a binding that no longer exists already satisfies the goal, so it's not an error.
			if isDelete {
				log.Info("ClusterRoleBinding already absent, nothing to remove", "name", string(ref))
				return nil
			}
			return fmt.Errorf("can't reconcile ClusterRoleBinding %s, it should be preinstalled; was it removed? - %w", ref, err)
```

That way, even if CRBs are broken, the user can still delete the FC resource and decide to reinstall the operator if he faces the issue at creation.

My fear is that operator uninstall is stuck if the user select the "delete related resources" popup in the console without that change.

internal/controller/reconcilers/reconcilers.go
jotak: sure let me do it

internal/controller/reconcilers/reconcilers.go
jotak: Ok I'll better test that case, indeed if all resources are deleted at once, I see why the finalizer could be deadlocking. Though it's not here that the fix should be set, there's a special function for the finalize path (line 132) "EmptyClusterRoleBinding".

internal/controller/reconcilers/reconcilers.go
jotak: There was an existing test for FlowCollector deletion & rbac, I extended it to cover the risk of deadlock situation: https://github.com/netobserv/netobserv-operator/pull/2922/changes/cf52e0334aa3acdc0535bffed387b79271343c8b

internal/controller/flp/envtest/flp_controller_envtest.go
jotak: ok, i've added SA namespace check as well

internal/controller/flp/envtest/flp_controller_envtest.go
jotak: I know it's not full coverage, but there's already a bunch of tests added, maybe it's good enough?

internal/controller/reconcilers/reconcilers.go
jotak: done - and refactored a bit narrowcache to make it directly manage the enqueueing reconcile requests (previously was done only in watcher.go, now centralized in narrowcache)

internal/controller/flp/envtest/flp_controller_envtest.go
jpinsonneau: Up to you; don't have a strong opinion on that

jotak: done

leandroberetta (dismissed):
LGTM, thanks for addressing the feedback @jotak

Issue #2850, opened 2026-07-21
Digest pinning in helm chart


jotak:
Current helm chart references operand images by tag. For immutability, it should use SHA digests instead.

Follow-up on #2490

Helion55:
Hi @jotak, I have created a PR for this https://github.com/netobserv/netobserv-operator/pull/2876.

PR #2929, opened 2026-09-15
Allow pinning digests when building bundle

jotak:
## Description

Fixes part of #2850

Still left to do: pinning the bundle inside a catalog, and pinning images in helm

To use it:

```bash
# All component's `main` tag
VERSION=main BPF_VERSION=main FLP_VERSION=main PLG_VERSION=main PIN_DIGEST=true make update-bundle
```

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional image digest pinning for the controller, eBPF agent, flowlogs pipeline, and web console images during deployment preparation.
  * Web console variants are pinned to their corresponding image digests for improved consistency and reproducibility.

* **Changes**
  * Tag-based image references remain available when digest pinning is disabled or unset.
  * Digest pinning requires Docker; Podman is not supported.


leandroberetta (approved):
LGTM, the CI failure seems unrelated. For the podman, I think we can use skopeo in a follow-up?.

jotak:
@leandroberetta yes, either skopeo or podman pull + podman inspect. Skopeo would add yet another dependency (although bypassed by docker)
I think this script will run mostly in CI, not on user's machines, so shouldn't be a problem to keep podman unsupported here - but like said, if someone complains we can change that later :-)

PR #2930, opened 2026-09-16
NETOBSERV-2944: fix: stabilize TLS profile baseline

leandroberetta:
## Description

This change prevents the TLS profile watcher from capturing a stale APIServer TLS profile as its startup baseline during a control-plane rollout.

Before configuring the operator and registering the watcher, startup now waits for two consecutive matching APIServer TLS profile reads, five seconds apart. The existing retry policy is still used for transient API failures, and the stabilization wait is cancellable.

The stabilized profile is used both to configure the operator servers and as the watcher baseline. Legitimate profile changes after startup continue to trigger a graceful exit 0 reload.

## Testing

Monitor the operator and APIServer rollout:

```
watch -n 5 '
oc -n netobserv get pods -l control-plane=controller-manager
oc get clusteroperator kube-apiserver
'
```

Check for BackOff events:

```
oc -n netobserv get events --field-selector reason=BackOff
```

Change to Modern:

```
oc patch apiserver cluster --type=merge \
  --patch='{"spec":{"tlsSecurityProfile":{"type":"Modern","modern":{},"intermediate":null}}}'
```

Change to Intermediate:

```
oc patch apiserver cluster --type=merge \
  --patch='{"spec":{"tlsSecurityProfile":{"type":"Intermediate","intermediate":{},"modern":null}}}'
```

The following transitions were tested using the operator image containing this change:

1. Implicit Intermediate → Modern
2. Modern → Intermediate

For each legitimate change, the expected behavior is:

1. The running operator detects the new profile.
2. It performs one graceful reload with exit code 0.
3. The restarted operator waits for a stable startup profile.
4. It does not trigger additional TLS-profile reloads.
5. No CrashLoopBackOff or BackOff events appear.

## Cluster validation results

The following transitions were tested using the operator image containing this change:

1. Implicit Intermediate → Modern
2. Modern → Intermediate
3. Intermediate → Modern

All three control-plane rollouts completed successfully, with all three API server nodes reaching the new revision:

Available=True
Progressing=False
Degraded=False

For every profile transition:

- Exactly one expected graceful TLS reload was observed.
- The reload exited with code 0.
- No stale-baseline TLS reloads were observed after restart.
- The operator returned to Running and ready.
- No BackOff events were recorded.
- The operator never entered CrashLoopBackOff.

During the second and third rollouts, isolated exit-code 1 restarts were observed when the API server was temporarily unavailable:

```
  Failed to renew lease: context deadline exceeded
  problem running manager: leader election lost
```
These were unrelated to the TLS profile watcher. Each container recovered immediately, the failures did not repeat rapidly enough to trigger kubelet backoff, and no CrashLoopBackOff occurred.

The final cluster state was stable with the Modern TLS profile and all API server nodes on the same revision.

## Dependencies

n/a

## Checklist

   * []  if so please describe it in PR description.
* [X] I have added thorough unit tests for the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * TLS configuration changes are now applied only after five seconds without further changes.
  * Rapid successive profile updates are consolidated, reducing unnecessary restarts.
  * Reverting to the currently applied profile no longer triggers a restart.
  * TLS profile monitoring now stops cleanly when the manager is cancelled.

* **Tests**
  * Added coverage for delayed restarts, debounce resets, no-op reversions, and clean cancellation.


internal/pkg/cluster/cluster.go
```diff
@@ -122,6 +122,59 @@ var startupRetryBackoff = wait.Backoff{
...
+// to configure the operator servers and as the SecurityProfileWatcher's startup baseline.
+func fetchStableTLSProfile(
+	ctx context.Context,
+	fetch func(context.Context) (*configv1.TLSSecurityProfile, error),
+) (*configv1.TLSSecurityProfile, error) {
+	logger := log.FromContext(ctx)
+	var previous *configv1.TLSSecurityProfile
+	hasPrevious := false
+
+	for {
```
jotak: if I understand correctly, this is a blocking call; which adds _at least_ 5 seconds fixed delay latency to every startup (best case scenario).

I'm starting to wonder if this whole pattern of "reconciling by restarting the pod" which sounded easier to do at first, is not becoming more complex and fragile

Have we considered at all the option to apply the TLS config "live" (meaning to restart the manager webhook/metrics servers, if we can do that, i guess).

jotak: @leandroberetta or keeping your current approach, instead of having that blocking 5s stabilization interval at startup, how about this:
- startup as usual without stabilization
- when a change is detected, do not exit immediately: start the stabilization check, but async (goroutine) that will wait 5s (or more) and see if the change is confirmed; if confirmed then exit.

It has several benefits:
- debounce time gives some "buffer" before taking the decision to restart; can accumulate several state changes during that time
- stabilizes not just as a startup mechanism, but also post-startup (e.g. if TLS profile state flaps rapidly A->B->A it can end up ignored)
- no fixed startup delay introduced

wdyt?

leandroberetta: Hi @jotak, thanks for your comments. I think that would be better, yes. 

About stop restarting the operator, I think this is a common thing to do, is even commented here:

https://github.com/openshift/controller-runtime-common/blob/e1187ec555fc5c3fd25017b284e14b8742e19ba4/pkg/tls/controller.go#L73

Doing a live change might have some challenges like ongoing connections and concurrent changes, but I didn't investigate that case deeply to be honest. I prefer keeping this way of restarting the operator I think.

I will try your suggestions.

leandroberetta: @jotak I addressed your concerns, I think, let me know! I tested again against a cluster changing profiles doing the same sequence and I didn't see any CrashLoopBackOff.

jotak (approved):
thanks @leandroberetta !
it sounds like a clean solution to me. We may even increase the debounce timer if we find it still too unstable, that won't hurt normal startup.

**to be included in 2.0**

PR #2937, opened 2026-09-17
ci: pin preview bundle images

leandroberetta:
Assisted-by: OpenAI Codex

## Description

Enables digest pinning in the image publishing workflow and updates workflow tests to verify that bundle image references use immutable SHA256 digests instead of tags.

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

- **Improvements**
  - Generated deployment bundles now reference immutable image digests rather than mutable tags.
  - Release builds include explicit version information and OpenShift bundle settings.
  - Operator and supporting component images are consistently pinned to validated SHA-256 digests.
  - Bundle generation now records the appropriate controller manager image.
  - Release validation verifies image digests and expected release versions.
  - Build output directories are excluded from version control.


.github/workflows/push_image.yml
```diff
@@ -51,7 +51,10 @@ jobs:
           MULTIARCH_TARGETS="${{ env.WF_MULTIARCH_TARGETS }}" IMAGE_ORG=${{ env.WF_ORG }} IMAGE=${{ env.WF_REGISTRY }}/${{ env.WF_IMAGE }}:${{ env.short_sha }} make manifest-build manifest-push
       - name: build bundle
         run: |
-          IMAGE_ORG=${{ env.WF_ORG }} \
+          operator_digest=$(docker buildx imagetools inspect ${{ env.WF_REGISTRY }}/${{ env.WF_IMAGE }}:${{ env.WF_VERSION }} \
```
jotak, line 54: why is it needed to get the digest here? It should be done already through the makefile when PIN_DIGEST=true

.github/workflows/release.yml
```diff
@@ -62,7 +62,17 @@ jobs:
           MULTIARCH_TARGETS="${{ env.WF_MULTIARCH_TARGETS }}" IMAGE_ORG=${{ env.WF_ORG }} IMAGE=${{ env.WF_REGISTRY }}/${{ env.WF_IMAGE }}:${{ env.tag }} make manifest-build manifest-push
 
       - name: build bundle
-        run: IMAGE_ORG=${{ env.WF_ORG }} BUNDLE_VERSION=${{ env.tag }} CLEAN_BUILD=1 make bundle-build
+        run: |
+          operator_digest=$(docker buildx imagetools inspect ${{ env.WF_REGISTRY }}/${{ env.WF_IMAGE }}:${{ env.tag }} \
```
jotak, line 66: same here, I'd prefer to keep that done only from one place, ie. Makefile

hack/test-workflow.sh
```diff
@@ -3,7 +3,8 @@
 test_out="test.out"
 bundle_csv="bundle/manifests/netobserv-operator.clusterserviceversion.yaml"
 short_sha=$(git rev-parse --short=8 HEAD)
-fake_tag="1.0.42"
+# Digest tests need a release whose operator and related images exist in Quay.
```
jotak, line 6: TBH I haven't used this test-workflow.sh for a long time, I wasn't sure it was still ok & not broken ... so it's working? :-)

.github/workflows/push_image.yml
leandroberetta: The digest lookup was added here because PIN_DIGEST=true currently resolves the operator digest only inside the set-manager-images recipe and uses it for the manager image. Later, bundle-nogen sets metadata.annotations.containerImage directly from $(IMAGE), which would still contain the tag.

That said, I agree this should be handled in one place. I’ll move the resolution into a shared Makefile variable and use the resolved reference for both the manager image and the containerImage annotation. Then these workflow-specific lookups can be removed.

jotak: can you check this: https://github.com/netobserv/netobserv-operator/pull/2942 ?
I had a similar problem when trying to reuse pinned digest for the helm target. I guess this could work for you too

leandroberetta: Ok, I will take a look, but I've already resolved it in the Makefile. Will check it anyway, thanks!

hack/test-workflow.sh
leandroberetta: After some modifications the test is working, I added a parametrized way to set my Quay account. I'm still checking it but I will try to push the changes later.

.github/workflows/release.yml
```diff
@@ -62,7 +62,18 @@ jobs:
...
-        run: IMAGE_ORG=${{ env.WF_ORG }} BUNDLE_VERSION=${{ env.tag }} CLEAN_BUILD=1 make bundle-build
+        run: |
+          # Upstream releases do not version PF compatibility images; pin their rolling tags by digest.
+          IMAGE=${{ env.WF_REGISTRY }}/${{ env.WF_IMAGE }}:${{ env.tag }} \
+            IMAGE_ORG=${{ env.WF_ORG }} \
+            VERSION=${{ env.tag }} \
+            PLG_PF4_VERSION=main-pf4 \
+            PLG_PF5_VERSION=main-pf5 \
+            BUNDLE_VERSION=${{ env.tag }} \
+            BUNDLE_TARGET=OpenShift \
```
jotak: It should not target openshift ; if you remove `BUNDLE_TARGET=OpenShift`, then there's no need to pass fake pf4/pf5 versions.

Makefile
```diff
@@ -114,14 +116,13 @@ DATE=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 
 # When PIN_DIGEST is true, store all digests in variables, and export them to avoid
 # duplicate image inspections on nested make calls
-# Pinning assumes quay.io/netobserv images; currently not supported for fork builds (contributions are welcome)
```
jotak, line 117: the rationale for that comment was that if a user wants to pin digests with their own forked repo, it opens the door to more questions like "what if I have forked not the operator, but the agent, and I want to pin that" ? Here you're basically saying: you can use your fork-based operator and pin it, but NOT the other images (agent, flp, etc.) which still have the `quay.io/netobserv/` suffix, and that raises the question why would we favor operator over the other components when pinning? So it's like a half-baked feature, which I wanted to avoid for simplicity, hence the "we don't support forks pinning, but feel free to contribute"

leandroberetta: Yes, I see, I added that comment back and revert that change.

.github/workflows/release.yml
leandroberetta: Fixed

jotak (approved):
thanks!
/lgtm

PR #2943, opened 2026-09-18
NETOBSERV-2990: remove redundant operand CRB creation from backend tests

Amoghrd:
## Summary
- The operator now manages the `netobserv-informers`, `netobserv-hostnetwork`, `netobserv-loki-writer` and `netobserv-token-review` ClusterRoleBindings itself (pre-installed as empty shells, subjects added/removed per FlowCollector feature state — see #2922).
- The backend integration suite's `createRoleBindings()` was still manually creating parallel, per-namespace CRBs granting these exact same permissions. That's redundant now, and worse, it would silently mask any regression in the operator's own CRB reconciliation logic since the tests would keep passing on the manually-granted access regardless.
- Removed that block; the unrelated `secretRBs` (operator secret-creator RoleBindings) and leader-election lease `Role` setup in the same function are untouched.

## Testing

PASSED TESTS (3):
1. [sig-netobserv] Network_Observability Multi-Tenancy Author:memodi-NonPreRelease-Longduration-High-63839-Verify-multi-tenancy [Disruptive][Slow] [874.252 seconds]
2. [sig-netobserv] Network_Observability with Kafka Author:aramesha-NonPreRelease-Longduration-High-57397-High-65116-Low-89197-Verify network-flows export with Kafka and netobserv installation without Loki[Serial] [765.895 seconds]
3. [sig-netobserv] Network_Observability Author:memodi-Critical-53844-Sanity Test NetObserv [Serial] [189.256 seconds]

 SUCCESS! 31m37.868831417s --- PASS: TestBackend (1897.87s)
PASS
ok      e2etests        1898.910s

## Summary by CodeRabbit

* **Refactor**
  * Flow Collector cleanup now waits for related resources to be fully deleted before completing.
  * Cleanup validation confirms that associated permissions no longer reference removed resources, while tolerating already-missing bindings and reporting other deletion errors.
* **Tests**
  * Updated integration coverage to reflect the revised role-binding cleanup behavior.


Amoghrd:
Regression run: [noo-backend-tests#180](https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/netobserv/job/noo-backend-tests/180/console) -> All tests passing

memodi:
@Amoghrd what happens when operator is uninstalled - are these CRBs left behind in cluster? if they're expected to be deleted then we should add that check?

jotak:
LGTM, assuming the tests install the OLM bundle, which carries the CRB

@memodi to answer on the uninstall part, CRB should go away with the OLM bundle - ie. should work well if uninstalled properly via OLM.

memodi:
> LGTM, assuming the tests install the OLM bundle, which carries the CRB

we use catalogsource, which internally should be using the OLM bundle, so I think we should be good here.

> @memodi to answer on the uninstall part, CRB should go away with the OLM bundle - ie. should work well if uninstalled properly via OLM.

in that case, a check/assertion would help to ensure they're indeed deleted on uninstall.

Amoghrd:
@jotak I tested this on a cluster with the PR image installed.

Uninstalled the standard OLM way — oc delete subscription netobserv-operator + oc delete csv netobserv-operator.v0.0.0-sha-e3688192 (same as what the console's "Uninstall Operator" button does under the hood). No FlowCollector was ever created, so the CRBs were still empty shells (no subjects) at uninstall time.

Result: the 5 operand CRBs (netobserv-flowcollector-viewer-role, netobserv-hostnetwork, netobserv-informers, netobserv-loki-writer, netobserv-token-review) and their corresponding ClusterRoles are left behind. Confirmed persistent after 2.5+ min, not a propagation delay.

What did get cleaned up correctly: the CSV, subscription, install plan, operator pod, and the operator's own auto-generated CRB/Role/RoleBinding (the ones OLM derives from the CSV's clusterPermissions, named with a random hash suffix like netobserv-operator.v0.0.-<hash>).

I also checked why: the leftover CRBs carry olm.managed/olm.owner labels, but those aren't wired to any actual teardown logic — they're just static bookkeeping. OLM's real cleanup/discovery mechanism uses a different label (operators.coreos.com/<csv-name>) that only gets applied to resources OLM generates itself from CSV permissions/clusterPermissions. These 5 CRBs are standalone manifests shipped separately in the bundle (from PR#2922), so they never get that tracking label and are effectively invisible to OLM's uninstall path.

I also tried deleting the openshift-netobserv-operator namespace afterward, in case that was gating cleanup — no effect. Expected, since ClusterRoleBindings are cluster-scoped with no ownerReference to the namespace, so namespace deletion can't cascade to them regardless.

OLM has no cleanup path for these CRBs. @memodi original ask for a check/assertion here seems right; worth deciding whether the operator itself should clean these up on shutdown/finalizer, or whether it's an accepted gap for now.

jotak:
@Amoghrd hmm I seem to recall that this is a known issue with olm v0 that v1 is addressing.
The real thing that matters I think, is to make sure that the subject list is emptied upon complete deletion. So we don't leave on the cluster some CRB that point to deleted service accounts in deleted namespace.

Amoghrd:
Added a check for CRB not pointing to a stale/deleted ServiceAccount upon flowcollector deletion

